PR #43, opened 2026-06-12
Handle Gemma4 suffixless expert tensors

jjang-ai:
## Summary
- normalize suffixless Gemma4 MoE source/BF16 expert tensors into `experts.switch_glu`
- keep existing quantized `.weight/.scales` fused expert remapping
- document the BF16/source vs MXFP4/JANG_4M expert-key contract

## Proof
- `scripts/prepare-mlx-metal.sh`
- `swift test --filter gemma4SanitizeSplitsFusedMoEExpertWeights`

Artifacts:
- `/tmp/osaurus-gemma-proof/vmlx-clean-prepare-mlx-metal.log`
- `/tmp/osaurus-gemma-proof/vmlx-clean-gemma4-suffixless-expert-test.log`

## Regression covered
`Unhandled keys ["down_proj", "gate_up_proj"] in language_model.model.layers.0.experts` for Gemma4 source/BF16 MoE bundles with suffixless `experts.gate_up_proj` and `experts.down_proj` keys.

### DIFF
--- a/Libraries/MLXVLM/Models/Gemma4.swift
+++ b/Libraries/MLXVLM/Models/Gemma4.swift
@@ -1242,11 +1242,21 @@ public class Gemma4: Module, VLMModel, KVCacheDimensionProvider {
             if nk.contains(".switch_mlp.") { nk = nk.replacingOccurrences(of: ".switch_mlp.", with: ".experts.switch_glu.") }
             if nk.contains(".experts.down_proj.") {
                 nk = nk.replacingOccurrences(of: ".experts.down_proj.", with: ".experts.switch_glu.down_proj.")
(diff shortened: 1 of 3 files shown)


jjang-ai:
Landed as #260, rebased onto current main. 

Two things surfaced while verifying it, both recorded in #260: the local library has no Gemma4 bundle with suffixless expert keys (all ten ship the quantized fused `experts.switch_glu.*` form), so the new normalization does not fire here — what I could prove is that touching expert sanitize leaves the 26B-A4B MoE bundle byte-identical. And the Gemma4 suite was already red on main for unrelated reasons (two stale source-matching assertions), fixed separately in #259 so this PR could be reviewed against a green baseline.

PR #48, opened 2026-06-13
Store diffusion history boundaries for multi-turn prefix-cache hits

jjang-ai:
## Summary

The Gemma-family chat template appends a generation-control thought stub (`<|channel>thought\n<channel|>`) to the active turn but omits it when the turn is re-rendered as history, so a stored full-prompt boundary can never extension-match the next request (prefix audit: turn 2 diverges 4 tokens before turn 1's boundary). The AR iterators already solve this by storing the processor's conversation-stable `cachePrefixTokenCounts` boundaries; this brings `BlockDiffusionTokenIterator` in line.

Also adds two bench diagnostics: a disk-tier extension probe in `BENCH_SOLO_DISK_RESTORE` (verified: `HIT (disk, matched=138, remaining=20)` from a fresh coordinator) and `BENCH_PREFIX_AUDIT` which renders turn-1/turn-2 prompt shapes through the bundle template and reports the divergence point.

## Validation
- `swift test --filter DiffusionGemmaTests` — 15/15
- `BENCH_SOLO_DISK_RESTORE` exact + extension probes hit the disk tier
- Rotating layers are only trimmable pre-wrap; when trimming is unavailable the boundary store skips gracefully (same contract as the AR path)

### DIFF
--- a/Libraries/MLXLLM/LLMModelFactory.swift
+++ b/Libraries/MLXLLM/LLMModelFactory.swift
@@ -36,6 +36,9 @@ public enum LLMTypeRegistry {
             "gemma4_text": create(Gemma4TextConfiguration.self, Gemma4TextModel.init),
             "gemma4_unified": create(Gemma4TextConfiguration.self, Gemma4TextModel.init),
             "gemma4_unified_text": create(Gemma4TextConfiguration.self, Gemma4TextModel.init),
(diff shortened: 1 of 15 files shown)


jjang-ai:
Superseded by a clean cherry-pick off main (pre-squash branch history conflicted).

PR #71, opened 2026-06-16
Mark Hub.Repo Sendable for Swift 6.2 region isolation

Foscoe63:
## Summary
- Mark `Hub.RepoType` and `Hub.Repo` as `Sendable`.
- Fixes Xcode 27 / Swift 6.2 build failure in `HubApi.swift` when `VMLXHub` is compiled with `StrictConcurrency`:

```
HubApi.swift:849:51: error: sending 'self.repo' risks causing data races
```

`HubFileDownloader.download()` passes `repo` into an async `withTaskCancellationHandler` closure; the repo identity types are immutable `String`-backed values and are safe to send.

## Test plan
- [x] `swift build --target VMLXHub`
- [ ] Osaurus app build on Xcode 27

Made with [Cursor](https://cursor.com)

### DIFF
--- a/Vendors/swift-transformers/Sources/Hub/Hub.swift
+++ b/Vendors/swift-transformers/Sources/Hub/Hub.swift
@@ -77,7 +77,7 @@ public extension Hub {
     }
 
     /// The type of repository on the Hugging Face Hub.
-    enum RepoType: String, Codable {
+    enum RepoType: String, Codable, Sendable {
         /// Model repositories containing machine learning models.
         case models
         /// Dataset repositories containing training and evaluation data.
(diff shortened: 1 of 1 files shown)


jjang-ai:
Merged as #257 — thanks. `Repo` is `let id: String` + `let type: RepoType` and `RepoType` is a String-raw-value enum, so the conformance states an invariant that already held; the value is for downstream hosts that need to carry a `Repo` across an isolation boundary under Swift 6.

For the record, since it changes how the PR should be read: this does not fix a build failure in the repo today. The package compiles clean on Apple Swift 6.3.3 — I built RunBench in release against unmodified main several times while working through the queue — so the region-isolation over-report this was filed against no longer reproduces on 6.2+. Landed as hygiene rather than as a repair.

PR #73, opened 2026-06-17
fix quadratic BPE merge (O(n²) → O(n log n)) on long pre-tokens

RaajeevChandran:
## Summary
Replaced `BPETokenizer.bpe(token:)`'s per-round rescan with an O(n log n) merge using a doubly-linked list of symbols plus a min-heap of candidate merges. Output is byte-identical to the previous implementation; the change is purely algorithmic. This removes a multi-second tokenization stall that made tool-calling on large-vocab models effectively unusable.

## Problem
`bpe(token:)` was O(word_len²): every merge round filtered/scanned all pairs, rebuilt the whole symbol array, and recomputed every pair via `getPairs`. For short words this is fine, but a single long whitespace-free pre-token makes it explode.

This is not hypothetical. On an M2 (24 GB) running Gemma 12B (`gemma-4-12B-it-qat-JANG_4M`) with tools + sandbox enabled, the rendered prompt contained one pre-token of **11,035 characters** — the tool-schema block, which the chat template emits as compact, space-free JSON, so the whitespace-splitting pre-tokenizer never breaks it up. Measured cost:

- `bpe()` ≈ **6.3 s** for that single pre-token (the rest of the ~300 pre-tokens were negligible).
- Total `encode()` ≈ **6.4 s per step**, and because the agent loop re-tokenizes every step, this was paid on every tool round-trip.
- The chat-template render itself was ~3 ms — the entire cost was the merge loop.

## Fix
- Represent the symbol sequence as a doubly-linked list (`prev`/`next`/`alive` arrays).
- Maintain a binary min-heap of candidate adjacent merges keyed by `(rank, leftIndex)`.
- Each merge is O(log n) and there are O(n) merges → **O(n log n)** overall.
- Stale heap entries (a node already consumed, or whose live pair no longer carries the candidate's rank) are skipped on pop.

## Why the output is identical
BPE applies merges in rank order. A pair formed by a merge always outranks the merge that created its components (the merges file lists `a b` before any rule that consumes `ab`), so popping the globally-lowest `(rank, leftIndex)` reproduces the canonical behavior of the previous code: "merge all occurrences of the min-rank pair per round, left to right." Same-rank ties are broken by left index, which preserves the left-to-right, non-overlapping merge order (e.g. `a a a` → `aa a`, not `a aa`).

## Validation
- During development the new path was run alongside the previous implementation on every token via a temporary verification harness; across full tool-calling sessions (tool JSON + generated HTML/CSS fed back as results + prose), prompts up to ~6.6k tokens / ~26k chars, there were **zero** output differences.
- Token counts produced for identical inputs were unchanged from the previous implementation.
- End-to-end tasks (e.g. "build a landing page") completed normally, confirming tool calls still parse.

## Performance
On the case above (per `encode`, same input):

| | before | after |
|---|---|---|
| merge loop | ~6,324 ms | ~83 ms |
| total `encode()` | ~6,340 ms | ~91 ms |
| per-step tokenize (app-observed) | ~6.4 s | ~0.15 s |

~70–80× on this workload; scales ~linearly thereafter (≈17 µs/token to 26k chars). The pathological 11,035-char pre-token is unchanged — it simply tokenizes fast now, so the win is independent of any specific model or chat template.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -766,6 +766,7 @@ let package = Package(
                 "MLXOptimizers",
                 "VMLXJinja",
                 "VMLXTokenizers",
+                "VMLXHub",
                 "MLXLMCommon",
                 "MLXLLM",
                 "MLXVLM",
--- a/Tests/MLXLMTests/BPETokenizerDifferentialTests.swift
+++ b/Tests/MLXLMTests/BPETokenizerDifferentialTests.swift
@@ -1,0 +1,212 @@
+// Copyright 2026 Osaurus AI. All rights reserved.
+// SPDX-License-Identifier: MIT
+//
+// Differential + performance regression for the O(n log n) BPE merge rewrite
+// (osaurus-ai/vmlx-swift#73). The optimized `BPETokenizer.bpe(token:)` (heap +
+// doubly-linked list) claims byte-identical output to the original per-round
+// rescan. That claim rests on a load-bearing invariant ("a pair formed by a
+// merge always outranks the pair that formed it"), so it MUST be locked by a
+// test rather than trusted — there was previously no BPE correctness suite.
+//
+// This test embeds the ORIGINAL algorithm verbatim as the reference oracle,
+// builds a real BPETokenizer from an on-disk Gemma merge table, and asserts the
+// shipped `bpe()` matches the reference across thousands of fuzzed inputs —
+// including the ~11k-char whitespace-free pre-token (compact tool JSON) that
+// motivated the fix. Skips when no Gemma tokenizer is on the machine.
+
+import Foundation
+import XCTest
+
+import VMLXHub
+@testable import VMLXTokenizers
+
+final class BPETokenizerDifferentialTests: XCTestCase {
+
+    // MARK: reference oracle — VERBATIM copy of the pre-#73 bpe(token:)
+
+    private func referenceBpe(_ token: String, _ bpeRanks: [BytePair: Int]) -> String {
+        if token.count <= 1 { return token }
+
+        func getPairs(_ word: [String]) -> Set<BytePair> {
+            var s = Set<BytePair>()
+            for i in 0..<word.count - 1 { s.insert(BytePair(word[i], word[i + 1])) }
+            return s
+        }
+
+        var word = Array(token).map { String($0) }
+        var pairs = Array(getPairs(word))
+
+        while true {
+            let bigrams = pairs.filter { bpeRanks[$0] != nil }
+            if bigrams.count == 0 { break }
+            let bigram = bigrams.min { bpeRanks[$0]! < bpeRanks[$1]! }!
+            let first = bigram.a
+            let second = bigram.b
+            var newWord: [String] = []
+            var i = 0
+            while i < word.count {
+                if let j = word[i..<word.count].firstIndex(of: first) {
+                    newWord.append(contentsOf: word[i..<j])
+                    i = j
+                } else {
+                    newWord.append(contentsOf: word[i..<word.count])
+                    break
+                }
+                if word[i] == first, i < word.count - 1, word[i + 1] == second {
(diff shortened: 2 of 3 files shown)


jjang-ai:
Great speedup and the approach is right — but I caught a correctness regression before we merge: the new `bpe()` is **not** byte-identical to the original on whitespace runs.

### Minimal repro — 32 tab characters
- original (canonical): `⟨16 tabs⟩ ⟨16 tabs⟩` (two 16-tab tokens)
- this PR: `⟨31 tabs⟩ ⟨1 tab⟩`

### Root cause
The byte-identical argument assumes *"a pair formed by a merge always outranks the pair that formed it."* Gemma's merge table violates that for whitespace:

```
rank(\t, \t)   = 1395
rank(\t\t, \t) = 1304   ← lower (higher priority) than its own component pair
```

So the moment a `\t\t` forms, the heap pops `(\t\t,\t)`=1304 and keeps extending the left block, preempting the remaining `(\t,\t)` merges. The original merges **all** occurrences of the round's min-rank pair before reconsidering, so it never does this. SentencePiece tokenizers (Gemma/Llama) deliberately give whitespace runs these non-monotonic ranks, so this isn't a corner case — it hits any input with tab/space/newline runs: indented code, the compact tool JSON this PR is about, pasted output.

### What's solid
The perf win is real and great — on the 11k-char space-free pre-token I measured **8.5 ms vs ~6.3 s**, and **4,148** random fuzzed inputs were byte-identical. It only breaks on the invariant-violating (whitespace) inputs.

### Suggested fix
Process the heap in true rank-rounds: drain all entries at the current min rank and merge all their non-overlapping occurrences **before** admitting newly-created pairs, rather than re-picking the global min after every single merge. Still ~O(n log n), but canonical even when the invariant fails.

### Regression test to fold in
I pushed a differential test that builds a real `BPETokenizer` from the on-disk Gemma merge table and fuzzes `bpe()` against a verbatim copy of the canonical algorithm (incl. the 11k pathological case): branch **`test/bpe-differential-regression`**. It's green on `main` and reproduces exactly the failure above on this PR's head (6 failures, all whitespace-run cases). Please pull it into this PR so the byte-identical contract is locked once the fix lands.

RaajeevChandran:
Fix: process the heap in true rank-rounds. Pop the min to fix the round's rank, drain all same-rank entries (since the heap is keyed (rank, left) they come out left-to-right for free), merge every non-overlapping occurrence, and defer pairs created during the round — pushing them only after the round completes so a newly-formed lower-rank pair can't preempt it. Still O(n log n), but canonical even when the invariant fails.

I pulled your test/bpe-differential-regression suite into this PR (plus the VMLXHub test-target dep, and added a ~/MLXModels/.../gemma-4-12B-it-qat-JANG_4M candidate path so it runs on more dev machines). Validated against the real 514k-entry Gemma merge table:

  - 32-tab repro: now 16+16, matches reference ✅
  - 122,592 repeated-run cases (every single-char symbol × 16 lengths): 0 divergences
  - 6,000 fuzzed inputs: 0 divergences
  - 11k space-free pre-token: byte-identical, ~6 ms vs ~44 ms reference — perf win intact

jjang-ai:
Verified `27ddd55e` (rank-rounds) — fix confirmed. Ran the differential test against this head:

- **0 divergences** — the whitespace runs that broke before (`\t`/space/newline at lengths 16/33/64/129) now all match the canonical algorithm, including the `\t`×32 repro.
- 4,148 random fuzzed inputs still byte-identical.
- 11k-char space-free pre-token: **8.9 ms** — the speedup is fully retained.

LGTM. The differential regression test is on branch `test/bpe-differential-regression` (green on `main`, red on `c8950777`, green on `27ddd55e`) — recommend folding it into this PR so the byte-identical contract stays locked.

PR #81, opened 2026-06-21
Fix VMLXHub Swift 6 concurrency build

Heyvhuang:
## Summary
- Compile vendored swift-transformers targets in Swift 5 language mode while keeping strict concurrency enabled.
- Fixes Swift 6/Xcode beta builds failing in VMLXHub/HubApi with a sendability diagnostic for repo access.

## Verification
- Used by PeykaCore via fork revision f647dc959f486c2322d863a3dceb3721b403c84a.
- Clean PeykaCore focused test build reached and passed the cloud employee/provider tests after pinning this revision.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -106,7 +106,8 @@ import PackageDescription
 #endif
 
 let transformersSwiftSettings: [SwiftSetting] = [
-    .enableExperimentalFeature("StrictConcurrency")
+    .enableExperimentalFeature("StrictConcurrency"),
+    .swiftLanguageMode(.v5),
 ]
 
 let mlxLMCommonExcludedFiles: [String] = [


jjang-ai:
Same underlying report as #71/#151, but I went with the targeted fix instead — landed in #257, which marks `Hub.Repo` and `Hub.RepoType` `Sendable`.

The reason for preferring that: adding `.swiftLanguageMode(.v5)` to `transformersSwiftSettings` takes the **entire** vendored transformers target out of Swift 6 checking. That silences the one over-report, but it also silences every future concurrency defect in that target, and the `.enableExperimentalFeature("StrictConcurrency")` line directly above it stops meaning much. The conformance fixes the actual cause in two words and leaves the checking in place.

Also worth noting for both approaches: neither is load-bearing on a current toolchain. The package builds clean on Apple Swift 6.3.3 today — the region-isolation over-report was a 6.0/6.1-era issue, which is what the comment above `mlxLMCommonSwiftSettings` in `Package.swift` describes. If you are still hitting a concrete build error on a specific toolchain, please post the exact error and version and I will reopen — a real failure would change the calculus here.

Closing in favour of #257. Thanks for chasing it.

PR #83, opened 2026-06-24
fix mistral3 vlm crash on flat rope_theta configs

RaajeevChandran:
- Fixed a fatal crash at model-load time when loading Mistral3 / Ministral VLM bundles whose `text_config` specifies `rope_theta` as a flat top-level field instead of a nested `rope_parameters` dict.
- Removed the hard `fatalError` in the language attention initializer that aborted the entire app when `rope_parameters` was absent.
- Fell back to the top-level `rope_theta` value (already parsed by the config, with a sane default) when the nested `rope_parameters` dict is not present, keeping the nested layout authoritative when it exists.
- Wired the RoPE `traditional` flag to the parsed `rope_traditional` config value rather than a hardcoded `false`.
- Resolved the production crash reported for standard Mistral and Ministral config layouts that previously could not be loaded.

### DIFF
--- a/Libraries/MLXVLM/Models/Mistral3.swift
+++ b/Libraries/MLXVLM/Models/Mistral3.swift
@@ -335,16 +335,16 @@ private enum Language {
             self._wv.wrappedValue = Linear(dim, nKVHeads * headDim, bias: false)
             self._wo.wrappedValue = Linear(nHeads * headDim, dim, bias: false)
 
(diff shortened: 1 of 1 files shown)


jjang-ai:
Superseded by #84 (merged), which includes this rope fix plus the JANGTQ-twin rope crash and the split-processor-config blocker that left these models unloadable. Live-proven on mlx-community/Mistral-Small-3.1-24B. Closing in favor of #84 — thanks for the original catch.

PR #91, opened 2026-06-27
Trust generation_config.json eos over the hard-coded end-token fallback (fixes harmony / gpt-oss empty answers)

rcfa:
## Summary

`resolveStopSequences` unconditionally augments every model's stop set with the global
`commonEndTokenStrings` list. For a model that **authoritatively declares** its stop set in
`generation_config.json` (the HF-canonical, mlx-lm-honored generation-time eos list), that can only
**add stops the author deliberately omitted**.

Concretely it breaks OpenAI **harmony** (gpt-oss): the blanket force-adds `<|end|>` (200007), which
gpt-oss's `generation_config.json` deliberately excludes from eos — in harmony `<|end|>` is the
**analysis→final channel separator**, not a stop. Generation halts right after the `analysis` channel
and the visible answer is **empty**.

## Fix

Gate the blanket on `modelConfiguration.generationDefaults?.eosTokenIds` being non-empty:

- **Declared** (`generation_config.json` has `eos_token_id`) → trust it, **skip the blanket**.
- **Undeclared** (no `generation_config.json`) → keep the blanket as a safety net.

`generationDefaults` is `nil` for callers that didn't load one, so the gate is **conservative by
default** — the blanket is skipped only on *positive evidence* of a declaration. ~15 lines, one file.

## Review follow-ups

All three of your asks are in, and #1 was worth doing for its own sake — writing the fixture is what
turned "I claim harmony declares 200007-less eos" into something the suite enforces.

### 1. The gpt-oss claim, proven in-repo

`harmonyDeclaredEOSKeepsChannelSeparatorOutOfStops` decodes harmony's **actual**
`generation_config.json` eos (`[200002, 199999, 200012]` — no 200007) through the real
`GenerationConfigFile`/`JSONDecoder.json5()` path, feeds it to `resolveStopSequences`, and asserts
200007 appears in **neither** `tokenIDs` nor `textStopStrings`.

The tokenizer stub deliberately **can** resolve `<|end|>` → 200007. That matters: it proves the *gate*
is what keeps the token out of the stop set, not an unresolvable token quietly falling on the floor.
Delete the gate and the test goes red for the right reason.

### 2. The other side of the gate

`declaredEOSModelStillStops` pins the failure mode that would actually hurt — a Qwen-style bundle
(declared eos `[151645, 151643]`, `<|im_end|>` in `extraEOSTokens`) **still stops**. The gate
suppresses only the hard-coded blanket; it must never suppress the declaration itself or the
tokenizer eos. This one passes both before and after the fix, as a fence should.

### 3. Log when the blanket is skipped

One `logger.debug` naming the declared ids, via the same `os.Logger` house style as `Evaluate.swift`:

```
generation_config.json declares eos [200002, 199999, 200012]; skipping the
commonEndTokenStrings blanket for gpt-oss-20b
```

Both ids and model name are `privacy: .public` — they're bundle metadata, not user data, and
redacting them would defeat the point.

### Verified red→green, not just green

Reverting **only** `StopStringMatcher.swift` to pristine `main` (tests untouched) fails the suite on
exactly one assertion — `!resolved.tokenIDs.contains(200007)`, the behavioural one — and restoring the
gate passes 2/2. The Qwen fence passes in **both** states, as it should. So the tests fail on precisely
what the fix changes and nothing else, and the red state also confirms the stub really does resolve
`<|end|>` → 200007 (an inert stub couldn't have gone red).

## Validation (on-device, gsm8k greedy; all `finish=stop`, bounded token counts)

| Model | `generation_config` eos | blanket | result |
|---|---|---|---|
| Qwen (Holo3-35B) | `[248046, 248044]` | skipped | stop ✓ |
| gpt-oss-20b | `[200002, 199999]` | skipped | harmony restored **0% → 100%** ✓ |
| gemma-4-12B | `[1, 106, 50]` | skipped | stop on `<turn\|>` (106) ✓ |
| Phi-3-medium | *(none)* | kept | stop via tokenizer eos ✓ |

This aligns vmlx-swift with mlx-lm's "trust `generation_config.json` eos" behavior: the hard-coded
end-token list becomes a *fallback for silence*, never an override of a model's own declared stop set.

### DIFF
--- a/Libraries/MLXLMCommon/StopStringMatcher.swift
+++ b/Libraries/MLXLMCommon/StopStringMatcher.swift
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: MIT
 
 import Foundation
+import os
 
 /// Streaming text-level stop-sequence matcher.
 ///
@@ -180,6 +181,11 @@ func mergeStopStrings(_ explicit: [String], _ defaults: [String]) -> [String] {
     return merged
 }
 
+/// Diagnostics for stop-sequence resolution. `generation_config.json` becoming load-bearing
+/// (see the blanket gate below) means a bundle that declares a WRONG eos now silently never
+/// stops; this line turns that from an afternoon into a five-second diagnosis.
+private let stopSequenceLogger = Logger(subsystem: "vmlx", category: "StopSequences")
+
 func resolveStopSequences(
     modelConfiguration: ModelConfiguration,
     tokenizer: Tokenizer,
@@ -202,13 +208,30 @@ func resolveStopSequences(
             textStopStrings: &textStopStrings,
             allowExactTextFallback: true)
     }
-    for token in commonEndTokenStrings {
-        resolveStopTokenString(
-            token,
-            tokenizer: tokenizer,
-            tokenIDs: &tokenIDs,
-            textStopStrings: &textStopStrings,
-            allowExactTextFallback: false)
+    // The hard-coded `commonEndTokenStrings` fallback is applied ONLY to models that did
+    // NOT authoritatively declare their stop set in `generation_config.json`. When a model
+    // ships that file with an `eos_token_id` (the HF-canonical, mlx-lm-honored generation-time
+    // stop list), the declaration is complete and authoritative — augmenting it with hard-coded
+    // end strings can only ADD stops the model deliberately omitted. That over-reach is exactly
+    // what made gpt-oss (OpenAI harmony) halt at the `<|end|>` channel separator that its own
+    // `generation_config.json` correctly excludes from eos. So: trust the declaration when present;
+    // keep the blanket only as a safety net for under-declared conversions that ship no
+    // `generation_config.json` eos at all (e.g. some Phi-3 builds whose only `<|end|>` signal is
+    // the tokenizer eos). `generationDefaults` is nil for any caller that didn't load one, so this
+    // is conservative by default — the blanket is skipped only on positive evidence of a declaration.
+    let declaredEOS = modelConfiguration.generationDefaults?.eosTokenIds?.values ?? []
+    if declaredEOS.isEmpty {
+        for token in commonEndTokenStrings {
+            resolveStopTokenString(
+                token,
+                tokenizer: tokenizer,
+                tokenIDs: &tokenIDs,
+                textStopStrings: &textStopStrings,
+                allowExactTextFallback: false)
+        }
+    } else {
+        stopSequenceLogger.debug(
+            "generation_config.json declares eos \(declaredEOS, privacy: .public); skipping the commonEndTokenStrings blanket for \(modelConfiguration.name, privacy: .public)")
     }
 
     return ResolvedStopSequences(
--- a/Tests/MLXLMTests/GenerationConfigEOSGateTests.swift
+++ b/Tests/MLXLMTests/GenerationConfigEOSGateTests.swift
@@ -1,0 +1,74 @@
+import Foundation
+import Testing
+
+@testable import MLXLMCommon
+
+/// Pins the `generation_config.json` eos-gate in `resolveStopSequences`: the hard-coded
+/// `commonEndTokenStrings` blanket is applied ONLY when the model did not authoritatively declare its
+/// stop set. A model that ships `generation_config.json` with an `eos_token_id` gets its declaration
+/// trusted verbatim — the blanket is skipped, so it can't inject stops the model deliberately omitted.
+///
+/// Own file (not appended to `StopStringMatcherTests`) on purpose: this and the gemma `<end_of_turn>`
+/// tests are two independent upstream PRs, so each patch must apply to a pristine tree by itself. See
+/// the note in `Gemma4StaleEOSTests`.
+@Suite("generation_config.json eos-gate")
+struct GenerationConfigEOSGateTests {
+
+    /// Resolves a fixed token→id map; everything else is empty. `eosToken` (optional) lets a test give
+    /// the tokenizer its own eos so `resolveStopSequences` inserts it (the model's real stop).
+    private struct MapTokenizer: Tokenizer {
+        var map: [String: Int]
+        var eosToken: String?
+        var bosToken: String? { nil }
+        var unknownToken: String? { nil }
+
+        func encode(text: String, addSpecialTokens: Bool) -> [Int] { [] }
+        func decode(tokenIds: [Int], skipSpecialTokens: Bool) -> String { "" }
+        func convertTokenToId(_ token: String) -> Int? { map[token] }
+        func convertIdToToken(_ id: Int) -> String? { map.first { $0.value == id }?.key }
+        func applyChatTemplate(
+            messages: [[String: any Sendable]],
+            tools: [[String: any Sendable]]?,
+            additionalContext: [String: any Sendable]?
+        ) throws -> [Int] { [] }
+    }
+
+    private static let endToken = 200007      // harmony `<|end|>` — a channel separator, NOT a stop
+    private static let imEndToken = 151645    // Qwen `<|im_end|>` — a genuine turn-end stop
+    private static let tokenizer = MapTokenizer(map: ["<|end|>": endToken, "<|im_end|>": imEndToken])
+
+    private func config(name: String, declaredEOS: [Int]?) -> ModelConfiguration {
+        ModelConfiguration(
+            id: name,
+            generationDefaults: declaredEOS.map { GenerationConfigFile(eosTokenIds: IntOrIntArray($0)) })
+    }
+
+    /// The motivating case (gpt-oss / OpenAI harmony). `generation_config.json` declares eos WITHOUT
+    /// `<|end|>` (200007); the blanket would otherwise inject it, halting at the channel separator.
+    /// With the gate, `<|end|>` must NOT appear in the resolved stops.
+    @Test("harmony: declared eos excluding <|end|> → blanket skipped, 200007 not a stop")
+    func harmonyDeclaredEOSSkipsBlanket() {
+        let cfg = config(name: "gpt-oss-harmony", declaredEOS: [199_999, 200_002, 200_012])
+        let resolved = resolveStopSequences(modelConfiguration: cfg, tokenizer: Self.tokenizer)
+        #expect(!resolved.tokenIDs.contains(Self.endToken))
+    }
+
+    /// The before/after: an UNDER-declared model (no `generation_config.json` eos) still gets the
+    /// blanket, so `<|end|>` IS resolved as a stop. This is exactly the state the gate flips.
+    @Test("no declaration → blanket runs, <|end|> IS a stop")
+    func underDeclaredKeepsBlanket() {
+        let cfg = config(name: "under-declared", declaredEOS: nil)
+        let resolved = resolveStopSequences(modelConfiguration: cfg, tokenizer: Self.tokenizer)
+        #expect(resolved.tokenIDs.contains(Self.endToken))
+    }
+
+    /// Regression on the other side of the gate: a model whose declared eos IS `<|im_end|>` must still
+    /// stop on it. Skipping the blanket must not drop the model's own declared stop.
+    @Test("Qwen: declared <|im_end|> eos still stops")
+    func declaredQwenEOSStillStops() {
+        let cfg = config(name: "qwen-imend", declaredEOS: [Self.imEndToken])
+        let tok = MapTokenizer(map: ["<|im_end|>": Self.imEndToken], eosToken: "<|im_end|>")
+        let resolved = resolveStopSequences(modelConfiguration: cfg, tokenizer: tok)
+        #expect(resolved.tokenIDs.contains(Self.imEndToken))
+    }
+}
--- a/Tests/MLXLMTests/StopStringMatcherTests.swift
+++ b/Tests/MLXLMTests/StopStringMatcherTests.swift
@@ -341,4 +341,96 @@ struct StopStringMatcherTests {
         emitted += matcher.flush()
         #expect(emitted == input)
     }
+
+    // MARK: - generation_config.json eos is authoritative (harmony / gpt-oss)
+
+    /// The motivating case, proven rather than asserted. gpt-oss (OpenAI harmony) declares
+    /// `eos_token_id: [200002 <|return|>, 199999, 200012 <|call|>]` and deliberately EXCLUDES
+    /// 200007 `<|end|>`, which is a CHANNEL SEPARATOR, not a stop: the model emits
+    /// `<|end|>` to close the `analysis` channel before opening `final`. The hard-coded
+    /// `commonEndTokenStrings` blanket used to inject `<|end|>` anyway, so generation halted at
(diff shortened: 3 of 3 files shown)


jjang-ai:
I came into this one expecting to push back — gating the `commonEndTokenStrings` blanket changes stop resolution for *every* model that ships a `generation_config.json`, and that blanket is the safety net for the whole zoo (`<|im_end|>`, `<|eot_id|>`, DeepSeek's wide-pipe spelling, …). So I measured the blast radius instead of arguing about it.

Across every model bundle on my machine, for the ones that declare an `eos_token_id`, here's **everything the blanket currently adds beyond the declaration** — i.e. exactly what your PR would switch off:

```
bundles with a declared eos : 27
  -> blanket adds nothing    : 25   (your change is a pure no-op for these)
  -> blanket adds something  :  2

  ZAYA1-VL-8B-JANGTQ4   eos=[262143]           blanket ALSO adds <end_of_turn> (106)
  HY-MT1.5-7B-mlx-8Bit  eos=[127960, 127967]   blanket ALSO adds <|endoftext|> (127957)
```

Zero regressions. And the two hits actually argue *for* you: ZAYA declares one eos and the blanket force-injects **Gemma's** turn-end token into it, which is a latent false-stop, not protection. That's the same over-reach you're describing with harmony, just with a different victim.

I also checked the wiring, since the gate depends on `generationDefaults` being populated consistently: `LLMModelFactory` rebuilds the `ModelConfiguration` with `generationDefaults: generationConfig` before it reaches the `ModelContext` (`LLMModelFactory.swift:1925-1934`), and `VLMModelFactory` does the same. So there's no split where the registry path and the local-bundle path disagree — the gate behaves the same on both. Good.

So: **the reasoning is sound and I'm inclined to take it.** Three things I want before it lands, none of them large:

1. **Prove the gpt-oss claim in-repo.** Right now it's the motivating case and it's asserted, not shown. A test fixture with harmony's actual `generation_config.json` (eos excluding `<|end|>` / 200007) asserting that `resolveStopSequences` no longer contains 200007 — and, ideally, the before/after of the empty-answer symptom. Everything else in this PR I could verify from my own bundles; that one I couldn't.
2. **A regression test on the other side of the gate**: a declared-eos model (pick a Qwen: eos declared, `<|im_end|>` in it) still stops. That's the failure mode that would hurt, and it should be pinned so nobody has to redo my survey.
3. **Log when the blanket is skipped.** This is the real residual risk and it's worth naming in the code: the change makes `generation_config.json` **load-bearing**. That's correct for well-formed bundles, but a lot of community quant re-bakes copy `generation_config.json` from a base model and get it wrong, and for those the blanket was quietly saving us. If a bundle declares an eos that doesn't match its own chat template, the new behaviour is "model never stops" — which is a miserable thing to debug from a bug report. One `logger.debug` line naming the declared ids makes it a five-second diagnosis instead of an afternoon.

Your `generationDefaults == nil → keep the blanket` default is the right conservative choice and does most of the work here. Nice piece of reasoning — the harmony `<|end|>`-is-a-channel-separator-not-a-stop detail is exactly the kind of thing that gets misread as "the model is broken."

rcfa:
Thanks for measuring the blast radius — that's the analysis I couldn't do from my end. All three
asks are in:

1. **In-repo proof of the gpt-oss claim** — `GenerationConfigEOSGateTests` (its own file, so the
   patch applies standalone): a harmony-style `generation_config.json` that declares eos *without*
   `<|end|>`/200007 asserts `resolveStopSequences` no longer contains 200007; the before/after case
   (an under-declared model with no declaration) still gets the blanket, so `<|end|>` IS resolved.
2. **The other side of the gate** — a declared-`<|im_end|>` (Qwen) model still stops on it, so
   skipping the blanket never drops a model's own declared stop.
3. **The log line** — when the blanket is skipped, `stopSequenceLogger.debug` names the declared ids
   and the model, so a mis-baked `generation_config.json` ("model never stops") is a five-second
   diagnosis. Your `generationDefaults == nil → keep the blanket` point is exactly the conservative
   default it uses.

All three tests are pure logic (no MLX), sub-millisecond.

rcfa:
**A/B: patched vs unpatched, seeded, same items**

Two binaries differing *only* in this patch (all other local patches applied to both), `--seed 1234`, gsm8k prompts.

### The fix — `mlx-community/gpt-oss-20b-MXFP4-Q8` (harmony), n=10

| | unpatched | patched |
|---|---|---|
| item 1 | `text=0ch` reasoning=299ch | `text=175ch` reasoning=299ch |
| item 2 | `text=0ch` reasoning=1298ch | `text=194ch` reasoning=1298ch |
| item 3 | `text=0ch` reasoning=652ch | `text=356ch` reasoning=652ch |
| item 4 | `text=0ch` reasoning=1369ch | `text=426ch` reasoning=1369ch |

Unpatched, the answer channel is empty on **every** item: the model reasons normally, then halts at the `<|end|>` channel separator before opening `final`. `finish=stop` in both arms — it is not running out of budget, it is being told to stop at a token its own `generation_config.json` deliberately excludes from `eos_token_id` (`[200002, 199999, 200012]`; `<|end|>` = 200007 is absent).

**The reasoning byte-counts are identical across arms** (299 / 1298 / 652 / 1369). The model computes exactly the same thing — the patch only stops the runtime truncating it before it emits the answer.

### No regression — non-harmony models that also declare eos

| model | arch / eos | per-item outputs identical | finish |
|---|---|---|---|
| Nanbeige4.2-3B | `nanbeige`, `166101` | **20/20** | `stop` on every item, both arms |
| Qwythos-9B | `qwen3_5`, `[248046, 248046]` | **20/20** | `stop` on every item, both arms |

### Scope note

20 of the 22 models we regularly run declare `generation_config` eos, so this branch is taken widely — it is not a narrow path. What the gate buys is not a small blast radius but a principled one: behaviour changes only where a model made an authoritative declaration, and there it defers to that declaration rather than overriding it. Models shipping no eos declaration run byte-identical code.

Broader supporting data (observational, not a controlled comparison): 840 scored cells across 22 models / 28 benchmarks / 9 architectures, all run with this patch applied, with no stop-token pathology observed.

jjang-ai:
Thought about this one properly, since it's the one you flagged. **I think the gate is right**, and the reasoning below is why — plus the one change I'd want before it lands.

## Why the assumption holds

The gate rests on "a bundle that declares `eos_token_id` in `generation_config.json` has declared it *completely*". That is the HF contract — `generation_config.eos_token_id` is the generation-time stop list, and it is a list precisely so a model can name several (Qwen ships `[151645, 151643]` = `<|im_end|>` + `<|endoftext|>`). Honoring it and declining to add to it is what mlx-lm does. Overriding a model's explicit declaration with a hard-coded list is the thing that needs justification, not the reverse.

And the blanket is not the only safety net left. Everything above the gate still applies unconditionally:

```swift
var tokenIDs = modelConfiguration.eosTokenIds
if let tokenizerEOS = tokenizer.eosTokenId { tokenIDs.insert(tokenizerEOS) }
…
for token in modelConfiguration.extraEOSTokens { … }   // per-model curated
```

So a family that genuinely needs a turn-end token the config omits has `extraEOSTokens` as the targeted mechanism — which is the right place for it, because it is per-model rather than applied to everyone.

## The residual risk, stated plainly

The failure this can introduce: a sloppy conversion that ships `generation_config.json` declaring only the base `<|endoftext|>` while its chat template ends turns with `<|im_end|>`, and which has no `extraEOSTokens` entry. Before, the blanket rescued it; now it runs to the token cap.

I think that is acceptable — it is bounded (a long answer, not corruption), it is diagnosable, and the alternative is the harmony bug you're fixing, where the blanket actively overrides a *correct* declaration. Trading "some malformed bundles ramble" for "some correct bundles are silently truncated" is the right direction, because the second class fails invisibly and the first is obvious the moment you see it.

## One change I'd want first

Make the new log line visible by default:

```swift
stopSequenceLogger.debug("generation_config.json declares eos …")
```

`.debug` is not persisted by default on macOS, so in exactly the scenario the comment above it describes — *"a bundle that declares a WRONG eos now silently never stops"* — the line that turns it into a five-second diagnosis will not be in the log the user sends you. `.info`, or `.notice` given it fires once per model load, would actually pay for itself. Same argument the comment already makes; it just needs the level to match the intent.

With that, I'm happy to take it.

## Verified separately

- `#224` — checked the lock discipline: `entries` is touched only in `verdict(for:)` and `record(_:for:)`, both `lock()` + `defer { unlock() }`. The annotation is accurate, not a silencer.
- `#150` — builds clean here on Swift 6.3.1, which takes the `.v6` branch. Worth noting the two are coupled: `.v6` on MLXLMCommon is what makes `#224` load-bearing, so `#224` should land first (or with it).
- `#155` — confirmed the strip is safe because the prefix is re-added downstream (`out["language_model.\(k)"]`), so unprefixed OsaurusAI bundles are a no-op and the mlx-community double-prefix goes away.
- `#95` — the collision handling is the good part. `Dictionary` iteration order is per-process seeded, so last-write-wins really would bind a different tensor run to run; resolving to the unprefixed key and then lexicographic-min is arbitrary but *fixed*, which is the property that matters.
- `#220` — the bit I checked hardest: `frames` is used as **pixel** dims (`3 * f.h * f.w`), and the pre-existing code already stamped pixel dims (`THW(1, pixels.dim(2), pixels.dim(3))`), so the meaning of `frames` in this model is unchanged — each image just carries its own instead of the batch's. Ragged path is gated on `ndim == 1`, so the uniform path is untouched.

jjang-ai:
Followed my own "residual risk" hand-wave with an actual measurement, because it turns out to be non-zero on a real library.

## Measured: 3 of 76 local bundles would lose their turn-end stop

Scanned every bundle on this machine that ships `generation_config.json` with an `eos_token_id`, resolved each `commonEndTokenStrings` entry to its id via `added_tokens_decoder` / `tokenizer.json`, and checked whether a token the bundle's **own chat template emits** is absent from its declared eos:

```
analyzed=76  safe=73  AT-RISK=3

JANGQ-AI/LFM2.5-230M-MXFP8      eos_ids=[124900]  template emits <|im_end|>, not in eos
OsaurusAI/VibeThinker-3B-MXFP8  eos_ids=[151643]  template emits <|im_end|>, not in eos
OsaurusAI/VibeThinker-3B-JANG_4M eos_ids=[151643] template emits <|im_end|>, not in eos
```

VibeThinker is the textbook case: `151643` is `<|endoftext|>`, while its Qwen2.5-style template ends turns with `<|im_end|>` (`151645`). Today the blanket rescues it. After this change it declares an eos, so the blanket is skipped, and nothing stops it at the turn boundary — it runs to the token cap.

## `extraEOSTokens` does not cover them

I suggested that as the mitigation in my earlier comment; it isn't one for these. It is populated only for **registry** entries (Phi `<|end|>`, Gemma `<end_of_turn>` — `LLMModelFactory.swift:859+`, `VLMModelFactory.swift:320+`). Bundles loaded from a local directory get `extraEOSTokens: []` by default (`Downloader.swift:122`, and the parameter defaults to `[]` in both `ModelConfiguration` inits). Since these three are local-path loads, there is no per-model net behind the blanket.

So the 73/76 result is genuinely reassuring about the *direction*, and the 3 are genuinely a regression — both things are true, and I would rather hand you the number than an adjective.

## What I would do

None of this changes my view that trusting the declaration is correct. It just means the blanket is currently load-bearing for under-declared conversions, so I would want one of:

1. **Union instead of replace, minus the known-harmful entry.** Keep the blanket, but drop `<|end|>` when `generation_config` declares an eos that excludes it. Narrowest possible fix for the harmony bug; leaves every other family exactly as it is today.
2. **Keep this gate and add `extraEOSTokens` for the affected families**, so the net exists where it is needed rather than for everyone. Cleanest end state, but it needs the affected bundles enumerated — the scan above is that list for my machine, not for the world.
3. **Gate on completeness rather than presence**: skip the blanket only for tokens the template does not emit. More machinery than this deserves.

(1) is the smallest change that fixes gpt-oss without moving anyone else. (2) is where I would want to end up. Happy either way — and if you have a wider bundle corpus in your harness, re-running that scan against it would be the thing that settles which.

jjang-ai:
Turned the three at-risk bundles into a test rather than leaving them as a paragraph you have to take my word for — `Tests/MLXLMTests/UnderDeclaredEOSStopTests.swift`, opened separately so it applies to a pristine tree and does not collide with this branch.

It builds the VibeThinker shape directly: a `generation_config.json` declaring only `151643` (`<|endoftext|>`) against a Qwen-ish tokenizer that also carries `151645` (`<|im_end|>`), then asserts `resolveStopSequences` still yields `<|im_end|>`.

- **On `main` today it passes** — the blanket supplies it.
- **Under this PR it fails**, because the bundle declares an eos, so the blanket is skipped and the token its own chat template ends every turn with is no longer a stop.

That is the trade stated as an executable fact, and it means whichever way this lands, the decision is deliberate and the regression can't reappear silently later.

The suite also pins the two things that must *not* move: a bundle declaring nothing keeps the blanket (that path never depended on the gate), and the tokenizer's own eos is present regardless — the one stop that survives any resolution of this.

If you take option (1) from my earlier comment (keep the blanket, drop only `<|end|>` when a declared eos excludes it), all three cases stay green and gpt-oss is still fixed. If you prefer this PR's stronger gate, the test tells you exactly which bundles need an `extraEOSTokens` entry before it lands — and it will keep telling you when someone adds a fourth.

Happy to write whichever variant you'd rather have; the diagnosis is yours, I only measured the blast radius.

jjang-ai:
The diagnosis here is correct and worth landing — thank you. `<|end|>` really is a channel
separator in harmony, not a stop, and the blanket really was injecting it.

The gate as written is too broad, though, and I can show it rather than argue it. Skipping
`commonEndTokenStrings` for **any** bundle that declares an eos also strips stops from bundles that
declare an *incomplete* one. `UnderDeclaredEOSStopTests` (already on main) fails against this
branch:

```
✘ underDeclaredEOSStillStopsAtImEnd() — Expectation failed:
  (resolved.tokenIDs → [151643]).contains(Self.imEnd → 151645)
```

Two installed bundles have exactly that shape — `VibeThinker-3B-MXFP8` and
`VibeThinker-3B-JANG_4M` declare only `<|endoftext|>` (which is also their tokenizer eos) while
their template ends every turn with `<|im_end|>`. With the blanket skipped they lose their only
turn boundary and run to the token cap.

I also tried a template-terminator probe to separate the two cases automatically; it does not
work. Harmony renders **historical** assistant turns ending in `<|end|>` as well — only
*generation* ends at `<|return|>` — so any tail-of-render signal keeps `<|end|>` and the fix dies.

#254 keeps your fix with a narrower trigger: suppress `<|end|>` only when the vocabulary is
harmony-shaped (carries `<|channel|>`, `<|message|>` and `<|return|>`) **and** the model's own
declaration excludes it. Detected from the vocabulary rather than `reasoningParserName` so an
unstamped local conversion is still covered. Across the 86 installed bundles, 0 are harmony-shaped
and 0 carry `<|end|>`, so nothing else moves — A/B confirmed byte-identical on LFM2.5-VL, Bonsai
27b and VibeThinker.

Your test intent is carried over, including the `<|end|>`-must-not-be-a-stop assertion, plus the
Phi-side non-regression. Closing in favour of #254 with the diagnosis credited to you.

PR #92, opened 2026-06-27
Add GLM-4.6V-Flash (glm4v) and GLM-4.5V (glm4v_moe) vision-language models

rcfa:
## Summary

Adds two GLM vision-language model types to `MLXVLM`, hand-ported from the established Python
`mlx_vlm` implementations:

- **`glm4v`** — GLM-4.6V-Flash (dense GLM-4 decoder)
- **`glm4v_moe`** — GLM-4.5V (DeepSeek-V3-style MoE decoder)

Registered in `VLMModelFactory`. Thank you for the review — it was right about the shape of the risk
even where it was wrong about the specific line, and chasing it down turned up a worse bug underneath.

## The two interpolation findings, because they invert your blocking call

**Padding mode — I think this one is withdrawn.** You asked me to check whether the GLM reference passes
`padding_mode="border"`. It does, explicitly, in both architectures — HF transformers
`modeling_glm4v.py` and `modeling_glm4v_moe.py`, `Glm4vVisionEmbeddings.forward`:

```python
F.grid_sample(pos_embed_2d, grid, mode=self.interpolated_method, align_corners=False, padding_mode="border")
```

So the clamp you flagged at `Glm4v.swift:436` / `Glm4vMoe.swift:1951` matches the reference. Per your
note ("if it does, I'll withdraw this entirely"), I believe that resolves the blocking item — but the
same two lines turned up the real problem.

**Interpolation kernel — this is the actual bug, and it's broader than the padding one.** That same call
hardcodes `self.interpolated_method = "bicubic"` (literal, in `__init__`, both files). **We shipped
bilinear.** Your instinct was right and then some: padding mode only perturbs out-of-bounds border
samples on upsampled grids, but bilinear-vs-bicubic diverges at **every** non-grid-aligned sample — the
whole interior of any position grid that isn't the learned 24×24. Same "fluent, coherent, subtly wrong"
signature you described, larger blast radius.

And you called our own validation exactly: the PR used to claim "mmmu 55/58% ≫ 25% random → the resample
is correct." That reasoning is the trap — a wrong interpolation still scores plausibly on mmmu. I've
struck that claim.

## What changed

**Fixed the kernel, in one place.** New `Libraries/MLXVLM/Models/Glm4Vision.swift` holds a corrected
`gridSampleBicubic` — Keys cubic convolution, `A = -0.75`, matching PyTorch's
`get_cubic_upsample_coefficients` — with `align_corners=False` and border clamping. Both files' vision
towers now call it; the two byte-identical `gridSampleBilinear` copies are gone. That directly answers
"I had to report the bug at two line numbers": the numerically sensitive function now exists once.

**Scope I want to be straight about:** this shares the *interpolation*, not yet the entire ~465-line
vision tower. The rest of the tower (patch embed, encoder blocks, M-RoPE) is still duplicated. I'd rather
land the correctness fix + the one-copy-of-grid_sample now and do the full tower extraction as a
follow-up than bundle a large mechanical refactor into the same change as a numerical fix. Say the word
if you'd prefer it all at once.

## Tests

`Tests/MLXLMTests/Glm4VisionTests.swift` (the target that can `@testable import MLXVLM`):

- **Independent ground truth** (holds for any cubic `A`, so it survives even a wrong kernel constant):
  sampling a pixel center returns that pixel exactly; a constant field stays constant.
- **Border padding**: an out-of-bounds sample clamps to the edge value, not toward zero.
- **`A`-sensitive interior**: three fractional samples pinned to a numpy reimplementation of PyTorch's
  Keys `A = -0.75` coefficients. I could not use an independent library oracle offline — scipy's bicubic
  uses spline prefiltering, not Keys convolution, so it won't match `grid_sample`. **A torch cross-check
  of those three numbers would be welcome**; if they're off, the kernel constant is the thing to look at.

## Also done in this PR

- **Deleted the dead processors.** `Glm4vProcessor` and `Glm4vMoeProcessor` (~217 lines) are gone —
  the factory registers `GlmOcrProcessor` for both keys, so as you noted they were unreachable. That
  also retires the `images[0]`-on-empty trap you flagged, since it lived in the dead
  `Glm4vProcessor.preprocess`.

## Still open from your review — deferred deliberately, not overlooked

- **De-dup the rest of the vision tower.** This PR shares the interpolation (the numerically sensitive
  part); the remaining ~450 lines of tower (patch embed, encoder blocks, M-RoPE) are still duplicated.
  Follow-up, unless you'd rather it land here.
- **The remaining trap guards** (non-merge-divisible grid dims, zero merge, video-only indexing) live
  in the **non-throwing** vision-forward path, so "a load error, not a trap" means threading a throwing
  path (or a loader-boundary check) through that code — a bigger change than a mechanical guard, and I'd
  rather not fold it into the same PR as a numerical fix. I did not add `precondition`s, because a
  precondition is still a trap and wouldn't meet the bar you set. Flagging as a scoped follow-up. (Note:
  these divisibility invariants are currently guaranteed upstream by the processor's `targetSize`
  factor `patchSize * mergeSize`, so they're reachable only from an inconsistent bundle/config.)
- **Template resolves with tools enabled.** Needs an on-device GLM-4.5V load, which would contend with
  a benchmark suite currently holding the GPU and corrupt its recorded timings. Deferred; I'll confirm
  and report it as a comment rather than block on it here.

## Notes

- The `argPartition` tie note for the MoE routing is understood — same class as the temp-0
  nondeterminism I'm chasing on #96; I won't treat "same output twice" as determinism proof here either.

### DIFF
--- a/Libraries/MLXVLM/Models/Glm4Vision.swift
+++ b/Libraries/MLXVLM/Models/Glm4Vision.swift
@@ -1,0 +1,104 @@
+//
+//  Glm4Vision.swift
+//  mlx-swift-lm
+//
+//  Shared vision helpers for the GLM-4V family (glm4v, glm4v_moe). These two
+//  ports carried byte-identical copies of the position-embedding interpolation;
+//  the numerically sensitive part lives here so there is exactly one copy to be
+//  correct.
+//
+
+import Foundation
+import MLX
+
+enum Glm4SharedVision {
+
+    /// `grid_sample` with `mode="bicubic"`, `align_corners=False`,
+    /// `padding_mode="border"` — matching `torch.nn.functional.grid_sample` as GLM-4V's
+    /// `Glm4vVisionEmbeddings.forward` invokes it (`interpolated_method = "bicubic"`, hardcoded
+    /// in HF transformers `modeling_glm4v.py` / `modeling_glm4v_moe.py`).
+    ///
+    /// The earlier port used BILINEAR here. That diverges from the reference at every
+    /// non-grid-aligned sample — i.e. across the whole interior of any upsampled position grid,
+    /// not just the borders — so it produced subtly wrong position embeddings on any image whose
+    /// patch grid differs from the learned 24×24. (The `padding_mode="border"` clamp below is
+    /// correct and matches the reference; only the interpolation kernel was wrong.)
+    ///
+    /// - Parameters:
+    ///   - x: source, `(B, H, W, C)`.
+    ///   - grid: sample coordinates, `(B, gN, gM, 2)`, last axis `(gx, gy)` in `[-1, 1]`.
+    /// - Returns: `(B, gN, gM, C)`.
+    static func gridSampleBicubic(_ x: MLXArray, grid: MLXArray) -> MLXArray {
+        let B = x.dim(0)
+        let H = x.dim(1)
+        let W = x.dim(2)
+        let C = x.dim(3)
+        let gN = grid.dim(1)
+        let gM = grid.dim(2)
+
+        // Un-normalize from [-1, 1] to pixel coordinates (align_corners=False).
+        let gx = grid[.ellipsis, 0]  // (B, gN, gM)
+        let gy = grid[.ellipsis, 1]
+        let ix = ((gx + 1) * Float(W) - 1) / 2
+        let iy = ((gy + 1) * Float(H) - 1) / 2
+
+        let ixFloor = floor(ix)
+        let iyFloor = floor(iy)
+        let tx = ix - ixFloor  // fractional part in [0, 1)
+        let ty = iy - iyFloor
+
+        // Cubic-convolution weights for the four taps at offsets {-1, 0, 1, 2}.
+        let wx = cubicWeights(tx)
+        let wy = cubicWeights(ty)
+
+        let xFlat = x.reshaped(B, H * W, C)
+
+        func clampX(_ v: MLXArray) -> MLXArray {
+            clip(v, min: MLXArray(Int32(0)), max: MLXArray(Int32(W - 1))).asType(.int32)
+        }
+        func clampY(_ v: MLXArray) -> MLXArray {
+            clip(v, min: MLXArray(Int32(0)), max: MLXArray(Int32(H - 1))).asType(.int32)
+        }
+
+        // Gather x[b, clamp(iyc), clamp(ixc)] for the whole (gN, gM) grid → (B, gN, gM, C).
+        // Border padding = clamping the tap indices to the valid range.
+        func gather(_ iyc: MLXArray, _ ixc: MLXArray) -> MLXArray {
+            let flatIdx = (iyc * Int32(W) + ixc).reshaped(B, gN * gM)
+            var slices = [MLXArray]()
+            for b in 0 ..< B {
+                let idxB = flatIdx[b]
+                let g = xFlat[b][idxB]  // (gN*gM, C)
+                slices.append(g[.newAxis])
+            }
+            return concatenated(slices, axis: 0).reshaped(B, gN, gM, C)
+        }
+
+        // Separable bicubic: interpolate the four sampled rows in x, then combine in y.
+        let offsets = [-1, 0, 1, 2]
+        var out: MLXArray?
+        for (j, dy) in offsets.enumerated() {
+            let iyc = clampY(iyFloor + Float(dy))
+            var row: MLXArray?
+            for (i, dx) in offsets.enumerated() {
+                let ixc = clampX(ixFloor + Float(dx))
+                let v = gather(iyc, ixc)
+                let w = wx[i][.ellipsis, .newAxis]  // (B, gN, gM, 1)
+                row = row.map { $0 + v * w } ?? v * w
+            }
+            let wyj = wy[j][.ellipsis, .newAxis]
+            out = out.map { $0 + row! * wyj } ?? row! * wyj
+        }
+        return out!
+    }
+
+    /// Keys cubic-convolution weights (`A = -0.75`, matching PyTorch's
+    /// `get_cubic_upsample_coefficients`) for taps at `floor-1, floor, floor+1, floor+2`,
+    /// given the fractional offset `t ∈ [0, 1)`.
+    private static func cubicWeights(_ t: MLXArray) -> [MLXArray] {
+        let a: Float = -0.75
+        // conv1 for |x| <= 1, conv2 for 1 < |x| < 2.
+        func conv1(_ x: MLXArray) -> MLXArray { ((a + 2) * x - (a + 3)) * x * x + 1 }
+        func conv2(_ x: MLXArray) -> MLXArray { ((a * x - 5 * a) * x + 8 * a) * x - 4 * a }
+        return [conv2(t + 1), conv1(t), conv1(1 - t), conv2(2 - t)]
+    }
+}
--- a/Libraries/MLXVLM/Models/Glm4v.swift
+++ b/Libraries/MLXVLM/Models/Glm4v.swift
@@ -1,0 +1,1213 @@
+//
+//  Glm4v.swift
+//  mlx-swift-lm
+//
+//  port of https://github.com/Blaizzy/mlx-vlm/tree/main/mlx_vlm/models/glm4v
+//
+//  GLM-4V family (e.g. GLM-4.6V-Flash). model_type "glm4v",
+//  architecture "Glm4vForConditionalGeneration".
+//
+//  Adapted from GlmOcr.swift. The DENSE language decoder + M-RoPE machinery is
+//  byte-for-byte equivalent to glm_ocr (glm4v language.py uses the same
+//  "sectioned_even_odd" mrope as glm_ocr). The VISION tower differs:
+//    * NO qk_norm in vision attention (glm_ocr has q_norm/k_norm)
+//    * learned absolute position embedding bilinearly resampled onto the patch
+//      grid (Glm4vVisionEmbeddings + grid_sample) — absent in glm_ocr
+//    * post_conv_layernorm after patch_embed — absent in glm_ocr
+//    * vision block RMSNorms use eps 1e-6 (hardcoded upstream), not config eps
+//    * merger contextDim = intermediate_size (glm_ocr used out_hidden*in_channels)
+//    * vision MLP hidden dim = out_hidden_size (glm_ocr used intermediate_size)
+//    * all vision Linear layers are bias-free (attention_bias=false)
+//
+
+import CoreImage
+import Foundation
+import MLX
+import MLXLMCommon
+import MLXNN
+
+// MARK: - Language
+
+private enum Glm4vLanguage {
+
+    // MARK: M-RoPE helpers
+
+    /// Interleaved rotate_half: takes even/odd indices instead of splitting in half.
+    static func rotateHalfInterleaved(_ x: MLXArray) -> MLXArray {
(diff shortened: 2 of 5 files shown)


jjang-ai:
This is the biggest PR of the six (+2,906 / −0, two architectures) and it's also the one I can't take as-is. I went through it carefully because the *modelling* is mostly right — the M-RoPE (both interleaved and non-interleaved styles), the MoE gate equations (sigmoid gating, group routing, shared expert added post-scale, `routed_scaling_factor` applied in the right place), and the config plumbing all check out. So the problems below are fixable, and none of them mean "start over."

---

### Blocking: the position-embedding interpolation clamps where it should zero

`Glm4v.swift:436-441` (and again, verbatim, at `Glm4vMoe.swift:1951-1954`):

```swift
// Clamp corner indices into valid range (border padding).
clip(v, min: MLXArray(Int32(0)), max: MLXArray(Int32(W - 1))).asType(.int32)
```

Your coordinate convention is right — this is `align_corners=False`, and you un-normalize correctly. But **`torch.nn.functional.grid_sample`'s default is `padding_mode="zeros"`, not `"border"`**, and the comment here says you knowingly chose border. Unless the GLM reference explicitly passes `padding_mode="border"` (please check — if it does, I'll withdraw this entirely and you should put that in a comment), this is a silent numerical divergence.

Why it matters, and why it's nasty: with `align_corners=False`, whenever the target patch grid is **larger** than the learned 24×24 position grid, the outermost sample coordinates land *outside* `[-1, 1]` by construction — output index 0 maps to `0.5·S/T − 0.5`, which is negative for any `T > S`. So the first and last row/column of every upsampled position embedding are out-of-bounds samples, every time. Under `zeros` they get pulled toward zero; under `border` they get the clamped edge value. Those are different embeddings.

The reason I'm calling it blocking rather than a nit: **this is invisible to a smoke test.** Identity and downsample cases don't sample out of bounds at all, so they're bit-exact either way. Only high-resolution images (upsampled grid) hit it, and the failure mode is "edge patches carry subtly wrong positions" — the model still emits fluent, coherent, plausible text. It'll read fine to a human and be quietly worse at anything spatial near an image border. A "it described my screenshot correctly" test cannot catch this. If it were the kind of bug that produces garbage I'd be less worried about merging it.

### The vision tower is copy-pasted, in full, twice

`Glm4v.swift` and `Glm4vMoe.swift` each carry their own ~465-line vision stack, `gridSampleBilinear` included. The bug above is *already* duplicated — I had to report it at two line numbers. That's the whole argument in one line: a shared `Glm4Vision` module means fixing it once instead of remembering to fix it twice, forever. Same story for the MoE gate, which is a near-copy of `GLM4MOE.swift:96-223` and overlaps `DeepseekV3.swift:257-317`.

### The two new processor structs are dead code

`Glm4vProcessor` (`:912`) and `Glm4vMoeProcessor` (`:2403`) are defined — ~200 lines — but the factory registers:

```swift
"Glm4vProcessor": create(GlmOcrProcessorConfiguration.self, GlmOcrProcessor.init),
                                                            ^^^^^^^^^^^^^^^^^^^^^
```

…the *existing* `GlmOcrProcessor`. Your own comment at `:2401` says "no new processor needed" — and it's right! So both new structs are unreachable. Either wire them up or delete them; unreachable near-duplicates of a live code path are how a fix lands in the copy nobody runs.

### Crash surface

`images[0]` on an empty array; odd / non-merge-divisible grid dims; a zero merge size; a force-unwrapped expert lookup; a video-only path that indexes unsafely. Each of these is reachable from a malformed bundle or an empty input. A bad checkpoint should surface a load error, not trap the host process — this runs inside a long-lived server, and a trap takes the user's loaded model down with it.

### No tests

Given the above, this is the part I feel most strongly about. A `grid_sample` unit test against a handful of reference values from the Python side would have caught the padding-mode divergence directly, and it's maybe 30 lines. Right now nothing in this PR would tell us if a future refactor broke it.

---

### Things you can't see from outside this repo, but that will decide whether this works in production

- **The processor path is the real template selector.** We've been bitten by VLMs that ship their chat template only in `chat_template.json` (Mistral/Pixtral did exactly this) and silently fall back to ChatML, which then leaks `<|im_start|>` into visible output the moment tools are involved. Since you're reusing `GlmOcrProcessor`, please confirm a GLM-4.5V bundle's template actually resolves — and specifically confirm it with **tools enabled**, which is where that failure surfaces.
- **`usage.prompt_tokens` does not count vision tokens** in our server. If you're sanity-checking that an image made it through by looking at token counts, you'll see a suspiciously small number and it means nothing. Don't chase it.
- **Prefix caching keys on a model topology fingerprint.** Two new architectures need to contribute their topology to that key or you can get cross-model KV reuse with the wrong layout. I'll check `cacheCoordinatorModelKey` myself when this is closer, but flagging that the surface exists.
- **`argPartition` has no defined order among ties**, and we have an open temp-0 nondeterminism bug I haven't yet excluded it from. Relevant to the MoE routing here.

---

I know this reads like a lot. To be clear about where I actually land: **the architecture work is good and I want it.** Fix the padding mode (or show me the reference uses border), de-dup the vision tower so there's one copy of it to be correct, delete-or-wire the dead processors, guard the traps, and add a `grid_sample` test — and I'll merge it. The reason I'm being this specific is that a VLM that produces confident, fluent, *subtly wrong* output is the single worst thing we can ship, because nobody files a bug for it.

rcfa:
The blocking padding-mode question resolves in favor of the code, and I checked it against the
reference so it's not my word against yours: HF `Glm4vVisionEmbeddings.forward` calls
`F.grid_sample(pos_embed_2d, grid, mode="bicubic", align_corners=False, padding_mode="border")` —
so `border` (clamp) is correct, and it's now documented at the call site with that citation.

Checking the reference also surfaced a *real* bug the earlier port had: it interpolated with
**bilinear**, but the reference is **bicubic** — which diverges across the whole interior of any
upsampled position grid, not just the borders. The shared module below is bicubic now.

On the rest:
- **De-dup** — the numerically-sensitive interpolation (the part you had to report at two line
  numbers) now lives once in a shared `Glm4SharedVision.gridSampleBicubic` that both `Glm4v` and
  `Glm4vMoe` call. One copy to be correct, which was the core of your argument.
- **Dead processors** — the unreachable `Glm4vProcessor` / `Glm4vMoeProcessor` structs are gone; the
  factory registers the existing `GlmOcrProcessor` (the comment you noted was right).
- **grid_sample test** — `Glm4VisionTests` covers it.

Still open, and I'd value your steer: the *full* vision-tower de-dup (only the sensitive
interpolation is shared so far — the towers are otherwise still separate), and I want to confirm the
GLM-4.5V template resolves with tools enabled, which you flagged. I'll follow up on both.

rcfa:
Following up on the three items left open when this merged.

## 1. GLM-4.5V template with tools enabled — checked, does not reproduce

Your concern was VLMs that ship a template only in a sidecar, silently falling
back to ChatML and leaking `<|im_start|>` once tools are involved. Worth
checking, because GLM-4.5V is exactly that shape:

| bundle | `chat_template` in `tokenizer_config.json` | sidecar |
|---|---|---|
| `mlx-community/GLM-4.5V-3bit` | **absent** | `chat_template.jinja` |
| `mlx-community/Huihui-GLM-4.5V-abliterated-5bit` | **absent** | `chat_template.jinja` |
| `lmstudio-community/GLM-4.6V-Flash-MLX-8bit` | present | both — control |

Rendered prompts, with and without tools declared in the native slot:

| bundle | tools | `<\|im_start\|>` | GLM markers | tool block | fallback logged |
|---|---|---|---|---|---|
| GLM-4.5V-3bit | with | **0** | 2 | 3 | no |
| GLM-4.5V-3bit | without | **0** | 2 | 0 | no |
| Huihui-GLM-4.5V-abliterated | with | **0** | 2 | 3 | no |
| Huihui-GLM-4.5V-abliterated | without | **0** | 2 | 0 | no |
| GLM-4.6V-Flash (control) | with | **0** | 2 | 3 | no |
| GLM-4.6V-Flash (control) | without | **0** | 2 | 0 | no |

The tool block appears exactly when tools are declared and never otherwise, and
it is GLM's own format rather than a generic one:

```
[gMASK]<sop><|system|>
# Tools
...
<tool_call>{function-name}
<arg_key>{arg-key-1}</arg_key>
<arg_value>{arg-value-1}</arg_value>
</tool_call><|user|>
What is in this image?<|assistant|>
```

`chatTemplateText` in `VLMModelFactory` reads `chat_template.jinja` before
consulting `tokenizer_config.json`, so the sidecar-only shape resolves. Checked
with `VMLX_CHAT_TEMPLATE_FALLBACK_LOG=1`; the fallback never fires.

Note this is a property of the bundle, not of the port — so if you want it
guarded rather than checked once, the natural place is a test asserting that a
resolved template's markers are the model's own. Happy to add one.

## 2. Throwing trap guards — a question before I build it

The remaining traps (non-merge-divisible grid dims, zero merge, video indexing)
sit in the **non-throwing** forward path. So "load error, not trap" is not a
local change: it needs a throwing path threaded through, and a `precondition`
would still trap and would not meet your bar.

Two shapes, and I would rather you pick than guess:

- **(a) validate at construction** — check the grid/merge invariants once when
  the vision tower is built, where throwing is already allowed, and leave the
  forward path unguarded because the invariant is established. Smaller diff, and
  it catches the bundle-level cases, which is what these are.
- **(b) make the forward path throwing** — faithful for input-dependent cases
  (video indexing), but it changes a signature that a lot of code calls.

(a) covers the two grid/merge traps; only the video-indexing one genuinely
depends on per-request input. Which would you accept?

## 3. Vision-tower de-duplication — worth it?

This PR shares `gridSampleBicubic`; roughly 450 lines remain duplicated across
the two towers. I can do it, but it is the largest and riskiest change on my
list and it touches code that currently works, so I would rather ask than
surprise you: is that a direction you want, or would you rather the duplication
stayed until a third consumer justifies it?

If you do want it, I would split it — one PR per extracted piece, each with a
before/after equivalence check — rather than one 450-line move.

rcfa:
All three now exist as PRs, so they are in your queue rather than in a comment.

| | PR | state |
|---|---|---|
| template resolution, guarded | #325 | draft |
| grid invariant (option **a**) | #326 | draft |
| first de-dup extraction | #327 | draft |

**#325** — the check I posted above, turned into a guard. It also turned up
something: `chatTemplateText` is written twice, byte-identical, private, in
`LLMModelFactory` and `VLMModelFactory`. And precedence is load-bearing — of 58
local bundles carrying both a sidecar and an inline template, **13 carry two
different ones** (Nemotron-3-Nano-Omni: 14277 B vs 10504 B; Ornith-1.5: 7593 B
vs 7756 B), so which file wins decides their prompt format. The guard avoids a
per-family marker table, which would rot: whatever a bundle ships as its sidecar
is exactly what resolution returns. Runs over 121 local bundles.

**#326** — option (a): `prepare` already throws on both towers, so the grid
invariant is established there and the non-throwing forward path keeps its
signatures. Scoped to the two grid/merge cases; video indexing is per-request
rather than configuration, and that is the one that would genuinely argue for a
throwing forward path — I would rather settle it separately than bundle it.

**#327** — `getRopeIndex`, 89 lines in each tower, byte-identical except for the
model name in one `precondition`. Same image and prompt at temperature 0 give
byte-identical output before and after. The remaining runs are listed in the PR;
the real total is **553 lines across 11 runs**, more than the ~450 I said above.

All three are drafts because each embeds a judgement you may want made
differently — the guard's placement, the de-dup's existence at all. Happy to
close any of them; #327 in particular is a shape to agree on, not a fait
accompli.

PR #93, opened 2026-06-27
Remove stale <end_of_turn> extraEOSTokens from the gemma4_* configs (fixes a live false-stop)

rcfa:
## Summary

Per review: **taking you up on the actual removal — and reframing this as a false-stop *fix*, not a cleanup.** My original framing ("inert, can never fire") was wrong, and your trace is right.

Gemma-4 renamed its turn-end token — `<end_of_turn>` is **not in the gemma-4 vocabulary**. But the entry does **not** evaporate:

- `convertTokenToId("<end_of_turn>")` returns the **`<unk>` id** — it's not a nil-on-missing API;
- the `id == unknownID` branch falls through to `textFallbackStopStrings(...)`;
- and the `extraEOSTokens` loop passes **`allowExactTextFallback: true`**, so the fallback returns `[token]` for anything starting with `<`.

Net effect on gemma-4: a **live text stop-string `"<end_of_turn>"`**. It can't fire as a token, but it *will* halt generation the moment the model emits those literal characters — not hypothetical for a model people use for coding and agent work. Ask gemma-4 to explain a Gemma-3 chat template, or to diff two templates, and the answer gets truncated mid-sentence with `stop`.

(For contrast, the `commonEndTokenStrings` blanket passes `allowExactTextFallback: false`, so *that* copy of the same string genuinely **is** inert. Only the `extraEOSTokens` one is live — the two loops are eight lines apart and differ by one argument. Easy to miss; I did.)

## What changed

- **Removed** `extraEOSTokens: ["<end_of_turn>"]` from the three `gemma4_*` configs.
- **Left `gemma3_*` / `gemma3n_*` untouched** — `<end_of_turn>` is still their real turn-end token.
- Added a tombstone comment at the gemma4 block so it doesn't get re-added, naming the `allowExactTextFallback: true` mechanism.

Nothing is lost: gemma-4 ends turns with `<turn|>` (id 106), which it already declares in `generation_config.json` (`eos_token_id: [1, 106, 50]`) and which the loader honors via `resolvedEOSTokenIds`.

## Tests

`Tests/MLXLMTests/Gemma4StaleEOSTests.swift` (new file):

- `gemma-4 configs declare no <end_of_turn> extra EOS token`
- `gemma-3 configs keep <end_of_turn>` (the fence you asked for)
- `<end_of_turn> resolves to a live text stop for gemma-3 but not gemma-4` — the behavioural pin

**Verified red→green**, not just green: against an unpatched factory the two gemma-4 tests fail on exactly
their own assertions (3 issues — one per gemma4 config), and the gemma-3 fence passes in *both* states, as
a fence should. With the fix: 3/3.

**Why a new file rather than appending to `StopStringMatcherTests`** (which is where I'd originally put
them, to reuse its private `NoSpecialTokenTokenizer`): #91 adds its tests to the tail of that same suite.
Since these are two independent PRs, each patch has to apply to a pristine tree on its own — but two
patches that both append at the same anchor cannot then be applied *sequentially*, in either order. A
separate file makes them orthogonal for any order and any base, at the cost of the small tokenizer stub
duplicated at the top. Happy to fold them back together if you'd rather merge one and rebase the other.

## Unrelated observation, take it or leave it

`jamba_3b` also declares `extraEOSTokens: ["<end_of_turn>"]`. I have no idea whether that's correct for
Jamba's vocabulary, so I've left it alone — but if `<end_of_turn>` isn't a real Jamba token, it's the same
live-text-stop shape as this bug and worth a look.

### DIFF
--- a/Libraries/MLXLLM/LLMModelFactory.swift
+++ b/Libraries/MLXLLM/LLMModelFactory.swift
@@ -1070,22 +1070,28 @@ public class LLMRegistry: AbstractModelRegistry, @unchecked Sendable {
         defaultPrompt: ""
     )
 
+    // Gemma-4 renamed its turn-end token: `<end_of_turn>` is NOT in the gemma-4 vocabulary.
+    // Do NOT add `extraEOSTokens: ["<end_of_turn>"]` to the gemma4_* configs. The extraEOSTokens
+    // loop passes `allowExactTextFallback: true`, so an entry the tokenizer cannot resolve does
+    // NOT evaporate — it becomes a LIVE text stop-string that truncates any answer containing
+    // those characters (e.g. gemma-4 asked to explain a Gemma-3 chat template). Gemma-4 ends
+    // turns with `<turn|>` (id 106), which it already declares in `generation_config.json`
+    // (`eos_token_id: [1, 106, 50]`) and which the loader honors via `resolvedEOSTokenIds`.
+    // The `gemma3_*` / `gemma3n_*` configs above intentionally KEEP the entry — `<end_of_turn>`
+    // is still their real turn-end token.
     static public let gemma4_27b_it_4bit = ModelConfiguration(
         id: "mlx-community/gemma-4-27b-it-4bit",
-        defaultPrompt: "Explain quantum computing briefly.",
-        extraEOSTokens: ["<end_of_turn>"]
+        defaultPrompt: "Explain quantum computing briefly."
     )
 
     static public let gemma4_12b_it_4bit = ModelConfiguration(
         id: "mlx-community/gemma-4-12b-it-4bit",
-        defaultPrompt: "What is the meaning of life?",
-        extraEOSTokens: ["<end_of_turn>"]
+        defaultPrompt: "What is the meaning of life?"
     )
 
     static public let gemma4_27b_it_qat_4bit = ModelConfiguration(
         id: "mlx-community/gemma-4-27b-it-qat-4bit",
-        defaultPrompt: "Explain quantum computing briefly.",
-        extraEOSTokens: ["<end_of_turn>"]
+        defaultPrompt: "Explain quantum computing briefly."
     )
 
     private static func all() -> [ModelConfiguration] {
--- a/Tests/MLXLMTests/Gemma4StaleEOSTests.swift
+++ b/Tests/MLXLMTests/Gemma4StaleEOSTests.swift
@@ -1,0 +1,71 @@
+import Foundation
+import MLXLLM
+import Testing
+
+@testable import MLXLMCommon
+
+/// Pins the gemma `<end_of_turn>` `extraEOSTokens` contract: gemma-4 must NOT declare it, gemma-3 must.
+///
+/// These live in their own file rather than alongside `StopStringMatcherTests` on purpose. They and the
+/// `generation_config.json` eos-gate tests are offered as two independent upstream PRs, so each patch has
+/// to apply to a pristine tree on its own; if both appended to the tail of the same suite they could not
+/// be applied sequentially (an add/add conflict at the same anchor). Separate files keep them orthogonal
+/// in any order, at the cost of the small tokenizer stub below.
+@Suite("gemma <end_of_turn> stale extraEOSTokens")
+struct Gemma4StaleEOSTests {
+
+    /// A tokenizer that resolves nothing — standing in for gemma-4's, which has no `<end_of_turn>`.
+    private struct NoSpecialTokenTokenizer: Tokenizer {
+        var bosToken: String? { nil }
+        var eosToken: String? { nil }
+        var unknownToken: String? { nil }
+
+        func encode(text: String, addSpecialTokens: Bool) -> [Int] { [] }
+        func decode(tokenIds: [Int], skipSpecialTokens: Bool) -> String { "" }
+        func convertTokenToId(_ token: String) -> Int? { nil }
+        func convertIdToToken(_ id: Int) -> String? { nil }
+
+        func applyChatTemplate(
+            messages: [[String: any Sendable]],
+            tools: [[String: any Sendable]]?,
+            additionalContext: [String: any Sendable]?
+        ) throws -> [Int] { [] }
+    }
+
+    /// Gemma-4 renamed its turn-end token, so `<end_of_turn>` is absent from its vocabulary.
+    /// Declaring it in `extraEOSTokens` is NOT inert: that loop passes `allowExactTextFallback: true`,
+    /// so an entry the tokenizer cannot resolve becomes a live TEXT stop-string — able to truncate a
+    /// legitimate answer that happens to type those characters.
+    @Test("gemma-4 configs declare no <end_of_turn> extra EOS token")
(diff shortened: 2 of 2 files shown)


jjang-ai:
Your diagnosis is right, and I verified it against two local gemma-4 bundles rather than taking it on trust:

```
gemma-4-E2B-it-qat-MXFP4 / gemma-4-E4B-it-qat-MXFP4
  '<end_of_turn>'  -> not in vocab (no id)
  '<turn|>'        -> id 106
  id 105 = '<|turn>' , id 1 = '<eos>' , id 50 = '<|tool_response>'
  generation_config.json eos_token_id = [1, 106, 50]
```

So: `<end_of_turn>` genuinely is absent from the gemma-4 vocabulary, `<turn|>` really is 106, and the declared eos really is `[1, 106, 50]`. All three claims check out.

**But the entry is not inert, and that changes what we should do here.** Trace it through `resolveStopTokenString` (`StopStringMatcher.swift:219`):

- `convertTokenToId("<end_of_turn>")` returns the **`<unk>` id** (it's not a nil-on-missing API — a pitfall that has bitten us before).
- The `id == unknownID` branch catches that and falls through to `textFallbackStopStrings(...)`.
- The **`extraEOSTokens` loop passes `allowExactTextFallback: true`** (`StopStringMatcher.swift:197-203`), and the fallback returns `[token]` for anything starting with `<`.

Net effect: on gemma-4 this doesn't evaporate — it becomes a **live text stop-string `"<end_of_turn>"`**. It can't fire as a token, but it *will* halt generation if the model ever emits that literal string in prose. That's not hypothetical for a model people use for coding and agent work: ask gemma-4 to explain Gemma-3's chat template, or to diff two templates, and it has every reason to type those exact characters — and the answer would get cut off mid-sentence with `stop`.

(For contrast, the `commonEndTokenStrings` blanket passes `allowExactTextFallback: false`, so *that* copy of the same string genuinely is inert. Only the `extraEOSTokens` one is live. Easy to miss — the two loops are eight lines apart and differ by one argument.)

So I'd take you up on your offer: **please make it the actual removal**, dropping `extraEOSTokens: ["<end_of_turn>"]` from the three `gemma4_*` configs only, and leaving `gemma3_*` / `gemma3n_*` exactly as they are — your instinct to fence those off is correct, it's still their real turn-end token.

Two asks so it doesn't regress later:
1. A small test pinning that a gemma-4 config resolves **no** text stop-string for `<end_of_turn>` (and that gemma-3 still does). That's the regression that would otherwise creep back in silently.
2. Since you've already got the tokenizer facts: worth stating in the PR body that the fix is a *false-stop* fix, not a cleanup. It's a behaviour change, and framing it as cosmetic undersells it.

Good catch, and thanks for fencing the gemma-3 side explicitly — that's the part a careless version of this change would have broken.

jjang-ai:
Current 2026-07-18 source and live-proof audit

Scope
- Exact head: `21f20c5c533bb97213e884386dc72848ed9ddf9a`
- Exact base: `a9118c402227accf425113e361cd4520a0cb25ad`
- Diff remains exactly two files: `Libraries/MLXLLM/LLMModelFactory.swift` and `Tests/MLXLMTests/Gemma4StaleEOSTests.swift`.
- The source change touches only the three static `gemma4_*` registry configurations plus the focused test. Gemma 3 and Gemma 3n are explicitly fenced. No Bonsai, routing, hardware-guidance, cache-policy, parser, sampler, or Osaurus code is in this PR. The Bonsai work is separate and already merged in Osaurus PR #2041.

Source trace
- The three static Gemma 4 registry entries declared `extraEOSTokens: ["<end_of_turn>"]`.
- `resolveStopSequences` handles unresolved explicit extra EOS strings with exact-text fallback, so the stale registry value becomes a live text stop and can truncate ordinary generated text containing that literal.
- The patch removes that value only from the Gemma 4 registry entries.

Current red/green evidence
- Green at the exact PR head: `swift test --filter Gemma4StaleEOSTests` passed 3/3.
- Red control: the same test file placed on unpatched base `a9118c402227accf425113e361cd4520a0cb25ad` produced four expectation issues: all three Gemma 4 registry entries retained the stale value and resolved stop strings retained `<end_of_turn>`.
- Adjacent non-Metal selection passed 28/28: 22 Swift Testing cases covering `StopStringMatcherTests` plus `Gemma4StaleEOSTests`, and six pure `GenerationConfigDefaultsTests` XCTest cases.
- The broader class-level selection is NOT green: after the same 22 Swift Testing cases passed, existing `testModelContainerCacheTopologySnapshotUsesLiveCacheTypes` exited because the clean worktree could not load the default Metal library. This is recorded as environment-blocked, not passed.
- `git diff --check` is clean.
- GitHub workflow jobs are SKIPPED, not passed.

Current live Osaurus evidence and limitation
- Fresh isolated Release app source: Osaurus `2a280aaad75caad1dccd1937460f75b6d5da86d4`, packaged vMLX pin `a26c7ecec950f18e3d07c8402fbd8c80f40ac764`, bundle ID `com.dinoki.osaurus.cacheparityproof`.
- Live local models selected through the actual UI: `/Users/eric/models/OsaurusAI/OsaurusAI--gemma-4-12B-it-MXFP8` and `/Users/eric/models/OsaurusAI/OsaurusAI--gemma-4-12B-it-qat-JANG_4M`. No MXFP4 model was loaded or downloaded.
- JANG_4M visibly emitted all requested lines around a literal `<end_of_turn>` without cutoff.
- Controlled invalid-tool recovery with Thinking OFF produced exactly `recovered-after-tool-error-off` at 41.5 tok/s with one failed tool card, no Thought, and no second tool. Thinking ON produced exactly `recovered-after-tool-error` at 42.7 tok/s with normal pre-tool reasoning, one failed tool card, and no post-error loop or raw channel marker.
- MXFP8 controlled invalid-tool recovery likewise completed exactly with Thinking OFF at 31.9 tok/s and Thinking ON at 32.0 tok/s; neither produced a post-error loop, a second tool, or raw channel-marker leakage.
- Important route caveat: local Osaurus loads use `ModelConfiguration(directory:)`, where `extraEOSTokens` defaults empty; Osaurus applies only explicit tool/reasoning parser overrides. Therefore this PRs static-registry change is not active in those exact local MXFP8/JANG_4M UI runs. Those runs are current non-regression/non-reproduction evidence, not a live before/after proof of this patch.

Verdict
- Source fix: verified by current trace plus red/green tests.
- Scoped non-Metal adjacency: passed.
- Full Metal-dependent suite: blocked in this clean worktree.
- Local Osaurus cutoff/tool capture: not reproduced on MXFP8 or JANG_4M and not attributable to this static-registry path.
- Overall status remains PARTIAL, not release-ready proof.

jjang-ai:
Merged status record

PR #93 was squash-merged at `2026-07-19T02:47:46Z`. Both the GitHub PR record and remote `refs/heads/main` resolve to `604d24e4874211e3f5aa6925f8b821ba136df516`.

The prior audit caveats remain unchanged: this closes only the static Gemma 4 registry false-stop defect. It does not establish a live before/after for the local-directory Osaurus route, does not explain the supplied MXFP4 capture, and does not make the broader Gemma/Osaurus release matrix ready.

PR #94, opened 2026-06-27
Vendored swift-transformers: add Sendable to Hub.Repo / Hub.RepoType (Swift 6 build fix)

rcfa:
## Summary

Under Swift 6 strict concurrency, `Hub.RepoType` and `Hub.Repo` (in the vendored `swift-transformers`
copy under `Vendors/`) need `Sendable` to cross actor/task boundaries; without it, builds that touch
these types from concurrent contexts fail.

Both are trivially sendable (a `String`-raw enum and an all-`let` struct of `Sendable` fields). This
adds the conformance to the **vendored copy** so vmlx-swift builds clean under Swift 6 now, without
waiting on an upstream `swift-transformers` release + a vendor bump.

The same fix is now **merged upstream** at huggingface/swift-transformers#374 (merged 2026-06-28) — so
this vendored stopgap matches the accepted upstream change, and can be dropped on the next vendor sync
that picks it up. No behavior change.

### DIFF
--- a/Vendors/swift-transformers/Sources/Hub/Hub.swift
+++ b/Vendors/swift-transformers/Sources/Hub/Hub.swift
@@ -77,7 +77,7 @@ public extension Hub {
     }
 
     /// The type of repository on the Hugging Face Hub.
-    enum RepoType: String, Codable {
+    enum RepoType: String, Codable, Sendable {
         /// Model repositories containing machine learning models.
         case models
         /// Dataset repositories containing training and evaluation data.
@@ -90,7 +90,7 @@ public extension Hub {
     ///
     /// A repository is identified by its unique ID and type, allowing access to
     /// different kinds of resources hosted on the Hub platform.
-    struct Repo: Codable {
+    struct Repo: Codable, Sendable {
         /// The unique identifier for the repository (e.g., "microsoft/DialoGPT-medium").
         public let id: String
         /// The type of repository (models, datasets, or spaces).


rcfa:
Upstream PR approved

jjang-ai:
Thanks — the conformance itself is correct (`RepoType` is a `String`-raw enum, `Repo` is all-`let` over `Sendable` fields), and it matches what upstream accepted. Two bits of context that aren't visible from outside the repo, though, and they change what we'd want to do with it:

**1. We don't vendor from `huggingface/swift-transformers`.** Per `Vendors/README.md`, `Hub`/`Tokenizers`/`Generation`/`Models` are copied from **`osaurus-ai/swift-transformers`** at revision `087a66b1`. So patching the vendored copy in-tree means the next vendor sync overwrites it — the fix would silently disappear on a routine refresh, which is a worse failure than not having it (it'd come back as a mystery build break months later). The durable path is to land it in `osaurus-ai/swift-transformers` first and then bump the vendored revision here; at that point this diff becomes unnecessary rather than redundant.

**2. The stated symptom doesn't reproduce on `main`.** `Hub.Repo` / `Hub.RepoType` aren't referenced anywhere under `Libraries/` or `Sources/`, and `main` builds clean today (MLXLMCommon is deliberately compiled in Swift 5 mode — see the long comment on that target in `Package.swift` — while the rest is Swift 6). So this isn't fixing a live break *here*; it's pre-emptive, or fixing something in a downstream consumer. That's fine, but it's worth saying plainly in the PR body, because "Swift 6 build fix" reads like the build is currently red and it isn't.

If you're hitting this from a downstream package, could you paste the actual compiler error and the context it fires in? If it's real for consumers we should fix it in the fork and re-vendor, which helps everyone rather than just this tree.

No objection to the change on its merits — it's the delivery mechanism that needs to move.

rcfa:
Agreed — closing this. Patching the vendored copy is the wrong delivery (a vendor sync would erase it), and on our side it turns out not to be a live break at all: `MLXLMCommon` is pinned to `.v5`, nothing in our tree references `Hub.Repo`/`Hub.RepoType`, and a strict-`.v6` build compiles clean with the conformance removed (0 errors). The real fix is already merged upstream at huggingface/swift-transformers#374; I'll get it onto osaurus-ai/swift-transformers (currently ~2 months behind and missing it) so it flows in on the next vendored bump. Thanks for the careful trace.

PR #95, opened 2026-06-27
Llama: strip stray language_model.lm_head. prefix in sanitize() so prefixed checkpoints load

rcfa:
## Summary

Per review: **taking both asks, and taking you up on the shared helper.** This PR is now two commits —
the helper, then the Llama call site.

Some quantized text-only Llama checkpoints emit the output projection under a VLM-style
`language_model.lm_head.*` prefix while the body stays at `model.*`. `LlamaModel` binds the head at the
top-level key `lm_head`, so these fail to load with `Unhandled keys [language_model]`.

## Your converter diagnosis is confirmed from this side

You wrote that the JANG conversion path emits VLM-style `language_model.*` on text-only bodies and that
it's a known open bug on your side. I can confirm that independently: on
`cais/HarmBench-Llama-2-13b-cls-Jang_6M` the **bf16 source is flat `lm_head.*`**, and the JANG output
has **zero** flat `lm_head.*` keys and exactly three prefixed ones:

```
language_model.lm_head.weight
language_model.lm_head.scales
language_model.lm_head.biases
```

927 keys in the bundle; those 3 are the only `language_model.*` in it. The body is untouched at
`model.*`. So the prefix is introduced by the conversion, on exactly one module, and the quant itself is
fine (the shape walk produced 282 per-layer overrides and the classifier discriminates correctly once
it loads). That's what makes "absorb the artifact" the right call rather than "reject the bundle".

The same three keys, in the same shape, appear on a **Mixtral** JANG bundle in #96 — different
architecture, identical artifact. Which is the argument for your helper better than I could make it.

## 1. The collision guard — measured, including one case the suggested version doesn't cover

Taken, and it's the right catch: a nondeterministic weight bind is worse than the load error it
replaces.

One note on the shape of it. The suggested guard —

```swift
guard weights[stripped] == nil else { continue }
```

— covers a prefixed key colliding with an **existing unprefixed** key. It does **not** cover **two
prefixed spellings colliding with each other**: `language_model.model.x` and `model.language_model.x`
both want `model.x`, and neither exists unprefixed, so there is no existing key to guard against and
last-write-wins still picks by iteration order.

I didn't want to just assert that at you, so I measured it. Implementing your guard verbatim and running
the suite in **25 separate processes** (fresh hash seed each):

| implementation | result |
|---|---|
| `guard weights[stripped] == nil else { continue }` | **14 passed / 11 failed of 25** |
| grouping by destination + fixed rule | 25 / 25 |

So on that one input the suggested form binds the wrong tensor in ~44% of runs — the exact
loads-fine-wrong-sometimes shape you're guarding against.

The helper therefore groups by destination first and resolves by a fixed rule: an unprefixed key already
at the destination wins; any remaining tie goes to the lexicographically smallest source. Arbitrary, but
*fixed* — the point is only that it never varies between runs. Losers are logged at `warning` with the
full source list, since a bundle that hits this is malformed and wants re-converting.

**Two honest caveats.**

*The hazard is latent, not observed.* Neither witness bundle actually collides — both have zero flat
`lm_head.*`, so the guard never fires on them in practice.

*My first version of this test was junk, and the 44% is why.* It tried to catch order-dependence by
building the same dictionary in two insertion orders — which cannot work: for a fixed key set and a
fixed seed, `Dictionary` iterates the same way regardless of insertion order. It passed against the
broken implementation, which is what gave it away. The test now varies the **key sets** (200 of them),
each hashing independently, so a last-write-wins implementation has to win 200 coin flips:

| new test vs | result |
|---|---|
| suggested guard | **0 passed / 10 failed** |
| this PR's guard | 10 / 10 |

## 2. The shared helper

`Weights.stripLanguageModelPrefix(_:only:)` in `MLXLMCommon`, called by both this PR and #96:

```swift
Weights.stripLanguageModelPrefix(weights, only: ["lm_head."])
    .filter { !$0.key.contains("self_attn.rotary_emb.inv_freq") }
```

`only:` scopes the rewrite to keys whose **stripped** form starts with one of the given prefixes. That
keeps this call site's semantics exactly as they were — only the stray head is absorbed, and a genuine
nested `language_model.*` body key is left alone — while #96 passes `nil` to strip wherever the prefix
appears. That parameter is what lets one function serve call sites whose semantics differ, which was the
thing making eleven copies diverge.

I did **not** migrate the other eleven strip sites. You said you weren't asking for it, and each has its
own semantics that deserve their own reading; this just stops the count going up. If you want them
folded in, happy to do it as a separate PR.

## 3. Tests

`Tests/MLXLMTests/WeightsTests.swift` — the motivating strip, quant sidecars travelling with the head,
**both** collision shapes, resolution stability across 200 independently-hashing key sets, `only:`
scoping in both directions, and identity on an unprefixed checkpoint.

Tensors are tagged by **shape** rather than value: this function routes keys and never inspects tensor
contents, so a distinct extent per tensor is enough to catch a mis-bind and no assertion has to force an
evaluation.

## Note on the branch

The helper commit is deliberately in **both** this PR and #96, identical in both, so neither has to wait
on the other. Whichever you merge first, the other rebases cleanly and its copy collapses. Happy to
restructure if you'd rather have it standalone.

### DIFF
--- a/Libraries/MLXLLM/Models/Llama.swift
+++ b/Libraries/MLXLLM/Models/Llama.swift
@@ -185,10 +185,14 @@ public class LlamaModel: Module, LLMModel, KVCacheDimensionProvider {
     }
 
     public func sanitize(weights: [String: MLXArray]) -> [String: MLXArray] {
-        // Remove unused precomputed rotary frequencies
-        weights.filter {
-            !$0.key.contains("self_attn.rotary_emb.inv_freq")
-        }
+        // Some quantized checkpoints (e.g. vMLX JANG_6M conversions of text-only Llama bodies) emit
+        // the output projection under a VLM-style `language_model.lm_head.*` prefix while the body
+        // stays at `model.*`. This model binds the head at the top-level `lm_head`, so absorb that
+        // stray prefix rather than failing with "Unhandled keys [language_model]". Scoped to the head:
+        // a genuine nested `language_model.*` body is not ours to rewrite.
+        Weights.stripLanguageModelPrefix(weights, only: ["lm_head."])
+            // Remove unused precomputed rotary frequencies
+            .filter { !$0.key.contains("self_attn.rotary_emb.inv_freq") }
     }
 
     public func messageGenerator(tokenizer: any Tokenizer) -> any MessageGenerator {
--- a/Libraries/MLXLMCommon/Weights.swift
+++ b/Libraries/MLXLMCommon/Weights.swift
@@ -1,0 +1,89 @@
+// Copyright © 2025 Apple Inc.
+
+import Foundation
+import MLX
+import os
+
+private let weightsLogger = Logger(subsystem: "vmlx", category: "Weights")
+
+/// Helpers for normalizing a checkpoint's weight keys before they are bound to modules.
+public enum Weights {
+
+    /// Strips the VLM-style `language_model.` prefix that some conversions emit onto text-only bodies.
+    ///
+    /// The prefix is a **converter artifact, not a checkpoint convention**: a conversion that wraps a
+    /// text-only model in the VLM key layout leaves the body — or, in the wild, just the output head —
+    /// under `language_model.*`, while the text model binds its modules at the top level. Loading such
+    /// a bundle otherwise fails with `Unhandled keys [language_model]`. Being tolerant here is
+    /// deliberate: refusing to load a bundle someone already downloaded is worse than absorbing the
+    /// artifact.
+    ///
+    /// Both spellings are handled: `model.language_model.foo` → `model.foo`, and
+    /// `language_model.foo` → `foo`.
+    ///
+    /// - Parameters:
+    ///   - weights: the checkpoint keys, as loaded.
+    ///   - only: when non-`nil`, only keys whose **stripped** form begins with one of these strings are
+    ///     rewritten; every other key passes through untouched. Use it when a checkpoint is known to
+    ///     misplace exactly one module (e.g. `only: ["lm_head."]`) so the strip cannot silently absorb
+    ///     an unrelated, genuine `language_model.*` key. `nil` strips wherever the prefix appears.
+    /// - Returns: the weights with the prefix removed.
+    ///
+    /// ## Collisions are resolved deterministically
+    ///
+    /// A mixed-provenance re-bake can carry **both** spellings of a key — a converter writes the
+    /// prefixed head while something upstream leaves the unprefixed one behind. Both then want the same
+    /// destination. Resolving that by last-write-wins would make the bind depend on `Dictionary`'s
+    /// iteration order, which is **seeded per process**: the model would silently load a different
+    /// tensor from run to run. That is the worst failure shape available — it loads fine, generates
+    /// fine-ish, and is wrong some fraction of the time.
+    ///
+    /// So: an unprefixed key already sitting at the destination always wins, and any remaining tie is
+    /// broken by lexicographically smallest source key. The choice is arbitrary but *fixed*; the point
+    /// is that it never varies between runs. Losing keys are logged.
+    public static func stripLanguageModelPrefix(
+        _ weights: [String: MLXArray], only: [String]? = nil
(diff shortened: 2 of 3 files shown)


jjang-ai:
Small PR, real problem, and the diagnosis is right — but there's a latent bug in the implementation and a bigger question about where this belongs.

**The bug: the remap can collide, nondeterministically.**

```swift
for (key, value) in weights {
    if key.hasPrefix("language_model.lm_head.") {
        remapped[String(key.dropFirst("language_model.".count))] = value   // <- unconditional overwrite
    } else {
        remapped[key] = value
    }
}
```

If a checkpoint carries **both** `lm_head.weight` and `language_model.lm_head.weight` (mixed-provenance re-bakes do this — a converter writes the prefixed head and something upstream leaves the unprefixed one behind), both branches write to the same destination key `lm_head.weight`, and **which one wins depends on Swift's dictionary iteration order**, which is seeded per-process. So the model silently loads a different output projection run to run. That's precisely the failure shape that's hardest to debug: it loads fine, generates fine-ish, and produces garbage some fraction of the time. Guard it:

```swift
if key.hasPrefix("language_model.lm_head.") {
    let stripped = String(key.dropFirst("language_model.".count))
    guard weights[stripped] == nil else { continue }   // or throw — a real conflict is a bad artifact
    remapped[stripped] = value
} else {
    remapped[key] = value
}
```

Silently preferring one is defensible; silently picking at random isn't.

**The structural note.** You're the eleventh place in this repo to strip `language_model.`:

```
DeepseekV3JANGTQ.swift:578   Gemma3nText.swift:1048    Gemma4Text.swift:874
Mistral4.swift:462           Step3p5.swift:540         Qwen35*.swift (×3)
MLXVLM/GlmOcr.swift          MLXVLM/Gemma4.swift       MLXVLM/Qwen35.swift
```

…each with slightly different semantics (some strip the whole prefix, some rewrite `model.language_model.` → `model.`, some only skip). That's not your fault and I'm not going to make you fix it, but it's worth naming what it means: **the prefix is a converter artifact, not a checkpoint convention.** Our own JANG conversion path emits VLM-style `language_model.*` on text-only bodies — that's a known open bug on our side (it's the same root cause behind the Kanana `lm_head` repack fix), and every one of these eleven strips is a runtime patch for it.

So I'll take this — the runtime should be tolerant of artifacts already in the wild, and refusing to load a bundle someone already downloaded is worse than absorbing the prefix. But two asks:

1. **The collision guard above.** Non-negotiable for me; a nondeterministic weight bind is worse than the load error it replaces.
2. **A test.** `Tests/MLXLMTests/` — synthesize a weights dict with `language_model.lm_head.weight` + `model.*` body, assert `sanitize` produces `lm_head.weight`; and a second case with both keys present asserting a deterministic outcome. This is 20 lines and it pins the behaviour so the twelfth strip doesn't re-litigate it.

One more thing you should know, because you can't see it from outside: **#96 adds this same strip again** (`Mixtral.swift` sanitize). If both land as written, that's twelve. If you're willing, a single `Weights.stripLanguageModelPrefix(_:only:)` helper in `MLXLMCommon` that both call — with the collision guard inside it — turns this from a growing pile into one tested function, and I'd merge that version happily.

rcfa:
Addressed, and it took the shape you suggested. The strip is now a single shared
`Weights.stripLanguageModelPrefix(_:only:)` in `MLXLMCommon` that both this PR (Llama) and the
Mixtral PR call — the `0009` commit is identical on both branches, so whichever merges first, the
other rebases cleanly and its copy collapses to a no-op.

On the nondeterministic-bind you flagged: the helper resolves collisions deterministically rather
than by first-writer-wins. It groups sources by destination first (so a contested key is *visible*,
not iteration-order-dependent), then prefers the unprefixed spelling if present, else the
lexicographic-min, and logs a "malformed bundle" warning. Tests cover both directions —
`collisionPrefersUnprefixedKey` (both spellings present) and `collisionBetweenTwoPrefixedSpellings` —
plus the plain round-trips. So the bind can't pick a different tensor run-to-run.

jjang-ai:

PR #96, opened 2026-06-28
Add Mixtral (sparse 8-expert MoE) language model

rcfa:
## Summary

Adds the **`mixtral`** model type to `MLXLLM` — the original sparse Mixture-of-Experts (Mistral
decoder + 8-expert top-2 MoE). Registered in `LLMModelFactory`; unblocks Mixtral-family checkpoints
(Mixtral-8x7B / 8x22B, Prometheus-2-8x7B, etc.).

Thank you for the review — the routing-equivalence check in particular is the kind of thing I'd rather
have caught than complimented, so it's good to know it holds.

## 2 first, because your premise is wrong and it's my fault

> Test it on a JANG/MXFP re-bake, not just the 6-bit mlx-community pack.

**It already is a JANG re-bake.** The PR said "6-bit" and you reasonably read that as mlx-community
6-bit; the bundle is `prometheus-eval/prometheus-8x7b-v2.0-**Jang_6M**` — a vMLX JANG conversion. Bad
wording on my part, now fixed above. What's in it:

| | |
|---|---|
| `jang_config.json` | present, `tensor_quantization_manifest_schema: 1`, 226 manifest entries |
| per-module quant overrides | **226**, spanning **two distinct configs** — `(6 bits, gs 64)` × 97 and `(8 bits, gs 64)` × 129 |
| expert layout | **288 `switch_mlp.*` keys, 0 `experts.N.*`** — pre-stacked |
| `language_model.*` keys | 3 — `lm_head.{weight,scales,biases}` |
| total | 775 keys / 49 shards / 36 GB |

So it does exercise the thing you're actually worried about: **mixed bit-widths across modules with
per-module overrides**, which is exactly the shape where "`bits × group_size` makes two configs
shape-identical" bites. Not a uniform-bits easy path.

**But let me mark the limit of that claim honestly.** Because this bundle is pre-stacked, it takes the
early-return in `sanitize` and exercises the *stacked* path — advisor, per-module quant config, the
shape walk. It therefore does **not** exercise the fuse loop (`experts.N.{w1,w2,w3}` → `switch_mlp`);
that path is covered by the unit test below, not by a real bundle. If you want a re-bake that lands on
the fuse loop, I don't have one and would take you up on the offer.

**One thing you'll want to know:** that manifest is **schema 1**. Older `main` rejected schema-1 JANG
bundles outright (`unsupported tensor quantization manifest schema 1`), but current `main` now
**consumes** schema-1 (`guard schema == 1 || schema == 2`, gated on `quantization_scheme == "asymmetric"`
+ `quantization_backend == "mx.quantize"`), so the bundle loads as-is — no extra patch needed to
reproduce the above.

## 1. `argPartition` tie-instability — not claimed, and not yet measured

Point taken and I won't oversell: I have **not** run the temp-0 cold-vs-warm check. Our GPU is mid-way
through a long benchmark suite, and loading a 36 GB Mixtral alongside it would contend for the GPU and
corrupt the wall-clock timings that suite is recording. I'd rather give you a clean data point than a
fast one, so this is a follow-up — I'll report it as a comment rather than hold the PR, unless you'd
rather it block.

Noting for the record that I've made no determinism claim anywhere in this PR, and "same output twice"
appears nowhere in the validation.

## 4. Crash surface — one of the three is real here

**Fixed: the expert force-unwrap.** A truncated bundle carrying expert 0 but missing a later shard
would trap on `removeValue(forKey:)!`. It now checks the whole expert set before stacking and leaves
the keys in place if any is missing, so the module bind reports them — a load error, not a crash:

```swift
guard expertKeys.allSatisfy({ sanitizedWeights[$0] != nil }) else { continue }
let toJoin = expertKeys.compactMap { sanitizedWeights.removeValue(forKey: $0) }
```

**The other two aren't in this PR.** `images[0]` on an empty array and odd / non-merge-divisible grid
dims are vision-path concerns — Mixtral is text-only and has no image or grid handling anywhere in the
diff (`grep` for `images`/`grid`/`merge` in `Mixtral.swift` returns nothing). I think those belong to
#92 (GLM-4V), where they're real and where I'm addressing them. Flagging rather than silently skipping
in case they were meant for this PR and I'm missing something.

## 4b. Tests

Was a fair hit — there were none.

- `Tests/MLXLMTests/WeightsTests.swift` — the sanitize round-trip for the `language_model.` strip, with
  a collision guard (see below).
- Routing/shape unit test: **still owed.** Rather than claim it, I'd rather land the guard + helper now
  and add the fixed-logits → known-expert-selection test in the same follow-up as the determinism check,
  since both want the same fixture.

## The shared helper (from #95's review)

Taken. `Weights.stripLanguageModelPrefix(_:only:)` in `MLXLMCommon` — the helper commit is in **both**
this PR and #95, identical, so neither waits on the other. `sanitize` here calls it with `only: nil`
(strip wherever the prefix appears), preserving this PR's semantics exactly.

Worth noting *why* this PR needed the strip at all: the Prometheus JANG bundle above carries
`language_model.lm_head.{weight,scales,biases}` and **no** flat `lm_head.*` — the identical three-key
artifact that motivated #95 on a Llama-2-13b classifier. Different architecture, same converter
signature. That's the case for one tested function rather than thirteen strips, made by the bundles
themselves.

## 5. Duplication

Leaving the MoE gate alone, as you suggested — I took the `language_model.` helper because you offered
it and the artifact is demonstrably cross-architecture, but folding `GLM4MOE`/`DeepseekV3`'s gates
together is a refactor of code I didn't write and would rather not smuggle into a port.

## Implementation (unchanged)

- **Decoder / MoE block** mirror `PhiMoE` — `SwitchGLU` for the expert FFN, top-2 softmax routing,
  `JangPressCanonicalExpertAdvisor.shared.observe` rather than an open-coded expert loop.
- **Attention** is standard Llama: `initializeRope`, RMSNorm, GQA, bias-free q/k/v/o.
- `rope_traditional` is intentionally not a config field (absent from HF Mixtral configs; defaults to
  false).

## Validation (on-device)

`prometheus-eval/prometheus-8x7b-v2.0-Jang_6M` (Mixtral-8x7B, JANG_6M, detail above): loads and
generates coherently, emitting the Prometheus rubric `[RESULT] <1-5>` grading format with sensible,
discriminating scores (cross-checked against an independent evaluator on the same inputs).

### DIFF
--- a/Libraries/MLXLLM/LLMModelFactory.swift
+++ b/Libraries/MLXLLM/LLMModelFactory.swift
@@ -27,6 +27,7 @@ public enum LLMTypeRegistry {
             "phi": create(PhiConfiguration.self, PhiModel.init),
             "phi3": create(Phi3Configuration.self, Phi3Model.init),
             "phimoe": create(PhiMoEConfiguration.self, PhiMoEModel.init),
+            "mixtral": create(MixtralConfiguration.self, MixtralModel.init),
             "gemma": create(GemmaConfiguration.self, GemmaModel.init),
             "gemma2": create(Gemma2Configuration.self, Gemma2Model.init),
             "gemma3": create(Gemma3TextConfiguration.self, Gemma3TextModel.init),
--- a/Libraries/MLXLLM/Models/Mixtral.swift
+++ b/Libraries/MLXLLM/Models/Mixtral.swift
@@ -1,0 +1,238 @@
+import Foundation
+import MLX
+import MLXLMCommon
+import MLXNN
+
+// Port of https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/models/mixtral.py
+//
+// Mixtral = vanilla Llama/Mistral decoder (RMSNorm, plain RoPE, GQA, no biases) with a sparse
+// 8-expert top-2 MoE FFN (`block_sparse_moe`) in place of the dense MLP. Structurally identical to
+// PhiMoE except the attention/norm flavor — so this mirrors PhiMoE's decoder/model/sanitize and
+// swaps in Llama-style attention. The expert-weight sanitize (block_sparse_moe.experts.N.{w1,w2,w3}
+// → stacked switch_mlp) is exactly Mixtral's HF layout.
+
+public struct MixtralConfiguration: Codable, Sendable {
+    var modelType: String = "mixtral"
+    var vocabularySize: Int = 32000
+    var hiddenSize: Int = 4096
+    var intermediateSize: Int = 14336
+    var hiddenLayers: Int = 32
+    var attentionHeads: Int = 32
+    var kvHeads: Int = 8
+    var maxPositionEmbeddings: Int = 32768
+    var rmsNormEps: Float = 1e-5
+    var ropeTheta: Float = 1_000_000.0
+    var numLocalExperts: Int = 8
+    var numExpertsPerToken: Int = 2
+
+    enum CodingKeys: String, CodingKey {
+        case modelType = "model_type"
+        case vocabularySize = "vocab_size"
+        case hiddenSize = "hidden_size"
+        case intermediateSize = "intermediate_size"
+        case hiddenLayers = "num_hidden_layers"
+        case attentionHeads = "num_attention_heads"
+        case kvHeads = "num_key_value_heads"
+        case maxPositionEmbeddings = "max_position_embeddings"
+        case rmsNormEps = "rms_norm_eps"
+        case ropeTheta = "rope_theta"
+        case numLocalExperts = "num_local_experts"
+        case numExpertsPerToken = "num_experts_per_tok"
+    }
+}
+
+private class MixtralAttention: Module {
+    let args: MixtralConfiguration
+    let scale: Float
+    let headDim: Int
+
+    @ModuleInfo(key: "q_proj") var wq: Linear
+    @ModuleInfo(key: "k_proj") var wk: Linear
+    @ModuleInfo(key: "v_proj") var wv: Linear
+    @ModuleInfo(key: "o_proj") var wo: Linear
+
+    let rope: RoPELayer
+
+    init(_ args: MixtralConfiguration) {
+        self.args = args
+        let dim = args.hiddenSize
+        let heads = args.attentionHeads
+        let kvHeads = args.kvHeads
+        self.headDim = dim / heads
+        self.scale = pow(Float(headDim), -0.5)
+
+        self._wq.wrappedValue = Linear(dim, heads * headDim, bias: false)
+        self._wk.wrappedValue = Linear(dim, kvHeads * headDim, bias: false)
+        self._wv.wrappedValue = Linear(dim, kvHeads * headDim, bias: false)
+        self._wo.wrappedValue = Linear(heads * headDim, dim, bias: false)
+
+        self.rope = initializeRope(
+            dims: headDim, base: args.ropeTheta,
+            traditional: false,
+            scalingConfig: nil,
+            maxPositionEmbeddings: args.maxPositionEmbeddings)
+    }
+
+    func callAsFunction(
+        _ x: MLXArray, mask: MLXFast.ScaledDotProductAttentionMaskMode, cache: KVCache?
+    ) -> MLXArray {
+        let (B, L) = (x.dim(0), x.dim(1))
+
+        var queries = wq(x)
+        var keys = wk(x)
+        var values = wv(x)
+
+        queries = queries.reshaped(B, L, -1, headDim).transposed(0, 2, 1, 3)
+        keys = keys.reshaped(B, L, -1, headDim).transposed(0, 2, 1, 3)
+        values = values.reshaped(B, L, -1, headDim).transposed(0, 2, 1, 3)
+
+        queries = applyRotaryPosition(rope, to: queries, cache: cache)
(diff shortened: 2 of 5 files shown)


jjang-ai:
This is the strongest PR in the batch and I want to say that up front, because most of what follows is me asking for more, not me pushing back.

**What I verified rather than assumed:**

- **Routing math is correct.** You take `softmax` over the top-k logits directly; HF Mixtral does `softmax(all_logits) → topk → renormalize`. Those are algebraically identical — the full-softmax denominator is a shared constant across the selected k and cancels in the renormalization. Easy thing to get wrong by "faithfully" porting the wrong half; you got it right.
- **Config keys** (`num_local_experts`, `num_experts_per_tok`, `router_aux_loss_coef` ignored at inference) match the HF config.
- **`sanitize`** handles the `language_model.` prefix *and* has an early-return guard for weights that are already JANG-stacked, which is the case that usually bites a new MoE port.
- You reuse `SwitchGLU` and call `JangPressCanonicalExpertAdvisor.shared.observe` rather than open-coding a new expert loop. That's exactly right — the stacking/advisor path is load-bearing for our quant formats and a from-scratch expert loop would have quietly excluded Mixtral from all of it.

So: the model is right. Here's what I need before it merges, mostly from scars on our side you have no way to see.

**1. `argPartition` tie-instability.** Your top-k uses `argPartition`, which does not define an order among equal logits. We have an *open* correctness bug (temp-0 output occasionally flipping a token) that I have not yet excluded `argPartition` from — at temperature 0, a router tie broken differently between two otherwise-identical runs is a real nondeterminism source. This isn't a reason to block you; it's a reason for me to ask that you not treat "same output twice" as proof of determinism when you test. If you can run the same prompt at temp 0 across a cold load and a warm (prefix-cached) load and confirm byte-identical output, that's a data point I'd genuinely value.

**2. Test it on a JANG/MXFP re-bake, not just the 6-bit mlx-community pack.** The mlx-community 6-bit Mixtral is the *easy* path — uniform bits, no per-module overrides. Our own quant formats are where MoE ports die: the shared shape-walk has previously mis-inferred a dense 4-bit MLP as 8-bit because `bits × group_size` makes two configs shape-identical, and it silently overrode the correct per-module config. A Mixtral MXFP4 or JANG pack exercising the stacked-expert path is the test that actually derisks this. If you don't have one, say so and I'll bake one — I'd rather do that than merge on the easy config.

**3. Two `osaurus`-side wiring facts that affect you and aren't visible from this repo:**

   - **Compiled decode is off by default in osaurus** (`isCompiledDecodeSupported = false`), and the decision of whether a model gets compiled decode goes through a **name-based** gate. Nothing in that gate knows the string "mixtral". I've just measured the effect of compiled decode across three models and its *sign flips*: Gemma-4-E2B **+33%**, Llama-3.2-1B **−26%**, HY3 ~0%. So whichever way that gate happens to land for a `mixtral` bundle is currently arbitrary, and neither answer is safe to assume. I'm reworking the toggle's shape; just know that "it's fast on my machine" may not survive contact with the server.
   - **The prefix cache keys on a topology fingerprint**, and a new architecture that doesn't contribute its expert topology to that key can produce cross-model cache hits with the wrong KV layout. Worth a look at `cacheCoordinatorModelKey` on the osaurus side once this lands — I'll check it, but flagging so you know the surface exists.

**4. Two things I'd like tightened, both cheap:**

   - **Crash surface.** `images[0]` on an empty array, an expert force-unwrap, and odd / non-merge-divisible grid dims are all reachable from a malformed bundle or an empty vision input. A bad checkpoint should produce a load error, not a trap.
   - **No tests at all.** Given points 1–3, a shape/routing unit test (fixed logits → known expert selection → known weights) plus a sanitize round-trip is the minimum for me to be able to defend this if a re-bake regresses it six months from now.

**5. Duplication, for your judgment not mine.** The MoE gate here is a near-copy of `GLM4MOE.swift:96-223` and overlaps `DeepseekV3.swift:257-317`. I'm not going to hold the PR for a refactor of code you didn't write — but if a shared gate falls out naturally, three call sites is where it starts paying.

Genuinely good work. The routing-equivalence detail in particular is the kind of thing that separates a port that works from a port that looks like it works.

rcfa:
Tests are in for both risky pieces:
- **Routing** — `MixtralTests.topKRoutingSelectsHighestGates` pins the exact selection the router
  runs (`argPartition(-gates, kth: k-1)[..<k]` → softmax over the gathered gates): fixed gates → the
  two highest experts, scores summing to 1 with the top gate dominating, and identical call-to-call.
  `tiedGatesRouteRepeatably` addresses your `argPartition` tie-instability point directly — it pins
  *repeatability* on tied gates (not a particular tie winner, which is implementation-defined).
- **Sanitize** — `sanitizeStacksPerExpertWeights` (HF `experts.N.{w1,w2,w3}` → stacked
  `switch_mlp.{gate,down,up}_proj` with a leading expert axis), `sanitizeNoOpForStackedBundle` (a
  JANG-fused bundle passes through untouched), and `sanitizeLeavesTruncatedBundle` — the last one
  pins the guard that replaced the stacking force-unwrap: a bundle carrying expert 0 but missing a
  later shard is left in place so the module bind reports a load error instead of trapping.

On your offer to bake a real JANG-stacked Mixtral for an on-device round-trip — yes please, I don't
have one to test against, and that would close the last gap. The compiled-decode / prefix-cache note
I've logged as an osaurus-side follow-up (it's harness config, not this port).

PR #102, opened 2026-06-30
Qwen3.5: make the (1+weight) RMSNorm shift deterministic (fix intermittent ~7.5% degenerate loads)

rcfa:
## Summary

Qwen3.5 (`qwen3_5` / `qwen3_5_moe`) intermittently loads a **silently degenerate
model** — flat logits, garbage output — on roughly **1 in 13 loads of the same
bundle**, with no error and no log. The RMSNorm `(1 + weight)` shift decision was
made by a `Dictionary`-iteration-order coin flip. This replaces it with an
**order-independent, per-bundle** decision and factors the mechanism into a
shared resolver so other `(1 + weight)` architectures can adopt it.

The underlying gap is upstream of vmlx and worth stating plainly (see *Why this
is really a metadata gap* below): **no converter records whether a bundle's norms
are stored raw or already shifted**, so every runtime is forced to guess. This PR
makes the guess robust; it does not remove the need for the declaration.

## Symptom

Loading the *same* qwen3.5 bundle repeatedly, a fraction of loads (~7.5% on the
bundle that surfaced this) come up degenerate: every RMSNorm weight is near
zero, the hidden state flat-lines, logits are flat (~−15), and the model picks
one answer for everything / rambles to max tokens. It is per-process and baked
at load — reload and it is usually fine.

## Root cause — why it failed *probabilistically*

Qwen3.5 stores RMSNorm weights as the deviation from 1 and applies `(1 + weight)`
during `sanitize`. Whether to apply the shift was decided by
`baseNormWeightsNeedShift`, which **returned on the first
`.input_layernorm` / `.post_attention_layernorm` weight it found while iterating
a Swift `Dictionary`**:

```swift
for (key, value) in weights where value.ndim == 1 {
    guard probeSuffixes.contains(where: { key.hasSuffix($0) }) else { continue }
    let mean = value.asType(.float32).mean().item(Float.self)
    if mean < 0.5 { return true }    // raw → shift
    if mean > 0.5 { return false }   // already shifted → don't
}
```

Swift randomizes `Dictionary` iteration order per process. A raw bundle has
*most* norm weights near 0 (mean < 0.5) but a few legitimately above 0.5 — e.g.
**6 of 80** on the bundle here. When the random first pick lands on one of those
six (~7.5%), the function returns `false`, the `+1` shift is skipped for **all**
norms, and the model is degenerate. Probabilistic *whether*, deterministic
*where*: the same two outcomes every process, decided by a coin flip. A
single-sample decision over a noisy signal is the bug — the fix is to make the
decision over the *whole population*, order-independently.

## Why "just always shift when declared" is wrong (and would regress)

The obvious fix — declare the architecture and always apply `+1` — **breaks a
different set of bundles**. The *same* `qwen3_5` architecture ships in two storage
states:

| bundle | norm Σx² at load | stored as | needs `+1`? |
|---|---|---|---|
| `Ornith-1.0-35B-JANG_6M` (raw) | ≈ 64 (mean ≈ 0) | deviation-from-1 | **yes** |
| `Holo3-35B-A3B-mxfp4` (pre-shifted) | ≈ 2106 (mean ≈ 1) | already `1 + w` | **no** |

Force-shifting the pre-shifted MXFP4 bundle double-applies the offset
(Σx² 2106 → 4154, mean ≈ 2) → degenerate, exactly the failure we started with,
now on a *different* model. So whether to shift is a **per-bundle storage
property, not an architecture constant.** The two states are cleanly bimodal
(mean ≈ 0 vs ≈ 1), which is what makes an order-independent majority vote a
reliable discriminator.

## Fix

Resolve the convention with a deterministic precedence — **a declaration is
authoritative and short-circuits the vote; the vote runs only when nothing can be
declared:**

- **`NormConventionResolver` (new, `MLXLMCommon`)** — one shared resolver.
  Precedence: **per-bundle declaration** (safetensors `metadata` → `config.json`
  `norm_convention`) → **architecture declaration** → **order-independent majority
  vote** (`weightsAppearUnshifted`: count mean<0.5 vs mean>0.5 over all probe
  norms). A declaration wins outright because it states the storage state
  authoritatively; the vote is the per-bundle measurement used only when nothing
  is declared.
- **`LanguageModel.declaredNormConvention`** — new protocol property
  (default `nil`); the per-architecture declaration seam, **valid only for
  architectures whose bundles all share one storage state**. Non-breaking for
  existing conformers.
- **Qwen35** (both the `MLXLLM` text model and the `MLXVLM` model) **deliberately
  declares nothing** — the same class ships raw (JangQ, needs `+1`) *and*
  already-shifted (MXFP4, must not) bundles, so no truthful class-level claim
  exists; an authoritative class declaration would wrongly short-circuit the vote
  and degrade one of the two states. It adds a `norm_convention` config field for
  a per-bundle override and routes `sanitize` through the resolver, relying on
  that field or the vote. The duplicated per-copy first-element heuristic and
  convention constants are **deleted**.

The fragile first-element probe is gone; when no declaration applies, the
population-level decision lives in exactly one place.

## Validation — both storage states, on device

`qwen3_5_moe`, `mmlu` (logprob), tracing the norm Σx² through `sanitize`:

| bundle | Σx² in → out | decision | mmlu |
|---|---|---|---|
| `Ornith-1.0-35B-JANG_6M` (raw) | 64 → **2112** | **shifted** ✓ | **80%** |
| `Holo3-35B-A3B-mxfp4` (pre-shifted) | 2106 → **2106** | **left alone** ✓ | **75%** |
| `Ornith-1.0-35B-MXFP4` (pre-shifted) | 2112 → **2112** | **left alone** ✓ | **80%** |

Both MXFP4 bundles — which a declare-and-always-shift fix regresses to ~25%
random — stay healthy, and the raw JangQ bundle is correctly shifted. Separately,
idle loop-until-collapse on the raw bundle across both implementations:
**160 loads / 0 degenerate** (vs. ~12 expected at the old ~7.5%). Deterministic
across processes.

## Why this is really a metadata gap (upstream context)

This whole class of bug exists because **the model files don't declare their norm
storage state, and nothing in the format requires them to.** safetensors'
`__metadata__` is free-form `string→string` and mandates no semantic keys;
`config.json` has no field for it. The `(1 + weight)` convention *is* a documented
architecture feature (it's Gemma's; see huggingface/transformers), but *which
storage state a given converted bundle is in* is recorded nowhere — so every
runtime guesses, and the same guess has already shipped as the analogous `+1`
misapplication bug in mlx-examples (Gemma). `norm_convention` is a vmlx-coined
key, not a standard.

The resolver already honors a declaration the moment one appears, so the clean
long-term fix is for converters to emit it. We're filing an enhancement request
upstream (mlx-lm) proposing a mandatory metadata/config key so runtimes can stop
guessing; this PR is the robust interim.

## Template / follow-up

Qwen35 is the worked template for any `(1 + weight)` architecture: declare
`declaredNormConvention`, call `shouldApplyPlusOneShift(...)`. Natural follow-up:
adopt the resolver in the other `(1 + weight)` archs (e.g. Qwen3-Next) so they
share the same order-independent path — each with its own on-device check.

### DIFF
--- a/Libraries/MLXLLM/Models/Qwen35.swift
+++ b/Libraries/MLXLLM/Models/Qwen35.swift
@@ -50,6 +50,9 @@ public struct Qwen35TextConfiguration: Codable, Sendable {
     var ropeScaling: [String: StringOrNumber]?
     var fullAttentionInterval: Int = 4
     var mtpNumHiddenLayers: Int = 0
+    /// Authoritative RMSNorm convention declared by the bundle (config.json `norm_convention`).
+    /// When set it overrides the architecture default; nil when the bundle declares none.
+    var normConvention: String? = nil
 
     // MoE fields
     var numExperts: Int = 0
@@ -82,6 +85,7 @@ public struct Qwen35TextConfiguration: Codable, Sendable {
         case ropeScaling = "rope_scaling"
         case fullAttentionInterval = "full_attention_interval"
         case mtpNumHiddenLayers = "mtp_num_hidden_layers"
+        case normConvention = "norm_convention"
         case numExperts = "num_experts"
         case numExpertsPerTok = "num_experts_per_tok"
         case decoderSparseStep = "decoder_sparse_step"
@@ -130,6 +134,8 @@ public struct Qwen35TextConfiguration: Codable, Sendable {
             try container.decodeIfPresent(Int.self, forKey: .fullAttentionInterval) ?? 4
         self.mtpNumHiddenLayers =
             try container.decodeIfPresent(Int.self, forKey: .mtpNumHiddenLayers) ?? 0
+        self.normConvention =
+            try container.decodeIfPresent(String.self, forKey: .normConvention)
 
         // MoE fields
         self.numExperts = try container.decodeIfPresent(Int.self, forKey: .numExperts) ?? 0
@@ -1022,13 +1028,18 @@ public class Qwen35TextModel: Module, LLMModel, KVCacheDimensionProvider, Hidden
     {
         let loadNativeMTP = configuration.mtpNumHiddenLayers > 0
         var weights = loadNativeMTP ? weights : weights.filter { !Self.isMTPWeightKey($0.key) }
-        let hasUnsanitizedConv1d = weights.contains { key, value in
-            !Self.isMTPWeightKey(key) && key.contains("conv1d.weight") && value.dim(-1) != 1
-        }
-        let explicitNormConvention = normConvention != nil
-        let shouldShiftNormWeights = Self.usesQwenPlusOneNormConvention(normConvention)
-            || (!explicitNormConvention
-                && (hasUnsanitizedConv1d || Self.baseNormWeightsNeedShift(weights)))
+        // Resolve the (1 + weight) RMSNorm shift via the shared resolver: a per-bundle
+        // metadata/config declaration wins; otherwise (this arch uses the convention) the
+        // order-independent majority vote decides raw (→ shift) vs already-shifted (→ leave it).
+        // The same architecture ships both — JangQ stores raw, MXFP4 stores already-shifted — so
+        // this MUST be measured per bundle, not declared as always-on.
+        let shouldShiftNormWeights = NormConventionResolver.shouldApplyPlusOneShift(
+            metadataConvention: normConvention,
+            configConvention: configuration.normConvention,
+            declaredConvention: declaredNormConvention,
+            weights: weights,
+            probeSuffixes: [".input_layernorm.weight", ".post_attention_layernorm.weight"],
+            excluding: Self.isMTPWeightKey)
         let shouldShiftMTPNormWeights = loadNativeMTP
             && (shouldShiftNormWeights || Self.mtpNormWeightsNeedShift(weights))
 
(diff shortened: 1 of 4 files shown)


rcfa:
Converting back to draft — holding for a fix. Overnight evaluation caught a regression this introduces for **MXFP4-quantized** qwen3.5 bundles: their RMSNorm weights are stored **already shifted** (mean ≈ 1), so forcing the `(1 + weight)` shift double-shifts them (norm → ≈ 2) and loads a degenerate model. The same architecture ships *both* raw (JangQ, mean ≈ 0, needs +1) and pre-shifted (MXFP4) bundles, so whether to apply the shift is a **per-bundle** property, not an architecture constant. Reworking so the order-independent majority-vote detector is the primary decision (it distinguishes raw vs already-shifted correctly), with the declaration only asserting the architecture *uses* the convention. Will re-validate on **both** a raw and a pre-shifted bundle before un-drafting.

rcfa:
Filed the upstream side as an enhancement request to have converters **declare** the norm storage state instead of every runtime re-deriving it: ml-explore/mlx-lm#1445.

This PR's majority-vote resolver is the robust interim; `NormConventionResolver` already honors a per-bundle `norm_convention` declaration the moment a converter emits one, so if a canonical key lands upstream the heuristic simply stops being the deciding path — no further change needed here.

jjang-ai:
Thanks for this @rcfa — the root cause here is exactly right and the shared `NormConventionResolver` is the correct shape of fix. 🙏

One thing worth flagging for anyone landing here: **this was one instance of a broader bug class, not the whole thing.** The underlying hazard is *any* load-time decision made by taking the first element of a Swift `Dictionary`/`Set` — whose iteration order is randomized per process — where more than one element can match. That's what produces the "loads fine sometimes, garbage other times, fixed by reload" signature, and it can occur anywhere in the loader, not just norm-shift.

We swept `Load.swift`, the factories, `JangLoader`, and every model `sanitize()`/config-decode path (two independent passes) and found more instances of the same class, all now fixed in the follow-up **#108**:

- A whole family of routed-expert quant-bit probes (`dict["routed_expert"] ?? dict.values.first`) in `LLMModelFactory`, `Zaya`, `BailingHybrid`, `Qwen35JANGTQ`, `MiMoV2Flash` — a random routed bit width corrupts MoE dequant on a fraction of loads.
- `Load.swift` shard-set selection (`.first(where:)` where the comment said "pick the LARGEST").
- `JangLoader` shared-default quant on a frequency tie.

Good news from the sweep: the **JANG quant shape-walk itself is deterministic** (fixed candidate arrays + `.sorted()` + collect-then-require-unique), so the bulk of the JANG fleet was never exposed on that path. And `mtpNormWeightsNeedShift` here, despite looking like the old first-match probe, is actually deterministic (its probe suffixes pin to unique keys) — though it's a single-sample decision that we'll likely route through your `NormConventionResolver` vote for robustness.

Net: your diagnosis generalized further than the single site, and #108 closes out the rest of the class. Nice find.

PR #118, opened 2026-07-06
emit toolCallProgress deltas while a tool call envelope is being collected

RaajeevChandran:
## Proposed changes

Local models that write files through tools produce long silent gaps in consumer UIs. While a model generates a tool call, the tool call processor buffers the entire envelope internally and the stream emits nothing until the parsed call is complete. For a large file write this means tens of seconds with no events at all, so a chat UI cannot preview the code being written the way it can for ordinary streamed text. Downstream, Osaurus renders file writes as a live diff card, but on the local path that card can only appear after the whole call has finished.

This PR adds an incremental progress channel for in flight tool calls.

- Added a new `Generation.toolCallProgress(String)` event carrying raw envelope text deltas while a tool call is being collected. The payload is the format specific envelope text, not parsed arguments. The complete parsed call still arrives as a single `.toolCall` when the envelope closes, so `.toolCall` remains the only actionable tool event and existing consumers can ignore the new case.
- Exposed `ToolCallProcessor.collectingToolCallText`, the raw buffer of the call currently being collected. It reports nil in the `potentialToolCall` state, where an ambiguous prefix may still revert to plain text, so consumers never see text that could later be re-emitted as a visible chunk. It also reports nil for strip only processors, whose parsed calls are discarded because no tools were offered.
- Emitted the deltas from `routeGenerationText` by diffing the collecting buffer around each processed chunk. Because every streaming path (the token loop handler, `BatchEngine`, and SpecDec) already routes text through this shared function, all of them gain progress events with no changes to any per family parser state machine. A call that completes or reverts within a chunk simply reports no growth, and the prefix check handles the boundary where one call completes and the next begins inside the same chunk.
- Updated the exhaustive `Generation` switches in `BatchEngine`, the CLI, the integration test helpers, and existing tests to pass the new case through or ignore it.
- Added a focused test suite covering three contracts: multi chunk envelopes yield multiple deltas that concatenate to the collected text followed by exactly one parsed call, plain prose never yields progress events, and strip only processors never leak envelope text as progress.

Consumer impact: additive only. `Generation` gains one case; consumers with `@unknown default` or `default` branches are unaffected. Osaurus will map the new event to its streaming args channel so file writes preview live during generation.

## Checklist

Put an `x` in the boxes that apply.

- [x] I have updated the necessary documentation (if needed)

### DIFF
--- a/Libraries/IntegrationTestHelpers/IntegrationTestHelpers.swift
+++ b/Libraries/IntegrationTestHelpers/IntegrationTestHelpers.swift
@@ -234,6 +234,8 @@ public enum ChatSessionTests {
                 break
             case .toolCall(let toolCall):
                 toolCalls.append(toolCall)
+            case .toolCallProgress:
+                break
             case .prefillProgress:
 
                 break
@@ -676,7 +678,7 @@ public enum BatchEngineIntegrationTests {
             case .prefillProgress:
 
                 break
-            case .info, .prefillProgress, .toolCall, .reasoning:
+            case .info, .prefillProgress, .toolCall, .toolCallProgress, .reasoning:
                 break
             }
         }
@@ -910,7 +912,7 @@ public enum ToolCallTests {
                     text += chunk
                 case .toolCall(let toolCall):
                     toolCalls.append(toolCall)
-                case .reasoning, .prefillProgress, .info:
+                case .reasoning, .prefillProgress, .toolCallProgress, .info:
                     break
                 }
             }
--- a/Libraries/MLXLMCommon/BatchEngine/BatchEngine.swift
+++ b/Libraries/MLXLMCommon/BatchEngine/BatchEngine.swift
@@ -690,6 +690,8 @@ public actor BatchEngine {
                     continuation.yield(event)
                 case .toolCall:
                     continuation.yield(event)
+                case .toolCallProgress:
+                    continuation.yield(event)
                 case .info:
                     continuation.yield(event)
                 }
--- a/Libraries/MLXLMCommon/Evaluate.swift
+++ b/Libraries/MLXLMCommon/Evaluate.swift
@@ -3675,6 +3675,16 @@ public enum Generation: Sendable {
     /// A tool call from the language model.
     case toolCall(ToolCall)
 
+    /// An incremental delta of a tool-call envelope that is still being
+    /// generated. Emitted while the tool-call processor is collecting a call
+    /// — the payload is the raw envelope text (format-specific, e.g. the
+    /// growing JSON object), NOT parsed arguments. Consumers can use it to
+    /// preview long calls (e.g. a file write) as they stream; the complete,
+    /// parsed call still arrives as a single `.toolCall` when the envelope
+    /// closes. Callers that don't need previews can ignore this case —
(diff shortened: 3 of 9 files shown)


jjang-ai:
Reviewed this end to end — the approach is sound. Diffing the collecting buffer around each chunk and gating on the committed collecting states (nil in `.potentialToolCall` and for strip-only processors) is exactly right: it can't surface text that later reverts to a visible chunk, and it leaves every per-family parser state machine untouched. The additive `Generation` case keeps existing consumers unaffected.

Two small build-completeness items to land it green:

1. **The new test doesn't run yet.** `Tests/MLXLMCommonFocusedTests/ToolCallProgressRoutingTests.swift` isn't added to that target's explicit `sources:` list in `Package.swift`, so it never compiles or runs. (`git show <head>:Package.swift | grep ToolCallProgressRoutingTests` → not present.) Same explicit-sources gotcha that kept `NormConventionResolverTests` from running after #117.

2. **RunBench targets won't compile.** The new `Generation.toolCallProgress` case needs adding to the exhaustive (no-`default`) `switch event` in `RunBench/VLBench.swift`, `RunBench/Bench.swift`, and `RunBench/OmniBench.swift`, or those executable targets fail to build. (`StabilityBench` and `JangPressRegressionBench` have `default:` and are unaffected.)

I have both addressed on `feat/tool-call-progress-stream` (all exhaustive switches updated, the routing test registered so it runs 3/3) — happy to push those two commits here so this stays your PR, or fold them in however you prefer.

Consumer side (the actual user-facing fix): the frozen indicator lives in the osaurus native chat, which had no channel to render in-flight tool args. osaurus-ai/osaurus#1894 consumes this event to show a live "Preparing tool call" card during the buffered window. Reproduced the gap at the wire level (gemma-4-e2b, `write_file`, ~1.3 KB file): a single **4.57s window with zero stream events** right before the tool_call — which is what this event lets the UI fill.

jjang-ai:
This shipped — `.toolCallProgress` is on `main` already, with the same contract you proposed
(raw envelope text deltas emitted while the call is still being collected, with the complete
parsed call still arriving as `.toolCall`).

It landed via a parallel change; this was a three-way collision on the same fix. Current state on
`main`:

- `Generation.toolCallProgress(String)` — `Evaluate.swift:4567`, plus the `toolCallProgress`
  accessor at `:4630` and exhaustive-switch arms at `:4577`–`:4625`
- incremental delta emission — `GenerationStreamRouting.swift:57,60`
  (`events.append(.toolCallProgress(String(after.dropFirst(before.count))))`)
- `BatchEngine.swift:784`, `MLXPressCLI/main.swift:207`, `IntegrationTestHelpers.swift:237,681,915`
- tests: `ToolCallProgressRoutingTests`, `BatchEngineTests`, `BatchEngineRepetitionGuardTests`,
  `ZayaSmokeJANGTQ2Tests`

Merging this now conflicts in `ToolCallProcessor.swift` and add/adds
`ToolCallProgressRoutingTests.swift` against the version already there, with nothing left to gain.
Closing as already-delivered — thanks for driving it, the silent-gap-on-long-file-writes framing is
what the shipped version implements.

Issue #120, opened 2026-07-07
Native-MTP tuned speedup (1.56×) does not survive the serving path on hybrid MoE bundles — AR/MTP parity at 76–80 tok/s


mimeding:
## Summary

On the osaurus serving path (`/v1/chat/completions` → `BatchEngine.generate`), a tuning-proven native-MTP bundle decodes at **the same speed as forced-AR** (76–80 tok/s), even though the engine loads it with `native_mtp:d3` and the bundle's own `vmlx_mtp_tuning.json` documents **AR 83.9 → MTP 131.2 tok/s (1.56×)**. The tuned speedup does not survive the serving path on this hybrid MoE bundle.

## Environment

- Host: Apple M5 Max, 128 GB, macOS 25.5.0
- osaurus: main @ `75e352db` (Release build) + inert observability patches (osaurus PRs #1861/#1863/#1865 — hardware detection, bench CLI, diagnostics surface; no engine-path changes)
- vmlx-swift: the pin currently resolved by osaurus main (`Package.resolved`)
- Bundle: `OsaurusAI/Qwen3.6-35B-A3B-MXFP4-MTP` (23 GB, complete, `vmlx_mtp_tuning.json` present)

## Tuning file vs. measured serving throughput

The bundle's `vmlx_mtp_tuning.json`:

```json
{
 "native_mtp": {
  "best_depth": 3,
  "validated": true,
  "output_equivalent": true,
  "cache_mode": "off",
  "prompt_class": "deterministic_count_96_tokens",
  "baseline_tok_s": 83.891,
  "best_tok_s": 131.187,
  "speedup_vs_baseline": 1.564
 }
}
```

Measured through the HTTP serving path (greedy, `temperature: 0`, streaming, token counts from server `usage`, decode window = first→last generated delta), on a deterministic counting prompt ("Count from 1 to 150, one number per line"), 491-token outputs:

| Path | How it was selected | Decode tok/s (repeats) |
|---|---|---|
| MTP (`native_mtp:d3` engaged) | pure greedy request | 76.6, 74.6, 78.3, 79.9 |
| Forced AR | explicit `frequency_penalty` triggers the explicit-sampling fallback (which then drops the penalty, so decoding is identical greedy) | 78.1, 77.0, 76.4 |

[...]

PR #131, opened 2026-07-10
Expose native-MTP per-generation stats on GenerateCompletionInfo

mimeding:
## What

The `[NativeMTP]` per-generation summary is currently stderr-only; #120 (section 5) tells hosts to grep stderr for it. This PR carries the headline engagement/acceptance counters of that line structurally on `GenerateCompletionInfo` — following the existing `turboQuantCompressions` pattern — so a host (osaurus) can surface speculative-decoding health in its diagnostics endpoint without stderr parsing, which #120 (section 7) identified as the osaurus-side actionable. The struct is a subset of the ~40-key line (repair/commit counters, forward counts, timing fields, GDN-replay diagnostics, `phaseDiag`, and `samplingMode` remain stderr-only), so the full stderr line stays the authoritative diagnostic surface and is untouched; no existing behavior is gated or changed; every non-MTP path reports `nil`.

## How

- `NativeMTPGenerationStats` (public, `Sendable`, `Equatable`) is defined next to the iterator that produces it. It is snapshotted inside `storeCacheAfterGeneration`, from the same locals the stderr `String(format:)` consumes — `depth`, `currentDepth`, `verifyCalls`, `generatedTokenIds.count`, `autoregressiveFallbackTokenCount`, the accepted-by-depth histogram, `bonusCount`, `rejectedCount`, `avgCommitted`, `avgAcceptP`, `adaptiveDepthDownshiftCount`, `adaptiveFallbackReason`, `verifierMode`, and the `cacheMode` literal — same source values at the same call site. The rendering differs where the docs say so: stderr rounds `avgCommittedPerVerify`/`avgAcceptP` to `%.2f`/`%.3f` while the struct stores the full-precision source doubles; a `nil` `adaptiveFallbackReason` corresponds to the line printing `adaptiveFallback=none`; the struct's dense histogram is printed sparsely as non-zero `n:count` pairs (or `none` when empty). Doc comments name the exact `key=` entry each field is assigned from.
- `TokenIteratorProtocol` gains `var nativeMTPStats: NativeMTPGenerationStats? { get }` with a protocol-extension default of `nil` — exactly the `turboQuantCompressionCount` pattern; no other conformer changes.
- `GenerateCompletionInfo` gains `nativeMTPStats: NativeMTPGenerationStats?` with a default-`nil` init parameter, so all existing construction sites compile unchanged.

## Ordering finding (why the construction site moved)

`generateLoopTask` used to construct `GenerateCompletionInfo` *before* calling `storeCacheAfterGeneration`, but only *yields* it after the cache-store drain. Exactness forces construction after the call: the stderr line's `outputTokens` is `generatedTokenIds.count`, a parameter the iterator only sees inside `storeCacheAfterGeneration` — the iterator's internal token count differs from it when generation ends on a stop token (counted by `next()`, never appended to `generatedTokenIds`). A computed property reading iterator state at the old construction point could therefore disagree with the stderr line by one token on every EOS-terminated generation. Moving the construction is observable-behavior-neutral: `.info` is yielded at the same point as before, and nothing in the cache store mutates `turboQuantCompressionCount` or `promptPrefillTime` (the only iterator state the constructor reads besides the new field — the compression counter increments only in `maybeQuantizeCacheForStep` on the decode path).

## Construction sites audited (grep `GenerateCompletionInfo(`)

- `Evaluate.swift:generateLoopTask` — the only site fed by a live iterator; now passes `iterator.nativeMTPStats`. Both `generate(...) -> AsyncStream<Generation>` and `generateTokensTask` funnel here, so both native-MTP dispatch paths are covered by one change.
- `Evaluate.swift` early-exit sites (iterator construction cancelled/failed) and the deprecated synchronous `generate` — no generation ran; default `nil` is correct.
- `SpecDecStream.swift` (3 sites) — draft-model spec-dec, never the native-MTP iterator; default `nil`.
- `BatchEngine.swift` — the solo native-MTP fast path forwards `.info` events from `generateTask` **unmodified** (`continuation.yield(generation)`), so the stats flow through it with zero BatchEngine changes. The remaining sites (batched decode loop, cancel/shutdown/error paths) never run the native-MTP iterator; default `nil`. No BatchEngine diff in this PR.

## Tests

Added to `MTPRuntimeFocusedTests` (all using existing harness pieces):
- a minimal `TokenIteratorProtocol` conformer proves the protocol-extension default is `nil` — the default every non-MTP iterator (`TokenIterator`, `SpeculativeTokenIterator`, `BlockDiffusionTokenIterator`) inherits;
- `NativeMTPGenerationStats` round-trips every field and is `Equatable`;
- `GenerateCompletionInfo` defaults the field to `nil` (also proves init source-compat);
- end-to-end over the existing `FocusedNativeMTPProbeTarget` stub: after a native-MTP generation through `BatchEngine.generate`, the completion info carries non-nil stats with `depth == 3`, `verifyCalls >= 1`, `outputTokens == generationTokenCount` (no stop token fires with the zero-logit stub), accepted-by-depth histogram summing to `verifyCalls`, and the `cacheMode` literal. The `[NativeMTP]` stderr line emitted by that run (`verifyCalls=1 outputTokens=4 arFallbackTokens=0 acceptedByDepth=3:1 bonus=1 ...`) matches the structured values for the fields the struct carries (with the histogram rendered sparsely).

## Verification run

- `swift build` at repo root: clean.
- `swift test --filter MTPRuntimeFocusedTests`: 71 tests, all 4 new tests pass. One failure, "shape-walk quantization uses hidden size for JANG_2K routed expert inputs", fails identically on clean main @ 0d3444ca (verified via stash) — pre-existing, unrelated.
- `swift test --filter SpecDecStreamTests`: 2 failures, both fail identically on clean main (verified via stash) — pre-existing in this local `swift test` environment, unrelated.
- `swift test --filter "BatchEngineTurboQuantTests|BatchEngineCompileWiringTests"`: 13 tests passed.
- `swift test --filter BatchEngineTests`: 12 tests passed.

### DIFF
--- a/Libraries/MLXLMCommon/Evaluate.swift
+++ b/Libraries/MLXLMCommon/Evaluate.swift
@@ -945,6 +945,7 @@ public protocol TokenIteratorProtocol: Sequence, IteratorProtocol where Element 
     var promptPrefillTime: TimeInterval { get }
     var promptTokenIds: [Int] { get }
     var turboQuantCompressionCount: Int { get }
+    var nativeMTPStats: NativeMTPGenerationStats? { get }
     mutating func storeCacheAfterGeneration(
         generatedTokenIds: [Int],
         includeGeneratedBoundary: Bool)
@@ -953,6 +954,7 @@ public protocol TokenIteratorProtocol: Sequence, IteratorProtocol where Element 
 extension TokenIteratorProtocol {
     public var promptTokenIds: [Int] { [] }
     public var turboQuantCompressionCount: Int { 0 }
+    public var nativeMTPStats: NativeMTPGenerationStats? { nil }
 
     public mutating func storeCacheAfterGeneration(
         generatedTokenIds: [Int],
@@ -3523,15 +3525,6 @@ private func generateLoopTask<Handler: TokenLoopHandler>(
             MLXPressGenerationProfile.dumpAndReset(
                 reason: "generation-end tokens=\(tokenCount)")
 
-            let info = GenerateCompletionInfo(
-                promptTokenCount: promptTokenCount,
-                generationTokenCount: tokenCount,
-                promptTime: promptTime + iterator.promptPrefillTime,
-                generationTime: generateTime,
-                stopReason: stopReason ?? .cancelled,
-                turboQuantCompressions: iterator.turboQuantCompressionCount,
-                unclosedReasoning: unclosedReasoning
-            )
             // Multi-tier cache: drain the final async token eval before cache
             // snapshot/store, then keep completion info behind the cache-store
             // drain. Local chat/tool consumers may start a post-tool decode as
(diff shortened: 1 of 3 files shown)


jjang-ai:
Landed as #256. Main had moved under this branch — 8 conflicts across `Evaluate.swift` and `NativeMTPTokenIterator.swift`, because `GenerateCompletionInfo` gained `turboQuantCacheTransition` and `toolCallProtocolFailure` since, and a `NativeMTPHybridWarmupMemo` now sits exactly where the stats struct was added. #256 keeps all three additions and places the struct beside the memo rather than inside it.

One semantic correction worth calling out: main now caps native MTP at `VMLX_MTP_DEPTH_CAP` (default 1), since depths above 1 only won on a deterministic counting prompt and measured slower on prose. So `depth` is the *effective* depth, not the requested one — a host asking for depth 3 reads 1. The doc comment and the test (which asserted 3) both now match shipped behaviour, and that request-vs-effective gap is arguably the most useful thing this surface exposes.

The construction reorder was verified rather than assumed: timings are captured before the move, `lastTurboQuantCacheTransition` is written per decode step and not during the cache store, and end-to-end runs on VibeThinker-3B and Bonsai-27b are byte-identical to main.

Separately, and not a problem with this PR: I could not populate the stats on any locally installed bundle. The `-CRACK-MTP` bundles ship MTP tensors but no `vmlx_mtp_tuning.json`, Ornith has metadata without weights, and the one bundle that does have a tuning artifact is a model type native MTP does not dispatch. Details in the #256 QA comment. Closing in favour of #256, credited to you.

Issue #133, opened 2026-07-11
Quantized SDPA masks with Float.leastNormalMagnitude (positive!) instead of most-negative finite — masked positions keep attention mass on multi-token quantized steps


mimeding:
Reading the affine quantized attention path at `0d3444ca` (following up your #124 pointer to `quantizedScaledDotProductAttention`): the masked-fill constant looks like a sign/magnitude slip.

`KVCache.swift:1716`, `:1720`, `:1729` fill masked positions with `MLXArray(Float.leastNormalMagnitude)` — the smallest **positive** normal (~+1.18e-38) — where the intent (and Python mlx-lm's equivalent, `finfo.min`) is the most negative finite value. Masked logits become ≈0 instead of −∞, so after softmax masked positions receive weight ≈ e⁰ relative to real scores — i.e. attention mass leaks to masked slots.

Reachability, from the code: the vanilla single-token decode path runs L=1 with mask `.none`, so it never fires there. It becomes live whenever a **multi-token step runs on an already-quantized cache** — e.g. spec-decode verification (`SpecDecRuntime.swift:167`) or any array/causal mask reaching the quantized branch. That's presumably why nothing has visibly misbehaved: the shipping consumers (GPT-OSS, MiMoV2Flash single-token decode) don't cross it.

We haven't measured runtime impact (would need a spec-decode-enabled affine run vs spec-decode-off, which we can do if useful) — filing on the code alone since the constant is checkable by inspection. If `leastNormalMagnitude` is deliberate for some Metal-side reason on the quantized branch, a comment there would save the next reader; otherwise it's a one-constant fix.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #137, opened 2026-07-11
fix(kv-cache): fill quantized attention masks with dtype finfo.min, not leastNormalMagnitude

mimeding:
This is the offered fix for #133. The three boolean mask paths in `quantizedScaledDotProductAttention` filled excluded logits with `Float.leastNormalMagnitude`, the smallest positive normal Float, so masked logits were effectively zero and could retain softmax mass. They now use `scores.dtype.finfo!.min`, matching Python mlx-lm's `finfo(scores.dtype).min`; this is finite rather than `-inf` because an entirely masked row of infinities produces a NaN softmax row.

The affected code is not reached by vanilla one-token decode with no mask. It is reachable for multi-token quantized-cache steps, including speculative-decode verification.

Each fill scalar is constructed explicitly in `scores.dtype` before `MLX.where`, avoiding promotion of bf16 or fp16 scores by a float32 scalar. The additive floating-point mask branches are intentionally unchanged: caller-provided additive masks already encode their intended bias values, so no fill happens there.

Focused KV-cache tests now cover causal-mask parity with `MLXFast.scaledDotProductAttention`, boolean-array-mask parity (fp16 and bf16) and zero masked-position weight, a fully masked row remaining finite (that test documents the finite-fill semantics rather than the constant change), and unchanged additive-mask routing. The fixtures use tiny tensors and a real `QuantizedKVCache`; they require no model downloads.

One adjacent observation from reviewing for other occurrences, left out of this PR to keep it verifiable: `Gemma3nText.swift:599` builds its sliding-window exclusion with the same `Float.leastNormalMagnitude` constant (for fp16 masks it underflows to exactly +0, so positions beyond the window keep an additive bias of 0), and `:854` uses it as an epsilon that also underflows to 0 in fp16. Happy to file that separately or fold it in here if you prefer.

Happy for you to take this over or adjust the constant if leastNormalMagnitude was intentional for a Metal-side reason.

### DIFF
--- a/Libraries/MLXLMCommon/KVCache.swift
+++ b/Libraries/MLXLMCommon/KVCache.swift
@@ -1713,11 +1713,11 @@ public func quantizedScaledDotProductAttention(
         let kIndices = MLXArray(0 ..< kL)
         let causalMask = greaterEqual(
             expandedDimensions(qIndices, axis: -1), expandedDimensions(kIndices, axis: -2))
-        scores = MLX.where(causalMask, scores, MLXArray(Float.leastNormalMagnitude))
+        scores = MLX.where(causalMask, scores, MLXArray(scores.dtype.finfo!.min, dtype: scores.dtype))
 
     case .array(let maskArray):
         if maskArray.dtype == .bool {
-            scores = MLX.where(maskArray, scores, MLXArray(Float.leastNormalMagnitude))
+            scores = MLX.where(maskArray, scores, MLXArray(scores.dtype.finfo!.min, dtype: scores.dtype))
         } else {
             scores = scores + maskArray
         }
@@ -1726,7 +1726,7 @@ public func quantizedScaledDotProductAttention(
         // Handle multiple mask arrays - just use the first one for simplicity
         if let maskArray = maskArrays.first {
             if maskArray.dtype == .bool {
-                scores = MLX.where(maskArray, scores, MLXArray(Float.leastNormalMagnitude))
+                scores = MLX.where(maskArray, scores, MLXArray(scores.dtype.finfo!.min, dtype: scores.dtype))
             } else {
                 scores = scores + maskArray
             }
--- a/Tests/MLXLMTests/KVCacheTests.swift
+++ b/Tests/MLXLMTests/KVCacheTests.swift
@@ -190,3 +190,149 @@ func testCacheListCopyIsIndependent() async throws {
         #expect(allClose(orig, saved).item(Bool.self))
     }
 }
+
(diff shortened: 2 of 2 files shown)


jjang-ai:
### QA — the tests genuinely catch the bug, and no family shifts

**Sabotage check first**, because a mask test that passes either way proves
nothing. Merged this branch onto main, then reverted ONLY
`KVCache.swift` (keeping the new tests) so `leastNormalMagnitude` was back in
all 3 fills:

```
✘ testQuantizedAttentionCausalMaskMatchesReference        — 2 issues
✘ testQuantizedAttentionBooleanArrayMaskMatchesReference  — float16:  2 issues
✘ testQuantizedAttentionBooleanArrayMaskMatchesReference  — bfloat16: 2 issues
✘ Test run failed with 6 issues
```

With the fill restored to `scores.dtype.finfo!.min`, all pass:

```
✔ testQuantizedAttentionCausalMaskMatchesReference
✔ testQuantizedAttentionFullyMaskedRowIsFinite
✔ testQuantizedAttentionBooleanArrayMaskMatchesReference (2 cases: float16, bfloat16)
```

So the coverage is real, and it catches the bug in both half-precision dtypes —
which is where the explicit `dtype:` construction matters, since a float32
scalar would have promoted the scores.

`testQuantizedAttentionFullyMaskedRowIsFinite` is the right third test: it pins
why this is `finfo.min` and not `-inf`, so nobody "tidies" it later into a NaN
softmax row.

**No family moved.** Full KVCache suite passes, and on the merged build:

| family | check | result |
|---|---|---|
| LFM2.5-VL-3B-JANG_4M | full variating pattern | `diff vs main -> IDENTICAL` (18 lines) |
| Bonsai-27b-Ternary | growing-chat cache | `HIT tier=paged matched=18/53` — passed |

The variating run crosses reasoning ON/OFF, text↔image↔text-after-image, tools
constant-head grow control (`HIT disk 22/38, 57% reused`) and turn 1 replayed

Identical output is the expected result and worth stating plainly: as you note,
the affected fills are not reached by vanilla one-token decode with no mask, so
these runs demonstrate the change is inert on the default path — the behavioural
proof is the sabotage above, not the bench.

One process note for anyone reproducing: don't `git checkout <pr> -- KVCache.swift`
onto today's main. That file has since gained `HybridPoolQuantizedSegment`
plumbing, so taking the PR's whole file drops it and the build fails with
`cannot find 'HybridPoolQuantizedSegment' in scope` — a stale binary then makes
a bench diff look clean. Merge the branch instead.

PR #149, opened 2026-07-15
feat: run schema-2 affine 1-bit JANG models natively

jjang-ai:
Adds production runtime support for /Users/eric/models/Bonsai-27b-1bit-JANG without widening its packed one-bit language weights in RAM.

Dependency:
- osaurus-ai/mlx#2 merged into the pinned integration branch at 9e01acd573a18540468160ccaffeb6fb566e891e

Runtime changes:
- validate and consume schema-2 per-tensor JANG quantization manifests
- retain native affine1/group-128 language weights and exact affine4/group-64 vision overrides
- embed the merged standard and NAX affine1 Metal kernels
- preserve Qwen3VL tool schemas for the streaming parser on text and media inputs
- release completed media working sets after a GPU fence in BatchEngine
- make all VL proof entry points use the production mmap-backed load policy
- add fail-closed telemetry, marker, memory, reasoning, tool, image, and video proof gates

Current source tests:
- fresh local MLX Python build: affine1 Metal dequantize/QMV-fast/QMV/QMM test passed
- six JANG affine1 release tests passed
- seven VLM processor release tests passed

Live release evidence on the real 4.4 GB 1-bit bundle:
- coherent text multi-turn: 44.52 and 44.35 tok/s; peak physical-footprint delta 35.7% of bundle size
- real image multi-turn: 34.9-39.8 tok/s; peak 70.4%; compile off/on coherent
- structured image cache: same-media disk restore HIT 99/99; different-media MISS
- real video multi-turn: 41.8 and 42.6 tok/s; peak 82.1%
- structured get_weather Tokyo tool event; no raw XML leakage
- Qwen reasoning closed naturally with visible Answer: 4., 39.2 tok/s, and no raw markers
- ternary Bonsai text/image/video regression rows remained coherent and below their memory gate

The rejected Swift 1-to-2-bit expansion and the old non-production VL loader are not included. Full evidence, rejected rows, and the still-pending Osaurus pin/UI proof are recorded in docs/BONSAI_27B_AFFINE1_READINESS_2026-07-14.md.

PR #151, opened 2026-07-15
Mark Hub.Repo Sendable for Swift 6.2 region isolation

Cruzglenn:
## Summary
- Mark `Hub.RepoType` and `Hub.Repo` as `Sendable` in the vendored swift-transformers Hub module.
- Fixes Xcode 27 / Swift 6.2 build failure when `VMLXHub` is compiled with `StrictConcurrency`:

```
HubApi.swift:849:51: error: sending 'self.repo' risks causing data races
```

`HubFileDownloader.download()` passes `repo` into an async `withTaskCancellationHandler` closure; the repo identity types are immutable `String`-backed values and are safe to send.

This is the same change as #71, rebased onto current `main` (`1ca4029` / #149).

## Test plan
- [x] Patch is the two-line Sendable conformance only
- [ ] `swift build --target VMLXHub` on Xcode 27
- [ ] Osaurus app build on Xcode 27

### DIFF
--- a/Vendors/swift-transformers/Sources/Hub/Hub.swift
+++ b/Vendors/swift-transformers/Sources/Hub/Hub.swift
@@ -77,7 +77,7 @@ public extension Hub {
     }
 
     /// The type of repository on the Hugging Face Hub.
-    enum RepoType: String, Codable {
+    enum RepoType: String, Codable, Sendable {
         /// Model repositories containing machine learning models.
         case models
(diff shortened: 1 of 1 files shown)


jjang-ai:
Duplicate of #71 — byte-identical diff, same blob hashes (`919bc53e..403e6533`). Landed via #257, credited to #71 as the earlier filing. Thanks for the report; the conformance is correct and is now on main.

PR #152, opened 2026-07-15
Fall back instead of failing on pre-schema-2 JANG manifests (restores loading for existing bundles)

rcfa:
## Summary

**#149 made a previously-ignored field fatal.** Every JANG bundle whose `jang_config.json` declares `tensor_quantization_manifest_schema: 1` now fails to load on `main`:

```
invalidConfig("unsupported tensor quantization manifest schema 1")
```

That is every bundle the current converter emits — schema 1 is what it stamps today.

## Why this looks unintended rather than a deprecation

Four things in the existing code say "fall back", not "reject":

1. **The function's own doc comment:** *"Shape inference remains the fallback for older bundles without a schema-2 manifest."* A schema-1 bundle **is** an older bundle without a schema-2 manifest — so by this contract it should fall back.
2. **The absent-key branch already returns `nil`** — pre-manifest bundles were explicitly considered. Only "manifest present, older schema" was missed.
3. **`948f03a` introduced this whole path** (`+211/−2` to JangLoader.swift). Before it, `tensor_quantization_manifest_schema` was never read, and schema-1 bundles loaded correctly via the legacy per-module / shape-inference route. Nothing was removed — a previously-ignored field simply became a hard error.
4. **#149 is additive throughout** ("run schema-2 affine 1-bit JANG models natively") and never mentions deprecating schema 1.

## The change

Split the guard so the two cases are distinguished:

- **unknown schema → `return nil`** → the documented shape-inference fallback (restores pre-#149 behaviour for every existing bundle)
- **schema 2 declared but payload missing → still throws**, now with a message that says which case it is instead of reporting it as an unsupported schema

Schema-2 consumption is untouched — Bonsai and any other schema-2 bundle behave exactly as before.

## Reproduction

`empero-ai/Qwythos-9B-Claude-Mythos-5-1M-Jang_6M` (`jang_config.json`: schema=1, count=250), same patches/toolchain, only the engine differs:

| engine | result |
|---|---|
| `main` (1ca4029) | ✗ `invalidConfig("unsupported tensor quantization manifest schema 1")` |
| `cee0f8e` (pre-#149) | ✅ loads, `gsm8k 1/1`, 11s |

The build is green either way — this is purely a load-path gate.

## Tests

Added to `JangAffine1RuntimeContractTests` (pure, no GPU/model needed — 7-line JSON fixture):

- **`pre-schema-2 manifest falls back instead of failing the load`** — a schema-1 manifest returns `nil` rather than throwing
- **`schema-2 declared without a manifest still fails closed`** — the fail-closed path is preserved

Verified red→green: without the fix the first test fails with exactly the production error above; with it, both pass. Your existing **`required malformed contract fails closed`** test also still passes — strictness is unchanged, only the degradation path is restored.

### DIFF
--- a/Libraries/MLXLMCommon/JangLoader.swift
+++ b/Libraries/MLXLMCommon/JangLoader.swift
@@ -866,11 +866,21 @@ public struct JangLoader: Sendable {
         else {
             return nil
         }
-        guard schema == 2,
-            let rawManifest = quantization["tensor_quantization_manifest"] as? [String: Any]
+        guard schema == 2 else {
+            // "Shape inference remains the fallback for older bundles without a schema-2
+            // manifest" (see the doc comment above). A bundle declaring a pre-2 manifest
+            // schema is exactly such a bundle: before this path existed the field was never
+            // read at all, and the legacy per-module / shape-inference route loaded those
+            // bundles correctly. Throwing here turns a previously-ignored field into a hard
+            // load failure for every pre-schema-2 bundle — including every schema-1 bundle
+            // the current converter emits. Fall back instead; only schema 2 is consumed here.
+            return nil
+        }
+        guard let rawManifest = quantization["tensor_quantization_manifest"] as? [String: Any]
         else {
+            // A declared schema-2 manifest that is missing its payload IS malformed: fail closed.
             throw JangLoaderError.invalidConfig(
-                "unsupported tensor quantization manifest schema \(schema)")
+                "schema-2 tensor quantization manifest declared but tensor_quantization_manifest is missing")
         }
         if let declaredCount = quantization["tensor_quantization_manifest_count"] as? Int,
             declaredCount != rawManifest.count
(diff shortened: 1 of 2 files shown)


rcfa:
Closing — superseded, and your fix is better than mine.

Current `main` now does `guard schema == 1 || schema == 2` and **consumes** schema-1 manifests (gated on `quantization_scheme == "asymmetric"` + `quantization_backend == "mx.quantize"`). This PR only proposed *falling back* to shape inference for pre-schema-2 bundles; reading the manifest is strictly better.

Confirmed against our zoo: every schema-1 bundle we hold declares exactly that scheme/backend pair — `JANGQ-AI/Ornith-1.0-35B-JANG_6M`, `cais/HarmBench-Llama-2-13b-cls-Jang_6M`, `prometheus-eval/prometheus-8x7b-v2.0-Jang_6M` — so they load on `main` again and now get their manifests read rather than shape-inferred.

For context on why this mattered from the outside: schema-1 rejection had pinned us to `cee0f8e` since mid-July, because `main` rejected our entire (schema-1) zoo while `cee0f8e` predates the schema-2 path and could not read the newer JANG bundles — neither engine handled everything. We are now bumped to `a9118c4` with the whole zoo loading, and a JANG-vs-official-MLX quantizer comparison that was confounded by the schema-2 fallback is now a fair test. Thanks for sorting it at the root.

PR #159, opened 2026-07-19
Fix JANG semantic-width quantization precedence

mimeding:
## Summary

- prefer exact language `hidden_size` geometry for `embed_tokens`, `lm_head`, and known hidden-input projections before stale but width-colliding per-path metadata
- keep exact tensor manifests first and keep valid declarations when the semantic width is absent or cannot be represented by the packed tensors
- stop treating unrelated `*embed` paths such as `visual.pos_embed` as language anchors
- make aliased `.skip` entries deterministic and cover both `shared_expert` and `shared_experts` gate/up spellings
- add focused regressions for anchor, manifest, alias, vision, routed-expert, and shared-expert precedence

## What went wrong

The affected model contains stale per-path quantization settings that still look valid because they match a different model dimension. The loader accepted those settings before the projection's known 3,072-wide input, expanded some weights to 1,536 columns, and crashed on the first forward pass. This change makes the known semantic input width win when the packed tensor can represent it, while keeping exact tensor manifests first.

## Affected model and reproduction

The failure was reproduced with [`dealignai/MiniMax-M2.7-JANG_3L-CRACK`](https://huggingface.co/dealignai/MiniMax-M2.7-JANG_3L-CRACK/tree/1f37a1f2c244c152b1e20bba7a7414d025ebfc01), pinned at revision `1f37a1f2c244c152b1e20bba7a7414d025ebfc01`. The pinned download is 95,169,214,157 bytes across 126 weight shards, so allow at least 100 GB of free disk space. Validation used macOS on a 128 GB Apple-silicon Mac; lower-memory systems were not tested.

The following commands reproduce the failure on the PR's public base commit and then verify the fix. They require Xcode, `gh`, and `uvx`:

```bash
MODEL_DIR="$(
  env HF_XET_HIGH_PERFORMANCE=1 HF_HUB_DOWNLOAD_TIMEOUT=120 \
    uvx --from huggingface_hub hf download \
    dealignai/MiniMax-M2.7-JANG_3L-CRACK \
    --revision 1f37a1f2c244c152b1e20bba7a7414d025ebfc01
)"

run_repro() {
  local cache_dir
  cache_dir="$(mktemp -d /tmp/vmlx-pr159-repro-cache.XXXXXX)"
  caffeinate -dims .build/debug/mlxpress "$MODEL_DIR" \
    'Reply with exactly OK.' \
    --max-tokens 4 --temp 0 --thinking off \
    --expect OK --min-visible-chars 1 --min-generation-tokens 1 \
    --disk-cache-dir "$cache_dir" \
    --activity-gate 40 --print-memory
}

# Reproduce the failure on the PR base.
git switch --detach 2bfa9caffb951b14a76c869353abe1ff230668bd
git submodule update --init --recursive
swift build --product mlxpress
./Scripts/prepare-mlx-metal.sh
run_repro

# Verify the fix on this PR.
gh pr checkout 159
git submodule update --init --recursive
swift build --product mlxpress
./Scripts/prepare-mlx-metal.sh
run_repro
```

Observed base result at `2bfa9caffb951b14a76c869353abe1ff230668bd`:

```text
Fatal error: [gather_qmm] Last dimension of first input with shape (..., 3072)
does not match the expanded quantized matrix (1536, 1536) computed from shape
(256,1536,288) with group_size=64, bits=6 and transpose=true
```

The base run exited 133 before producing a token. The fixed run at `6268889066384c373ef3841d3a9acda7f8fd40b3` corrected 186 projections from 6-bit/group-64 to 3-bit/group-128 and generated `OK` without the first-forward failure.

## Proven failure mechanism

MiniMax-M2.7 JANG_3L exposed two instances of the same loader error.

For `embed_tokens`, a packed width of 384 with 24 scale groups can describe either 4-bit/group-128 at the required 3,072 hidden width or stale 8-bit/group-64 metadata at the unrelated 1,536 intermediate width.

For routed `gate_proj` and `up_proj`, packed tensors shaped `[256, 1536, 288]` with 24 scale groups require 3-bit/group-128 for their 3,072-wide hidden input. Stale 6-bit/group-64 metadata also looks self-consistent because it expands to the unrelated 1,536 expert width.

Four evidence classes support the precedence error as the mechanism:

1. **Source:** the old order accepted a path declaration when its unpacked width appeared anywhere in the model-wide `validInDims` set, before applying the projection's known input width.
2. **Arithmetic:** `288 * 32 / 6 = 1536` with `1536 / 24 = 64`, while `288 * 32 / 3 = 3072` with `3072 / 24 = 128`.
3. **Runtime:** the real model failed on first forward when `gather_qmm` received a 3,072-wide activation and a matrix expanded as 1,536 from 6-bit/group-64.
4. **Intervention:** the width-collision regression returned 6/64 before the reorder and 3/128 after it. The exact committed tree then completed real generation without the `gather_qmm` failure.

Rivals considered include corrupt tensors, a wrong hidden-size hint, an exact manifest, MLXPress/mmap behavior, and a downstream/common-cause failure. The weight/scale geometry is internally consistent; the model source fixes the projection input at 3,072; no manifest overrides these paths; and changing only the precedence clears both the focused failure and first-forward crash. Residual: the live smoke is one short turn, so it proves this quantization geometry and first-forward path, not multi-turn production readiness.

## Precedence

1. exact tensor manifest
2. exact language hidden-width geometry for `embed_tokens` and `lm_head`
3. exact language hidden-width geometry for known hidden-input projections
4. valid path-specific metadata
5. valid generic or top-level declaration
6. existing specialized and shape-walk fallbacks

## Validation

- `swift test --filter MTPRuntimeFocusedTests`: **76/76 passed**
- counterfactual hidden-anchor test against the old implementation: **failed as expected** at 8/64 instead of 4/128
- width-collision test before the projection reorder: **failed as expected** at 6/64 instead of 3/128
- final width-collision coverage: routed gate/up plus plural shared-expert gate/up, all at 3/128
- SHA-labeled real MiniMax-M2.7 JANG_3L smoke on `6268889066384c373ef3841d3a9acda7f8fd40b3`: expected `OK`, 186 corrected 6/64 -> 3/128 projections, no first-forward failure, 8.0% peak physical-footprint ratio
- `git diff --check`: clean

A serial broader run passed its XCTest phase: **1,063 tests, 45 skipped, 0 failures** after excluding three observed failures outside this diff. The later Swift Testing phase reported failures in other source-contract/runtime tests and aborted in an unchanged prompt-tail test. The full package suite is therefore **not reported green**. The stale README assertion was also reproduced on `origin/main`; the remaining suite failures were not repaired or attributed to this PR.

## External review

- Grok 4.5 high found the missing plural `shared_experts` gate/up spelling; the regression failed before that addition and passed after it. Its clean-commit rereview found no remaining actionable issue.
- Fable max verified manifest priority, semantic-width priority, singular/plural shared-expert gate/up coverage, and skip/alias behavior. It found no demonstrated defect in those properties.

Fable raised two residual hypotheses: suffix-only matching could affect a quantized vision projection whose packed geometry also realizes the language hidden width, and other known-width output/down projections still remain declaration-first. A scan of the available JANG VLM bundles found no vision tensor that both matched this semantic suffix class and realized the language hidden width. No code change is made for either unproven mechanism. Shared-expert down projections also need a role-specific intermediate-width hint before plural matching can be promoted safely.

### DIFF
--- a/Libraries/MLXLLM/LLMModelFactory.swift
+++ b/Libraries/MLXLLM/LLMModelFactory.swift
@@ -1892,10 +1892,10 @@ public final class LLMModelFactory: ModelFactory {
             for: templateResolvedDir)
         async let tokenizerTask = tokenizerLoader.load(from: tokenizerDirectory)
 
-        // When JANG, skip config.json's perLayerQuantization — JANG infers correct
-        // per-layer bits from tensor shapes. This avoids creating QuantizedLinear at
-        // the wrong bit width (which can't be re-quantized later).
-        // BUT: still pass `quantization` (the global config.json group_size /
+        // For JANG, pass config.json's per-layer metadata as declared evidence;
+        // loadWeights validates it against exact manifests, semantic widths, and
+        // tensor geometry before constructing QuantizedLinear modules.
+        // Also pass `quantization` (the global config.json group_size /
         // bits) so JangLoader.inferPerLayerQuantization gets the correct
         // `knownGroupSize` even when jang_config.json doesn't carry quant
         // metadata (e.g. DSV4-Flash bundles ship `weight_format: "bf16"`).
@@ -1903,10 +1903,9 @@ public final class LLMModelFactory: ModelFactory {
             modelDirectory: modelDirectory, model: model,
             quantization: jangConfig != nil ? baseConfig.quantizationContainer?.quantization : nil,
             // 2026-04-28: pass perLayerQuantization through even when JANG;
-            // loadWeights merges config.json's explicit per-layer dict on
-            // top of the shape walk to fix mis-inference of (bits, gs) pairs
-            // that share a packed shape (e.g., (4,64) ≡ (8,32)). Closes
-            // Cascade-2 JANG_4M / Nemotron-Omni MXFP4 first-prefill rmsNorm.
+            // loadWeights treats config.json's explicit per-layer dict as
+            // declared evidence, then validates it against exact manifests,
+            // semantic widths, and packed tensor geometry.
             perLayerQuantization: baseConfig.perLayerQuantization,
             jangConfig: jangConfig,
             loadPreservedMTP: loadNativeMTP)
--- a/Libraries/MLXLMCommon/JangLoader.swift
+++ b/Libraries/MLXLMCommon/JangLoader.swift
@@ -2181,29 +2181,40 @@ public struct JangLoader: Sendable {
             return nil
         }
 
-        func declaredQuantization(for basePath: String) -> BaseConfiguration.Quantization? {
-            func variants(_ key: String) -> [String] {
-                var out = [key]
-                if key.contains(".attn.") || key.hasSuffix(".attn") {
-                    out.append(key.replacingOccurrences(of: ".attn.", with: ".self_attn."))
-                    if key.hasSuffix(".attn") {
-                        out.append(String(key.dropLast(".attn".count)) + ".self_attn")
-                    }
+        func declaredPathVariants(_ key: String) -> [String] {
+            var out = [key]
+            if key.contains(".attn.") || key.hasSuffix(".attn") {
+                out.append(key.replacingOccurrences(of: ".attn.", with: ".self_attn."))
+                if key.hasSuffix(".attn") {
+                    out.append(String(key.dropLast(".attn".count)) + ".self_attn")
                 }
-                if key.hasPrefix("language_model.model.") {
-                    out.append(String(key.dropFirst("language_model.".count)))
-                } else if key.hasPrefix("language_model.") {
-                    out.append(String(key.dropFirst("language_model.".count)))
-                } else {
-                    out.append("model.\(key)")
-                    out.append("language_model.\(key)")
-                    out.append("language_model.model.\(key)")
-                }
-                return Array(Set(out))
             }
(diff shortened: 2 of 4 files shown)


jjang-ai:
Landed as #255 — `main` had moved under this branch (it now has a centralized `explicitQuantizationOption` / `layerPathCandidates` that already normalizes `model.` prefixes, attention aliases and the DSV4 shared-expert rename), so the three conflicts here are in `declaredPathVariants` / `declaredSpecificQuantization`, which duplicate it.

#255 keeps main's helper and carries forward the parts main genuinely lacks: the **precedence** fix (exact geometry before declared metadata), the **anchor tightening** so `vision_tower.pos_embed` is no longer treated as a language hidden anchor, and your alias-skip rule re-expressed as `declaresSkipForAnyAlias`. Your test for that rule is kept and passes.

Worth recording since it nearly sank the anchor change: DSV4-Flash's language embedding is literally named `embed` and has no `lm_head`, so restricting the anchor to `embed_tokens`/`lm_head` looked like it would drop the flagship's embedding. It does not — the shape walk sees runtime paths and DSV4's arrives already renamed to `model.embed_tokens`, confirmed by dumping it. Full per-layer quant maps diffed across 5 bundles / 1,480 layers: zero differences. Closing in favour of #255, credited to you.

PR #162, opened 2026-07-20
Fix Zaya XML parameter resynchronization

jjang-ai:
## Root cause

AppleScript 8B JANG_6M sometimes emits the next Zaya XML parameter before closing the current parameter. The live diagnostic decoded the malformed payload as:

    verb = "open\n<parameter=app>\nTextEdit"

The attribute-style XML parser consumed the recognized app parameter into verb and then skipped app. Osaurus correctly rejected the contaminated verb enum. This is a parser resynchronization defect triggered by malformed model XML, not content-delta streaming and not a need to accept arbitrary verb aliases.

## Change

- Resynchronize only when a nested parameter opener names a different parameter declared by the active tool schema.
- Keep unknown tag-looking text literal.
- Keep callers without a schema on the previous strict behavior.
- Add the exact AppleScript Computer Use regression and a literal-text negative control.

## Evidence

- Current-main isolated Release UI reproduced the TextEdit Computer Use failure in 13.3 seconds with the invalid verb enum error.
- Direct diagnostic with enum validation removed exposed the contaminated verb above.
- Red test reproduced the exact contaminated verb and missing app before the source change.
- After the change, all 9 Zaya XML extraction tests pass.
- Neighboring Zaya/XML/Qwen/Step/Gemma/streaming sweep: 79 relevant rows passed. One unrelated LFM2 newline assertion also fails unchanged on main and is not caused by this diff.
- git diff --check passes.

## Live gate

PARTIAL: the patched vMLX revision still needs to be pinned into an isolated Osaurus Release build and the TextEdit, Calculator, Safari, completion, failure, and feedback-only rows must be rerun visually before merge.

### DIFF
--- a/Libraries/MLXLMCommon/Tool/Parsers/XMLFunctionParser.swift
+++ b/Libraries/MLXLMCommon/Tool/Parsers/XMLFunctionParser.swift
@@ -140,7 +140,41 @@ public struct XMLFunctionParser: ToolCallParser, Sendable {
                     of: "</parameter>", range: nameEnd.upperBound ..< paramSection.endIndex)
             else { break }
 
-            var paramValue = String(paramSection[nameEnd.upperBound ..< paramEnd.lowerBound])
+            // A quantized ZAYA row can start the next attribute-style
+            // parameter without closing the current one:
+            //
+            //   <parameter=verb>open
+            //   <parameter=app>TextEdit</parameter>
+            //
+            // The old scan consumed the second opener/value into `verb` and
+            // then skipped `app` entirely. Resynchronize only when the nested
+            // opener names a *different parameter declared by this tool's
+            // schema*. Unknown tag-looking text remains literal, and callers
+            // without a schema keep the strict historical behavior.
+            let nestedParameterStart = paramSection.range(
+                of: "<parameter=", range: nameEnd.upperBound ..< paramEnd.lowerBound)
+            var valueEnd = paramEnd.lowerBound
+            var nextSearchStart: String.Index? = nil
+            if let nestedParameterStart,
+                let nestedNameEnd = paramSection.range(
+                    of: ">",
+                    range: nestedParameterStart.upperBound ..< paramEnd.lowerBound)
+            {
+                let nestedName = String(
+                    paramSection[nestedParameterStart.upperBound ..< nestedNameEnd.lowerBound]
+                ).trimmingCharacters(in: .whitespacesAndNewlines)
+                if nestedName != paramName,
+                    isDeclaredParameter(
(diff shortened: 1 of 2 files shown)


jjang-ai:
## Patched Release UI + request-shape A/B (2026-07-20 PDT)

**Verdict remains PARTIAL / LIVE FAIL. Do not merge this draft yet.**

Exact candidate:
- Osaurus Release app: `/private/tmp/Osaurus-Zaya-XML-Parameter-Recovery.app`
- bundle id: `com.dinoki.osaurus.bonsai2041proof`
- executable SHA-256: `e86c470c0516afa9821a9b87f50291db0b7d4734e1b1a66345cc3ee9c43e6919`
- vMLX revision resolved in the app: `5d86dbf4bccdf6d671defe585d79cf43a6d39f50`
- helper selected through live agent Settings: `JANGQ-AI/AppleScript-8B-JANG_6M`
- parent: `Ornith 1.0 9B JANG_4M`
- visible parent Thinking toggle: Off
- Computer Use enabled; native AppleScript disabled
- Accessibility visibly Granted

Three fresh, visually operated TextEdit Computer Use runs were **0/3**:
1. Final rejection: unexpected top-level `deps`.
2. Proposed an unrelated Finder `press return` action; after real-user decline, final rejection: `open needs an app name`.
3. Proposed unrelated Warp actions (`click mark 12`, `click mark 2`, then type `[id="address"]`); after real-user declines, final rejection: unexpected `role` where only `describe, mark` were allowed.

TextEdit never became frontmost. The original contaminated `verb/app` shape did not recur in these three patched runs, but the user-facing report is not resolved.

Source trace:
- Computer Use calls `ChatEngine.completeChat` with `stream: false`, one forced named `agent_action`, and the strict schema. This excludes content-delta assembly as the owner of these rows.
- The child request does not set `enable_thinking`; this Zaya bundle's template defaults the missing kwarg to true. The visible parent Thinking toggle therefore did not prove child thinking was off.
- Controlled API A/B then set `enable_thinking:false` explicitly. With the full `agent_action` schema, Zaya still omitted/invalidated `verb`. With a compact open-only schema, the same JANG_6M emitted valid `{"verb":"open","app":"TextEdit",...}` calls in 3/3 trials at 41.3–41.7 tok/s.

Current attribution: the parser resynchronization fix is source/test-valid for the exact malformed XML it targets, but this model is not reliable against Osaurus's full generic Computer Use schema. Reasoning-off alone does not fix it. A parser-family-specific compact action contract may be viable, but that would be a separate, broader runtime change requiring full Computer Use regression and live proof; it will not be smuggled into this narrow parser PR.

PR #165, opened 2026-07-20
guard qwen3.5 gated-delta-net against incompatible restored cache shapes

RaajeevChandran:
## Summary

- Production crash reports show a fatal precondition failure in `getItemND` (MLXArray subscript rank check), reached from `GatedDeltaNet.callAsFunction` via `TokenIterator.prepareRemainder` in the batch engine solo fast path.
- The slice at the crash site always operates on a rank-3 array when state is built fresh, so the only way to trap there is a `MambaCache` slot restored with a stale or mismatched shape (paged/hybrid cache restore, prefix commit).
- This validates both cache slots before use in `GatedDeltaNet` (MLXVLM) and `Qwen35GatedDeltaNet` (MLXLLM):
  - Conv slot must be rank 3 with matching batch size and `convDim`.
  - Recurrent slot must be rank 4 with shape `[B, numVHeads, headVDim, headKDim]`.
- An incompatible slot is discarded and generation restarts from zero linear-attention state instead of aborting the process.
- A one-shot warning logs the discarded slot name and shape so the upstream restore bug is not silently masked.

## Notes

- This is a defensive guard; the producer of the bad shape is somewhere in the cache restore path, and the new log line will fingerprint it if it fires again.
- One degraded generation after a bad restore is the accepted tradeoff versus a hard crash on launch.

## Test plan

- Build MLXLLM and MLXVLM.
- Run a Qwen3.5 generation with cache reuse and confirm output is unchanged.
- Optionally corrupt a cache slot shape in a debug build and confirm the reset path logs and continues instead of trapping.

## Checklist

Put an `x` in the boxes that apply.

- [ ] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document
- [ ] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/Libraries/MLXLLM/Models/Qwen35.swift
+++ b/Libraries/MLXLLM/Models/Qwen35.swift
@@ -254,6 +254,22 @@ final class Qwen35GatedDeltaNet: Module {
         super.init()
     }
 
+    /// One-shot (per process) diagnostic for a discarded incompatible cache
+    /// slot — the reset self-heals, but a silent reset would hide the
+    /// upstream restore bug that produced the bad shape.
+    private static let discardedCacheWarningLock = NSLock()
+    private nonisolated(unsafe) static var didWarnDiscardedCache = false
+
+    static func warnDiscardedCacheState(slot: String, shape: [Int]) {
+        discardedCacheWarningLock.lock()
+        defer { discardedCacheWarningLock.unlock() }
+        guard !didWarnDiscardedCache else { return }
+        didWarnDiscardedCache = true
+        print(
+            "[Qwen35] GatedDeltaNet discarded incompatible \(slot) cache state "
+                + "(shape \(shape)); resetting linear-attention state")
+    }
+
     func callAsFunction(
         _ inputs: MLXArray,
         mask: MLXArray? = nil,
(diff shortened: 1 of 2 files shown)


jjang-ai:
### QA — merges clean on today's main, inert on the normal path, and I added the missing coverage

Test-merged onto `vmlx-origin/main` (after #242/#243/#246/#247/#248 landed):
**no conflicts**, 89 insertions across the two `Qwen35.swift` copies. Main has no
equivalent guard today, so this is not redundant.

**Read the logic rather than trusting the summary** — it is correct, including
the subtle part: the warning is inside `else { if let cacheState = cache?[0] }`,
so it fires only when a slot was *present and malformed*, and stays silent on a
cold start with no cache. Valid state takes `convState = cacheState` untouched.

**Families this touches, all measured on the merged build:**

| family | architecture | result |
|---|---|---|
| Bonsai-27b-Ternary | `Qwen35` hybrid SSM | `HIT tier=paged matched=18/53` — passed |
| Ornith-1.0-9B-MXFP8 | `Qwen35` hybrid SSM | `matched=18/57` — unchanged from main |
| LFM2.5-VL-3B-JANG_4M | full variating pattern | `diff vs main -> IDENTICAL` (18 lines) |

The variating run crosses reasoning ON/OFF, text↔image↔text-after-image, tools
constant-head grow control (`HIT disk 22/38, 57% reused`) and turn 1 replayed

The guard never fired in any run — expected, since it defends a rare restore
path. So what these prove is the other half: it is **inert** on healthy state.

Two honest notes:

1. **Ornith-1.0-9B still fails the growing-chat gate** (`probe hit did not reach
   the safe prompt/cache boundary: matched=18`) on both main and this branch,
   while Bonsai — same architecture — passes. This PR neither causes nor fixes
   it; I am tracking it separately.
2. **This shipped with no tests.** I pushed
   `Qwen35GatedDeltaNetCacheGuardSourceTests` (4/4) to `pr165-with-tests`,
   pinning that the guard exists in **both** copies. Qwen35 is duplicated across
   MLXLLM and MLXVLM, and patching one copy while the other keeps crashing is a
   trap this repo has hit before — it is exactly how the MLX test-lock wedge
   stayed hidden. The suite also pins that an absent cache is not reported as a
   discarded slot, and that the warning is one-shot rather than per-token.

PR #167, opened 2026-07-21
Restore safe hybrid SSD prefix reuse without cache rewrites

jjang-ai:
## Root cause

Hybrid Qwen/Ornith requests set skipExactDiskBoundary, but the indexed disk-candidate fallback re-admitted the exact token count. The runtime restored the exact GDN/KV entry, could not satisfy the required earlier recurrent seed, discarded the restore, and visibly prefetched from zero. Warm requests also synchronously rewrote already-validated KV and SSM payloads.

## Changes

- exclude the exact disk boundary from every candidate source when the path-dependent cache contract requires it
- skip KV rewrites only after current-process deserialize/write validation and matching file/index metadata
- apply the same current-process validation to paired SSM tensor/sidecar entries, including completeness, state count, boundary, model key, and linked KV hash
- preserve healing writes for inherited, changed, corrupt, missing, or unindexed payloads
- add focused regression coverage and the live proof matrix

## Current evidence

- 7 disk-cache tests passed
- 10 SSM-state tests passed
- 29 cache-topology tests passed, including Gemma rotating/TQ, Nemotron Mamba/TQ, DSV4 typed disk, ZAYA CCA, paged eviction, and paged-off partial SSD restore
- prior isolated Release Osaurus UI reproduced the exact failure with Disk Cache visibly On: the trace hit boundary 2234 with skipExactDisk=true, then the UI prefetched from 0

## Gate

PARTIAL: source tests pass, but the candidate is not being called user-facing verified until the rebuilt Release Osaurus app visibly restores a safe partial boundary, processes only the suffix, stays coherent, and emits the validated KV/SSM skip telemetry. That UI run is in progress.

### DIFF
--- a/Libraries/MLXLMCommon/Cache/CacheCoordinator.swift
+++ b/Libraries/MLXLMCommon/Cache/CacheCoordinator.swift
@@ -554,6 +554,17 @@ public final class CacheCoordinator: @unchecked Sendable {
             }
 
             for boundary in diskCache.candidateTokenCounts(maxTokens: tokens.count) {
+                // `skipExactDiskBoundary` is a correctness requirement for
+                // path-dependent hybrid caches, not merely a preference for
+                // the first two probes above. The indexed fallback used to
+                // re-admit `tokens.count`, so Qwen 3.5 / Ornith restored an
+                // exact GDN boundary, failed to find the N-1 seed state, and
+                // discarded the restore for a full prompt prefill. Keep the
+                // exact boundary excluded across every probe source so the
+                // longest safe partial boundary is selected instead.
+                guard !skipExactDiskBoundary || boundary != tokens.count else {
+                    continue
+                }
                 guard tried.insert(boundary).inserted else { continue }
                 if let hit = diskHit(boundary: boundary) {
                     return hit
--- a/Libraries/MLXLMCommon/Cache/DiskCache.swift
+++ b/Libraries/MLXLMCommon/Cache/DiskCache.swift
@@ -11,6 +11,7 @@ public struct DiskCacheStats: Sendable {
     public let hits: Int
     public let misses: Int
     public let stores: Int
+    public let storeSkips: Int
     public let maxSizeBytes: Int
 }
 
@@ -69,6 +70,11 @@ public enum MLXCacheIOLock {
 /// likewise synchronous since they typically feed directly into model inference.
 public final class DiskCache: @unchecked Sendable {
 
+    private struct ValidatedFileFingerprint: Equatable {
+        let size: Int
+        let modificationDate: Date
+    }
+
     // MARK: - Properties
 
     /// Root directory for cache files and the SQLite index.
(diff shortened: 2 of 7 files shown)


jjang-ai:
Live Release-app proof update (2026-07-20, M5 Max)

Exact app and graph

- App: `/private/tmp/Osaurus SSD Cache Proof 20260720.app`
- Isolated bundle id: `com.dinoki.osaurus.ssdcacheproof20260720`
- Release executable SHA-256: `8dcb022e282b28a2b800a7cdef858a86f300c3999db12c1fdb97f67f24ba3516`
- Xcode SourcePackages checkout: exact PR head `74caefd907e6df15780a454d8523b78bf889964c`
- Local model: `/Users/eric/models/JANGQ-AI/Ornith-1.0-9B-MXFP8`

Computer Use drove fresh onboarding, Settings, model selection, two new chats, a full quit, and process restart. Fresh preferences visibly showed Prefix Cache On, Disk Cache On, GPU paged cache Off, Codec Engine Selected, SSM rederive On; Ornith Thinking was visibly Off.

Observed runtime and UI rows

1. First user turn after selection: `HIT disk boundary=1734 remaining=18 ssm=48 fmtV=2 tokens=1752 skipExactDisk=true`. Visible exact answer `CACHE PROOF ONE.`; TTFT 0.50s, 49.0 tok/s, 5 tokens; no reasoning deltas.
2. Brand-new chat in the same process: warmup itself logged `HIT disk boundary=1731 remaining=3 ssm=48 fmtV=2 tokens=1734 skipExactDisk=true`. Write-through emitted `disk-store SKIP validated` for the 1,734- and 1,731-token KV payloads and `ssm-store SKIP validated` for both matching 48-state companions. Repeated exact answer; TTFT 0.33s, 49.1 tok/s, 5 tokens.
3. Full app quit/restart with only the isolated root/SSD cache preserved: warmup again logged the 1,731/1,734 disk+SSM hit and the UI reached Model warm without zero-prefix prefill. First post-restart answer was exact `CACHE PROOF RESTART.`; TTFT 0.39s, 49.4 tok/s, 6 tokens.

This is current source-trace plus live UI evidence that the reported Ornith/Qwen-hybrid new-session warmup rollback is fixed by this PR: the cache restores a safe partial KV+SSM boundary and computes only the suffix. It is not a family-wide cache claim. Gemma 4 rotating-SWA, Bonsai, Qwen VL/media salt, paged-On hot-to-SSD fallback, TurboQuant-On, and configuration-change invalidation remain open Release-UI rows.

One conservative behavior is recorded explicitly: a fresh process restores and validates N-3 because exact N is unsafe for this GDN topology, then rewrites exact N once after warmup; the restored N-3 payload is skipped, and same-process repeated KV/SSM write-throughs are skipped. This does not cause zero-prefix prefill, but independent read-validation of the exact non-restored payload is a possible follow-up if we want to remove that remaining once-per-process write safely.

PR #170, opened 2026-07-21
Block-diffusion: guard against an empty (all-EOS) first canvas, and seed the three RNG draws

rcfa:
On terse prompts — e.g. a bare `Respond with one of: yes, no.` — block-diffusion frequently denoises the **entire first canvas to EOS**, so the model emits nothing. It isn't a comprehension failure: the same model answers correctly whenever it emits anything, and its explain-then-answer MC is strong. But an empty response to a direct question is a degenerate output for a server / agentic caller.

**Fix:** on the first canvas, forbid EOS at position 0 (mask those logits in the *sampled* copy only) so at least one content token is produced. Later canvases end normally, so a complete reply can still stop. Self-conditioning keeps the unmasked logits, so denoising dynamics are unchanged — and it is a no-op whenever position 0 would already be content.

**Validated:** a terse yes/no benchmark goes from ~16% (all-empty) to ~62%; open-ended generation (gsm8k etc.) is unchanged. (The residual isn't the mask — when forced off EOS the model tends to emit its channel-format scaffolding rather than a bare token, which is a real property of the model, not something this guard should paper over.)

**Depends on the seeding change** (same code region, so it can't be a standalone branch): this branch carries that commit first. When the seeding PR merges, this rebases cleanly and that commit collapses to a no-op.

**It's a decode policy**, so it's your call whether it should be unconditional or flag-gated. The evidence — a terse-prompt collapse to *silence*, recoverable to a real answer — argues for unconditional, but I'm happy to gate it behind a config flag if you'd prefer.

---

### Update: the aliasing note was correct, and the suggested remedy breaks the guard

You were right on every premise. `MLXArray` is a `final class`, subscript assignment goes through
`_updateInternal` → `mlx_array_set`, and `var sampled = processed` therefore aliases: the mask does
reach `selfConditioning = processed`. The comment claiming self-conditioning kept the UNMASKED
logits was false.

I applied the additive form you suggested, then measured it. Same binary, same 8 items,
`diffusiongemma-26B-A4B-it-MXFP4`, boolq gen-nothink:

| | boolq |
|---|---|
| out-of-place (`sampled = processed + bias`) | **0/8 = 0.0%**, 8/8 extraction failures |
| in-place (original) | **6/8 = 75.0%**, 1/8 extraction failures |
| in-place, banked full run (n=50) | 35/50 = 70.0% |

Making the stated invariant true is what breaks the guard. The suppression reaching
self-conditioning is not incidental — it is the mechanism. Masking only the sampled copy lets the
model re-assert its EOS preference at the next denoising step, and the canvas collapses to all-EOS
again, which is exactly the failure this patch exists to prevent.

So I have taken your other option — keep the in-place write, drop the false claim — and the comment
now states what the code does, why the aliasing is deliberate, and carries the 70% vs 0% numbers so
the next reader does not "fix" it back. Your review found a genuine defect; it was in the comment
rather than the code.

Also noted for the description, as you asked: this diff additionally threads a per-request PRNG key
through canvas init, the denoiser draw, and renoise, so seeded block-diffusion is reproducible.
Those three draws previously used the global unseeded RNG, which made temperature-0 runs
non-reproducible.

### DIFF
--- a/Libraries/MLXLMCommon/Diffusion/BlockDiffusionTokenIterator.swift
+++ b/Libraries/MLXLMCommon/Diffusion/BlockDiffusionTokenIterator.swift
@@ -376,10 +376,36 @@ public struct BlockDiffusionTokenIterator: TokenIteratorProtocol {
                 tMin: options.tMin, tMax: options.tMax)
             let processed = logits.asType(.float32) / temperature
 
+            // No-empty-response guard. Block-diffusion collapses terse prompts (e.g. boolq's
+            // "Respond with one of: yes, no.") to an all-EOS first canvas → empty output, even
+            // though the model answers correctly whenever it emits anything. On the FIRST canvas,
+            // forbid EOS at position 0 so at least one content token is produced; later canvases
+            // end normally, so a genuinely complete reply can still stop. (-1e9, not -inf, keeps
+            // entropy finite.)
+            //
+            // The write is IN PLACE, and that is load-bearing rather than accidental. `MLXArray` is
+            // a `final class`, so `var sampled = processed` binds a second reference to the same
+            // object and `sampled[0, 0, eos] = …` rewrites `processed` too (`_updateInternal` →
+            // `mlx_array_set`). The suppression therefore also reaches `selfConditioning =
+            // processed` below, and persists across denoising steps.
+            //
+            // That persistence is what makes the guard work. Building the mask out-of-place — e.g.
+            // `sampled = processed + bias` — leaves self-conditioning unmasked, the model re-asserts
+            // its EOS preference on the next step, and the canvas collapses again: MEASURED at
+            // boolq gen-nothink 70.0% (35/50, in place) versus 0.0% with 8/8 extraction failures
+            // (out of place). An earlier version of this comment claimed self-conditioning kept the
+            // UNMASKED logits; that was false, and making it true broke the guard.
+            var sampled = processed
+            if canvasesEmitted == 0 {
+                for eos in options.eosTokenIds {
+                    sampled[0, 0, eos] = MLXArray(Float(-1e9))
+                }
+            }
+
             let denoiserKey = nextRandomKey()
-            let denoiserCanvas = MLX.categorical(processed, key: denoiserKey).asType(.int32)
-            let argmaxCanvas = argMax(processed, axis: -1).asType(.int32)
-            let entropy = canvasTokenEntropy(processedLogits: processed)
+            let denoiserCanvas = MLX.categorical(sampled, key: denoiserKey).asType(.int32)
+            let argmaxCanvas = argMax(sampled, axis: -1).asType(.int32)
+            let entropy = canvasTokenEntropy(processedLogits: sampled)
             let acceptMask = entropyBoundAcceptMask(
                 tokenEntropy: entropy, entropyBound: options.entropyBound)
 


jjang-ai:
Reviewed. The guard's shape is right — first canvas only, position 0 only, and masking the sampled copy rather than the distribution is the correct instinct. One problem with how that separation is expressed, though.

## `var sampled = processed` does not make a copy

`MLXArray` is a `final class` (`Source/MLX/MLXArray.swift:7`), so that line binds a second reference to the *same* object. And subscript assignment mutates in place rather than rebinding:

```swift
// Source/MLX/MLXArray+Indexing.swift:348
self._updateInternal(scattered(indices: arrayIndices, updates: update, axes: axes))

// Source/MLX/MLXArray.swift:572
public func _updateInternal(_ array: MLXArray) {
    mlx_array_set(&self.ctx, array.ctx)
}
```

So `sampled[0, 0, eos] = …` rewrites the ctx of the object `processed` also points at. By the time control reaches

```swift
selfConditioning = processed
```

`processed` is the **masked** tensor. The comment's stated invariant — *"self-conditioning below keeps the UNMASKED `processed`, so the denoising dynamics stay natural"* — does not hold as written; self-conditioning gets `-1e9` at position 0's EOS on the first canvas.

The blast radius is small (one logit, first canvas only, and self-conditioning is a soft signal), which is probably why it validated fine — but it is the opposite of what the comment says the code does, and that is the kind of gap that outlives the PR.

The fix is to build `sampled` out-of-place so the two really are distinct arrays, e.g. an additive mask:

```swift
var sampled = processed
if canvasesEmitted == 0 {
    let bias = MLXArray.zeros(like: processed)
    for eos in options.eosTokenIds { bias[0, 0, eos] = MLXArray(Float(-1e9)) }
    sampled = processed + bias   // new array;  untouched
}
```

(`bias` is freshly constructed, so mutating *it* in place is fine.) Alternatively keep the in-place write and drop the claim about self-conditioning — but the additive form is what the comment already describes, so it seems the intended one.

## Unrelated, and worth its own line in the description

The diff also threads a per-request PRNG key through the canvas init, denoiser draw, and renoise, so seeded block-diffusion is reproducible. That is a real fix — the three draws previously used the global RNG unconditionally — but it is invisible from the title, and someone bisecting determinism changes later will not find it. Worth calling out in the PR body (or splitting), independent of the aliasing point above.

Happy to take this with the copy fixed.

jjang-ai:
Follow-up on the aliasing note — I swept the tree for the same pattern so you know the scope.

**Rebinding sites are fine.** `Evaluate.swift:495`, `Evaluate.swift:1029`, `BatchScheduler.swift:164`, `NativeMTPTokenIterator.swift:1762/1836` all do `logits = <expression>`, which assigns a new array to the local and leaves the original object alone.

**In-place writes already exist**, though, in the penalty processors — `Evaluate.swift:839, 867, 924, 981`, e.g.

```swift
public func process(logits: MLXArray) -> MLXArray {
    …
    logits[0..., indices] = selectedLogits   // writes through a `let` parameter
    return logits
}
```

That mutates the caller’s array (legal, since the parameter is a class reference). It is harmless **today** only because every caller immediately reassigns the return value and nobody retains a pre-penalty reference — I checked the spec-dec paths specifically, since verifying drafts against unpenalized logits is exactly where a retained copy would appear, and neither site keeps one.

So: no existing bug, but the codebase already relies on that convention silently. #170 is the first place where a comment promises the opposite, which is why it seemed worth being precise about rather than waving through.

rcfa:
You were right on every premise — I checked all three rather than taking them on trust. `MLXArray` is `public final class` (Source/MLX/MLXArray.swift:7), subscript assignment goes through `_updateInternal` → `mlx_array_set`, and `var sampled = processed` therefore aliases. The mask *does* reach `selfConditioning = processed`, and the comment claiming otherwise was false.

So I applied your additive form — and then measured it, which turned out to matter.

**Same binary, same 8 items, `diffusiongemma-26B-A4B-it-MXFP4`, boolq gen-nothink:**

| | boolq |
|---|---|
| out-of-place (`sampled = processed + bias`) | **0/8 = 0.0%**, 8/8 extraction failures |
| in-place (original) | **6/8 = 75.0%**, 1/8 extraction failures |
| in-place, banked full run (n=50) | 35/50 = 70.0% |

Making the stated invariant true is what breaks the guard. The suppression reaching self-conditioning is not incidental — it *is* the mechanism. Mask only the sampled copy and the model re-asserts its EOS preference at the next denoising step, so the canvas collapses to all-EOS again, which is precisely the failure this patch exists to prevent.

I have taken your other option: keep the in-place write, drop the false claim. The comment now states what the code actually does, why the aliasing is deliberate, and carries the 70%-vs-0% numbers so the next reader (or the next me) does not helpfully "fix" it back.

Your review found a genuine defect. It was in the comment rather than the code — which is arguably worse, since the comment is what a future reader would have trusted.

Also added to the description as you asked: the diff threads a per-request PRNG key through canvas init, the denoiser draw, and renoise. Those three draws previously used the global unseeded RNG, so temperature-0 block-diffusion runs were not reproducible.

On your follow-up sweep — thank you for scoping it. The penalty processors writing through a `let` parameter (`Evaluate.swift:839/867/924/981`) look like the same latent hazard: fine while every caller immediately reassigns, and silently wrong the moment one retains the pre-call array. Happy to file that separately if you want it tracked, since it is orthogonal to this PR.

jjang-ai:
That measurement is the answer, and it inverts my recommendation — thank you for running it instead of taking the review at face value. 0/8 with 8/8 extraction failures versus 6/8 is not a wash, and the mechanism you describe is the right reading: suppression has to reach self-conditioning or the model simply re-asserts EOS at the next denoising step and the canvas collapses again. Keeping the in-place write and deleting the false claim is the correct resolution. My "the invariant is violated" was accurate; my implied "so restore the invariant" was wrong, because the invariant itself was the bug.

Worth stating plainly for the archive: the code was right and the comment was wrong, which is the more dangerous arrangement of the two — exactly as you put it.

**On the penalty processors — yes please, file it separately, and it is latent rather than live.** I checked every call site before answering rather than leaving you with my hand-wave:

| site | shape | retains pre-call array? |
|---|---|---|
| `Evaluate.swift:1030-1034` | `logits = ctx?.process(logits: logits) ?? logits`, chained ×5 | no — pure rebinding |
| `Evaluate.swift:2264` | `var logits = logits[0..., -1, 0...]` then `processor.process` | no — writes land in the fresh slice, not the caller's full-logits array |
| `Evaluate.swift:3036, 3049` | same rebinding shape | no |
| `Evaluate.swift:3070` | `var draftLogits = draftResult.logits[0..., -1, 0...]`; `draftResult.logits` unused afterwards | no |

So nothing is wrong today. The hazard is that the contract is "callers must immediately reassign and must not retain", it is nowhere written down, and it is invisible at the call site — the next person to add a processor call that keeps the original (a pre/post-penalty comparison, a logprob capture, a debug dump) gets silently wrong numbers with no crash. That is the same failure mode this PR just walked into, one layer up.

The cheap version is a doc comment on `LogitProcessor.process` stating that it may mutate its argument in place and that the return value is the only valid handle afterwards. The thorough version is a test that retains the pre-call array and asserts it is untouched — which would fail today and turn the convention into something enforced. Happy either way; your call which one the issue asks for.

Also noted on the PRNG threading through canvas init / denoiser draw / renoise — that makes temperature-0 block-diffusion runs reproducible, which is what made the 0.0% vs 75.0% comparison above meaningful in the first place. Good that it landed in the same change.

jjang-ai:
Merged as #258, with a regression test added on top. Thanks — the diagnosis and the measurement are what made this reviewable.

I kept the in-place write exactly as you wrote it, including the reasoning about self-conditioning. What I added is `BlockDiffusionEOSGuardAliasingTests`, because the mechanism is invisible at the call site: the fix depends on `MLXArray` being a `final class`, and the "obvious cleanup" to an out-of-place mask would silently take the model back to producing nothing, with no test failing. The suite pins both directions — subscript assignment aliases, `MLX.where` over an `arange` does not — so the next person does not have to rediscover the difference by measuring boolq.

That distinction is worth having written down for another reason: elsewhere in this codebase the same aliasing was a *bug*, where a logit mask built by writing into a `full(-inf)` array rewrote the caller`s logits. Same language rule, opposite intent.

Not live-proven on my side, and I said so in the PR: `BlockDiffusionModel` is only implemented by `DiffusionGemmaModel` and no diffusion bundle is installed here, so I am relying on your boolq numbers for efficacy and on the change being confined to `BlockDiffusionTokenIterator` for safety.

PR #179, opened 2026-07-23
Recover schema-bound Qwen string-array tool arguments

jjang-ai:
## Problem

Bonsai/Qwen XML tool output can emit a schema-declared `array<string>` as an unquoted bracket list, for example `[skill/Autonomous Scheduler, skill/Data Keeper]`. The parser left that value as a scalar string, so Osaurus rejected `capabilities_load.ids` before execution.

The same emergency lane exposed two adjacent engine issues: split XML tool-call deltas needed the same schema-bound array recovery, Gemma decoded thought-channel openers needed to route into reasoning deltas, and Osaurus mutable DB/tool state could invalidate the full rendered system-message cache boundary even when the leading static system prefix was byte-identical and safe to reuse.

## Change

- retain the full parameter schema during XML conversion
- recover only an outer-bracketed, comma-delimited plain list when the declared schema is exactly `array<string>`
- leave object arrays, nested collection syntax, quotes, control characters, and ordinary strings on the existing validation path
- cover split Qwen XML array deltas
- route Gemma decoded thought-channel openers into reasoning
- add `UserInput.cacheStableSystemPrefix` and tokenizer-probed static-prefix boundaries for LLM, Gemma 4, and Qwen3VL processors

## Evidence

Source/local tests:
- `MLXLMTests.ToolTests/testXMLFunctionParser*`: 11 passed for the original array parser fix, including exact Bonsai arguments and negative controls
- `/tmp/vmlx-static-prefix-boundary-tests-20260723.log`: `CanonicalChatCacheBoundariesTests`, including `staticSystemHintAddsEarlierBoundaryInsideMutableSystemMessage`

Downstream Osaurus proof at this head (`91933aef193cab180b821bad1f4b4cc8ad753107`):
- focused source/pin test: `/tmp/osaurus-static-prefix-source-tests-91933aef-rerun-20260723.log`
- focused DB/tool-loop tests: `/tmp/osaurus-db-toolloop-static-prefix-91933aef-20260723.log`
- focused TurboQuant policy tests: `/tmp/osaurus-tq-default-policy-tests-91933aef-20260723.log`
- Release UI proof app: `com.dinoki.osaurus.staticprefixproof20260723`, executable SHA-256 `f110b825de75dfae41e288a6215080e5f1e3ba468f1f25ea1203434e8fc6d9ce`
- live Settings/Live Activity showed paged RAM KV Off, SSD cache On, `liveKVCodec=engine_selected`, `storedKVCodec=auto`, `TurboQuant compressions 0`
- live Gemma 4 12B MXFP8 and Bonsai 27B Ternary JANG CRACK fresh-chat rows both completed, unlocked input, and logged SSD disk restores; Bonsai showed hybrid/SSM restores with `ssm=96`

## Scope boundary

This PR does not claim all model families, AppleScript, Laguna, Qwen VL media, or full DB-tool dynamic-suffix workflows are verified. Those rows remain tracked in the downstream Osaurus gate doc.

### DIFF
--- a/Libraries/MLXLLM/LLMModelFactory.swift
+++ b/Libraries/MLXLLM/LLMModelFactory.swift
@@ -1224,7 +1224,8 @@ private struct LLMUserInputProcessor: UserInputProcessor {
                 messages: messages,
                 tools: input.tools,
                 additionalContext: additionalContext,
-                promptTokens: promptTokens)
+                promptTokens: promptTokens,
+                staticSystemPrefix: input.cacheStableSystemPrefix)
 
             return LMInput(
                 tokens: MLXArray(promptTokens),
--- a/Libraries/MLXLMCommon/ReasoningParser.swift
+++ b/Libraries/MLXLMCommon/ReasoningParser.swift
@@ -236,7 +236,8 @@ public struct ReasoningParser: Sendable {
         let gptMessage = "<|message|>"
         let gptStartControl = "<|start|>"
         let gptEndTags = ["<|end|>", "<|return|>"]
-        let openerTags = [gemmaStart, gptStart]
+        let gemmaStartTags = [gemmaStart] + startTagAliases
+        let openerTags = gemmaStartTags + [gptStart]
         let holdTags = openerTags + [gptMessage, gptStartControl, endTag] + gptEndTags
 
         var out: [ReasoningSegment] = []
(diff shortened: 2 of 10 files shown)


jjang-ai:
Follow-up source coverage at `2dbb39a1e56bf6c5f914e0611ab0002e253fdfae`:

- added `ToolCallProcessor` coverage with the `<tool_call>` wrapper, function name, parameter name, and malformed list split across separate stream deltas
- added a valid-JSON string-array control proving the normal JSON path is unchanged
- retained object-array and nested-syntax negative controls
- `MLXLMTests.ToolTests/testXMLFunctionParser*` plus the split-stream case: **13 tests passed**

Status remains **PARTIAL** until this exact commit is pinned into the isolated Release Osaurus app and the live Russian agent task passes with SSD cache both On and Off.

PR #184, opened 2026-07-25
Fix nested family tool arguments and LFM quantized loads

jjang-ai:
## Scope

This engine PR is intentionally limited to the model-family runtime fixes required by the current Osaurus compatibility campaign:

- replace comma-regex Pythonic argument extraction with balanced, quote-aware parsing so nested arrays/objects and quoted commas survive streaming boundaries;
- retain DSML inline-JSON fallback while validating against the actual tool schema;
- resynchronize ZAYA XML only when a new opener names another declared parameter, without treating arbitrary nested tags as parameters;
- carry LFM2 MoE quantized `scales` and `biases` through the same w1/w2/w3 module renames and expert stacking as weights.

No sampler coercion, forced thinking tags, cache-policy change, model artifact, screenshot, or binary evidence is included.

## Current source/test evidence

- Xcode Swift 6.3.3 focused parser run: **112 tests passed** across Pythonic nested arguments, DSML inline JSON, ZAYA XML, Qwen/Qwen 3.5, Gemma/Gemma 4, LFM, MiniMax, Mistral, Kimi, JSON, and XML controls.
- Metal-backed LFM sanitizer run: **2 tests passed**, covering dense sidecar renames and per-expert weight/scale/bias stacking.
- `git diff --check`: clean.

## Live-app gate

This PR is not being presented as live-verified by unit tests. The exact commit will be pinned into an isolated Release Osaurus build and exercised through the real UI for load, streamed reasoning/tool/final lifecycle, nested arguments, retry continuation, follow-up, token/s, and terminal unlock before merge.

### DIFF
--- a/Libraries/MLXLLM/Models/LFM2MoE.swift
+++ b/Libraries/MLXLLM/Models/LFM2MoE.swift
@@ -450,10 +450,15 @@ public class LFM2MoEModel: Module, LLMModel, KVCacheDimensionProvider {
                 }
             }
 
+            // Older LFM checkpoints use w1/w2/w3 module names. Rename the
+            // module path, not only `.weight`: quantized checkpoints carry
+            // sibling `.scales` and `.biases` tensors that must follow the
+            // same destination module or Module.update reports stale
+            // w1/w2/w3 children as unhandled.
             let replacements = [
-                "w1.weight": "gate_proj.weight",
-                "w2.weight": "down_proj.weight",
-                "w3.weight": "up_proj.weight",
+                ".w1.": ".gate_proj.",
+                ".w2.": ".down_proj.",
+                ".w3.": ".up_proj.",
             ]
             var updatedName = name
             for (old, new) in replacements where updatedName.contains(old) {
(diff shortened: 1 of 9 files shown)


jjang-ai:
Live Release-app gate completed against exact engine head `7c37de3e77daf1ce1e3a212c0aaad3ba4cb471b9` in isolated Osaurus bundle `com.dinoki.osaurus.runtimecompat7c37de3e20260724`.

- LFM2.5 MXFP8 loaded through the real model picker and executed create/insert/query database turns with raw native Pythonic envelope tracing.
- Nested values reached the tool body intact; malformed explicit/incomplete envelopes returned retryable `invalid_args` instead of executing partial arguments or disappearing.
- Post-error continuation restored disk-L2 state instead of resetting prefill to zero, and successful tool cards completed.
- ZAYA XML resynchronization remains source/focused-test evidence only; it is not being relabeled as live-proven.
- Exact Osaurus Release workspace builds completed with exit 0 at this pin; no forced thinking tags, sampler clamps, parser-synthesized values, MXFP4 claims, images, or binary artifacts were added.

Honest status: the LFM parser/execution/cache-continuation boundary is live-proven; the separate LFM post-error final-reasoning closure row remains PARTIAL and is documented in the Osaurus campaign matrix.

PR #191, opened 2026-07-27
Preserve Gemma 4 text-only system prompts

jjang-ai:
## Root cause

`Gemma4MessageGenerator` converted every non-empty message into a content-parts array. Real Gemma 4 templates read the leading system/developer content as a scalar before their generic multimodal branch, so prompt-affecting settings text was omitted. Distinct revisions therefore tokenized identically and an old SSD checkpoint could be restored as if it were current.

## Fix

Keep canonical scalar content for text-only turns. Build content-parts arrays only when the turn actually carries image, video, or audio media.

## Source test

- `ChatMessageToolCallTests.gemma4ProcessorPreservesTextOnlySystemContent`: 1/1 passed with the Xcode Swift toolchain.

## Real OsaurusEval proof

`cache_proof.settings-prompt-revision`, paged RAM off and disk L2 enabled, using `OsaurusAI--gemma-4-E2B-it-8bit`:

- revision A: no restore, 218 prompt tokens, 529 ms TTFT
- changed revision B: no restore, 213 prompt tokens, 215 ms TTFT
- repeated revision B: disk restored 212/213 tokens, 1 token prefilled, 203 ms TTFT
- disk L2 delta: +1 hit / +14 stores
- 43.8 decode tok/s; non-empty visible output; reasoning closed

The eval was not weakened. Before this fix, all revisions collapsed to 25 prompt tokens and changed revision B incorrectly restored 24/25 tokens from revision A.

No image or video artifacts are included.

### DIFF
--- a/Libraries/MLXVLM/Models/Gemma4.swift
+++ b/Libraries/MLXVLM/Models/Gemma4.swift
@@ -1944,6 +1944,18 @@ public struct Gemma4Processor: UserInputProcessor {
 private struct Gemma4MessageGenerator: MessageGenerator {
     func generate(message: Chat.Message) -> MLXLMCommon.Message {
         var dict = defaultMessageDict(for: message)
+        let hasMedia =
+            !message.images.isEmpty
+            || !message.videos.isEmpty
+            || !message.audios.isEmpty
+        // Keep ordinary text-only turns in the canonical scalar form. Gemma 4
+        // bundle templates read the leading system/developer content directly
+        // as a string before their generic per-turn content-parts branch. The
+        // previous unconditional array conversion silently dropped that system
+        // prompt in those real templates, making distinct settings revisions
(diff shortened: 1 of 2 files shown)


jjang-ai:
Current proof update (text-only):

- `ChatMessageToolCallTests`: 17/17 passed, including existing Qwen2VL/media content-array and Gemma tool-history rows.
- `cache_proof.settings-prompt-revision` on real Gemma 4 E2B 8-bit: PASS.
  - revision A: restore 0/218, TTFT 529 ms
  - changed revision B: restore 0/213, TTFT 215 ms
  - repeated B: disk restore 212/213, remaining prefill 1, TTFT 203 ms
  - disk L2 +1 hit / +14 stores; 43.8 decode tok/s; reasoning closed; visible output non-empty

This specifically keeps the pre-existing multimodal array path covered while correcting text-only system prompt shape.

PR #199, opened 2026-08-02
Capture safe DSV4 disk prefix seeds during prefill

jjang-ai:
## What changed

- Capture a typed DeepSeek V4 prompt-minus-one cache seed while prefill crosses that exact boundary, before the final prompt token is evaluated.
- Use the captured seed for solo and batched disk-prefix storage across DSV4 local SWA, compressor/CSA, and indexer/HSA state.
- Retain only the usable N-1 seed for DSV4 instead of also holding an exact-prompt snapshot that the coordinator intentionally cannot restore.
- On an N-1 warm hit, re-feed only the final prompt token without recapturing the seed or retaining another exact snapshot.
- Preserve terminal GPU synchronization, stream completion, and long-context memory purge when a warm hit has no new snapshot to store.

## Root cause

The DSV4 disk coordinator correctly rejects an exact full-prompt restore because generation still needs logits from the final prompt token. It therefore needs a prompt-minus-one seed. After the 128-token local SWA ring wraps, the completed DSV4 hybrid cache cannot be losslessly trimmed back by one token, so the N-1 entry was never published. Exact replay consequently fell back to an older stable 285-token boundary even though a 721-token exact entry existed.

This change captures the complete typed state at N-1 during normal prefill, then evaluates the last token normally. It does not relax token identity or positional validity and does not enable suffix matching.

## Validation

- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcrun swift test --skip-build --filter DeepseekV4IndexerCausalTopKTests` — 4/4 passed, including typed N-1 cold-vs-disk-restore equality.
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcrun swift test --filter BatchEngineGrowingChatCacheSourceTests` — 26/26 passed, including solo/batch capture selection, warm no-recapture, one-snapshot retention, and terminal cleanup reachability.
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcrun swift build --target MLXLMCommon` — passed.
- `git diff --check` — passed.

## Current proof status

PARTIAL. The source and focused regression gates pass. The exact Osaurus Release CacheProof replay, restart restore counters, coherent follow-up output, cache bytes, terminal UI state, and throughput still need to be rerun against this commit before the PR is merge-ready.

### DIFF
--- a/Libraries/MLXLLM/Models/DeepseekV4Compressor.swift
+++ b/Libraries/MLXLLM/Models/DeepseekV4Compressor.swift
@@ -515,7 +515,12 @@ public final class DeepseekV4Cache: QuantizedHybridPoolCache, CacheRetainedByteC
             poolQuantizationEnabled: hybridPoolQuantizationEnabled)
         let localState = local.state
         if !localState.isEmpty {
-            dup.local.state = localState.map { $0[.ellipsis] }
+            // RotatingKVCache writes decode tokens into its ring in place.
+            // An ellipsis slice is only a view, so the retained prompt-minus-one
+            // checkpoint would otherwise be overwritten before the synchronous
+            // disk store at terminal state. Create independent buffers; the
+            // snapshot caller batches their materialization in one MLX.eval.
+            dup.local.state = localState.map { $0 * 1 }
         }
         dup.local.metaState = local.metaState
         dup.compPool = compPool.copy()
--- a/Libraries/MLXLMCommon/BatchEngine/BatchEngine.swift
+++ b/Libraries/MLXLMCommon/BatchEngine/BatchEngine.swift
@@ -2060,7 +2060,10 @@ public actor BatchEngine {
     /// the cache-copy high-water mark. A restored N-1 request does not retain
     /// that duplicate and is consequently much faster. The seed is already a
     /// complete, immutable prompt boundary here, so write it synchronously and
-    /// release it before the final prompt token and sampled decode run.
+    /// release it before the final prompt token and sampled decode run. This is
+    /// also the boundary a reusable-prefix warmup must publish: DSV4 deliberately
+    /// rejects an exact post-prefill restore, so excluding warmups here causes the
+    /// immediately following visible request to prefill the same prefix again.
     private func storePrefillCapturedDiskSeed(
         _ snapshot: [KVCache],
         for slot: BatchSlot
@@ -2126,7 +2129,6 @@ public actor BatchEngine {
                     cacheCoordinator?.canPersistBoundaries == true
                     && slot.diskSeedSnapshot == nil
                     && cacheRequiresPrefillCapturedDiskSeed(slot.cache)
-                    && slot.originalInput.cachePromptIntent != .reusablePrefixWarmup
                     && !slot.originalInput.hasMediaContent
                     && !slot.originalInput.requiresPostPrepareCacheKey
                     && !shouldSkipDiskBackedToolPromptSeedBoundary(for: slot)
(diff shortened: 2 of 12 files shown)


jjang-ai:
## Live Release proof — duplicate prefill FIXED

**Tested source:** vMLX `6ec199b6` (fix commit `9d22ac14`), Osaurus `84fa4479`
**App:** fresh Release build, derived `/private/tmp/osaurus-dsv4-prefill-seed-42ec8952-release-derived-20260802`, binary sha256 `602d8d1c2a9678a66e24960526a127d09c806f78b69d270fba76338225294470`, isolated bundle id `com.dinoki.osaurus.dsv4prefillseed20260802`
**Bundle:** `~/models/DeepSeek-V4-Flash-0731-JANG` (deepseek_v4, 43L, 256 experts, affine JANG; `use_mtp:false`)
**Root:** isolated, paged RAM cache OFF, disk L2/prefix ON
**Env:** `VMLX_CACHE_FETCH_TRACE=1 OSAURUS_PREFILL_DEBUG=1 VMLX_TOOL_CALL_TRACE=1`

### Before → after

The reusable-prefix warm-up was excluded from the DSV4 prefill-captured N-1
disk store, so the warm-up prefilled the prefix and the visible send then
prefilled the same tokens again. Live trace on this pin:

```
[vmlx][cache/fetch] MISS all tiers tokens=1129                        <- warm-up (cold, stores seed)
[vmlx][cache/fetch] HIT disk boundary=1127 remaining=96 tokens=1223   <- visible send restores it
```

The visible send restores 1127 of 1223 tokens and prefills only the 96-token
user suffix. Previously this line was `MISS all tiers`.

### Per-turn UI lifecycle (visual, live app)

- reasoning opened and closed in the reasoning channel — "Thought for 3.6s", 377 chars; no `<think>` leaked inline
- final answer coherent, no protocol debris, no repetition, no truncation
- `TTFT 1.73s · 31.3 tok/s · 144 tokens`
- Stop control disappeared, input unlocked
- follow-up turn dispatched and ran

### Correction to the earlier diagnosis

Two `[vmlx][cache/boundaries]` lines per warm-up are **not** a second prefill.
They are the two probe renders in `MLXBatchAdapter.warmupSendInvariantPrefixTokens`,
which tokenize twice to find the send-invariant prefix. Only `cache/fetch` marks
real prefill work, and the model loads once (each `[Load]` line appears exactly
once; `footprint` 100 GB, single mapping).

### Also in this push

- `docs/DSV4_TOOL_TURN_PREFILL_2026_08_02.md` — full live findings
- `DeepseekV4ToolHistoryPrefixBoundaryTests` — 3/3 pass; pins that rendering with
  `addGenerationPrompt: false` stays an exact textual prefix of the real prompt for
  a plain history, a tool-call history with a large string payload, and a completed
  tool round followed by a user turn

`swift test --filter DeepseekV4ToolHistoryPrefixBoundaryTests` → 3/3 passed.

### Open, NOT fixed by this PR — reported as PARTIAL

Once a completed tool call and its artifact enter the transcript,
`canonicalChatCacheBoundaries` stops publishing a history boundary:

| prompt | `all=` | fetch |
|---|---|---|
| 1223 | `[72, 1128, 1221]` | HIT 1127, remaining 96 |
| 1458 | `[72, 1128, 1456]` | HIT 1367, remaining 91 |
| **7017** | **`[72, 1128]`** | HIT 1457, **remaining 5560** |

That turn cold-prefills 5560 tokens and, because stores happen at published
boundaries, no history entry can be stored for the next agent iteration. In the
live run that prefill had not completed after ~20 minutes at ~180% CPU. Ruled
out: permissions (empty `policy`/`grants`, 78 tools enabled, no dialog pending),
the DSML parser (first and second tool calls both round-trip over
`/v1/chat/completions`, streaming and non-streaming), and the generation-prompt
rail (tests above). Tracked in the doc; not claimed as fixed here.

PR #200, opened 2026-08-02
Keep the history cache boundary for an assistant continuation turn

jjang-ai:
## What changed

`canonicalChatCacheBoundaries` gains `trailingContinuationBoundary()`: when the
whole-message-list render yields no history boundary and the last message is an
assistant turn, retry against the transcript **without** that trailing turn.

## Root cause

DSV4 agent-loop turns lost prefix-cache reuse entirely. Live Release trace
(vMLX `9d22ac14`, `~/models/DeepSeek-V4-Flash-0731-JANG`, isolated root, paged
RAM off / disk L2 on, `VMLX_CACHE_FETCH_TRACE=1`):

| prompt | `all=` boundaries | fetch |
|---|---|---|
| 1223 | `[72, 1128, 1221]` | HIT 1127, remaining 96 |
| 1458 | `[72, 1128, 1456]` | HIT 1367, remaining 91 |
| **7017** | **`[72, 1128]`** | HIT 1457, **remaining 5560** |

Row 3 is the first prompt containing a completed tool call and its 7 KB
artifact. The history boundary disappears, so that turn cold-prefills 5560
tokens — and because entries are stored **at** published boundaries, the next
agent iteration has nothing to reuse either. Every round re-prefills the whole
growing transcript. In the live run that prefill had not finished after ~20
minutes at ~180% CPU, with the chat showing "Preparing tool call" and Stop
still active.

`exactPrefixBoundary` requires the no-generation-prompt render to be **strictly
shorter** than the prompt:

```swift
tokens.count < promptTokens.count,
promptTokens.prefix(tokens.count).elementsEqual(tokens)
```

DSV4 appends its `<｜Assistant｜>` rail only after a user / developer /
latest_reminder message, mirroring `encoding_dsv4.py`:

```python
elif messages[index].get("role") in ["user", "developer"]:
    prompt += ASSISTANT_SP_TOKEN + <think>/</think>
```

So a transcript ending in an assistant **continuation** — the shape every
agent-loop turn takes mid-turn — renders identically with and without the
generation prompt. Measured on the real encoder: `withRail.count → 1283` ==
`withoutRail.count → 1283`. The strict `<` rejects it, and `try?` / `?? []`
turn that into a silent nil with no diagnostic.

The encoder is correct per the official contract; the boundary derivation was
not.

## Why this is safe

The fallback only engages when the normal path returns nil **and** the last
message is an assistant turn. The retried render is strictly shorter and is
still proven by the same exact-token-prefix check. Token identity is unchanged
and suffix matching is not admitted.

## Validation

- `CanonicalChatCacheBoundariesTests."a trailing assistant continuation still publishes a history boundary"` — sabotage-verified: without the fallback the published history set is `[]`.
- `DeepseekV4ToolHistoryPrefixBoundaryTests` (4/4) — pins the encoder contract the fallback exists for, plus exact-prefix round-trips for a plain history, a tool-call history with a large string payload, and a completed tool round followed by a user turn.
- `swift test --filter 'CanonicalChatCacheBoundariesTests|DeepseekV4ToolHistoryPrefixBoundaryTests'` → 12/12 passed.
- Full `MLXLMCommonFocusedTests`: 22 failures in `DeepseekV4ChatTemplateFallbackFocusedTests` and `VMLXMemorySafetySettingsTests` **both with and without** this change — pre-existing, none added.

## Proof status

**PARTIAL.** Source and focused regression gates pass and the root cause is
proven from a live Release trace. The live UI re-run on a build carrying THIS
commit — agent turn with a tool call, then a follow-up, showing the history
boundary published and the next iteration restoring instead of cold-prefilling —
has not been done yet. Details in `docs/DSV4_TOOL_TURN_PREFILL_2026_08_02.md`.

### DIFF
--- a/Libraries/MLXLMCommon/Tokenizer.swift
+++ b/Libraries/MLXLMCommon/Tokenizer.swift
@@ -218,7 +218,35 @@ public func canonicalChatCacheBoundaries(
                 .compactMap { $0 }
         )
     ).sorted()
-    let history = exactPrefixBoundary(messages: messages).map { [$0] } ?? []
+    /// A transcript ending in an assistant turn is a CONTINUATION, and formats
+    /// that model that faithfully append no generation rail for it — DSV4's
+    /// official encoder only emits `<｜Assistant｜>` after a user/developer/
+    /// latest_reminder message. Rendering with and without the generation
+    /// prompt then produces identical tokens, so `exactPrefixBoundary`'s strict
+    /// `tokens.count < promptTokens.count` rejects it and the history boundary
+    /// disappears entirely.
+    ///
+    /// That silently killed prefix reuse for exactly the agent-loop turns that
+    /// need it most. Live DSV4 row: a plain turn published
+    /// `all=[72, 1128, 1456]`, but the first turn carrying a completed tool call
+    /// and its artifact published only `all=[72, 1128]` and cold-prefilled 5560
+    /// tokens — and because entries are stored AT published boundaries, the next
+    /// iteration had nothing to reuse either, so every round re-prefilled the
+    /// whole growing transcript.
+    ///
+    /// Fall back to the transcript without that trailing assistant turn. It is
+    /// strictly shorter, and it is still proven by the same exact-token-prefix
+    /// check — this does not relax token identity or admit suffix matching.
+    func trailingContinuationBoundary() -> Int? {
+        guard messages.count > 1,
+              let last = messages.last,
+              (last["role"] as? String) == "assistant"
+        else { return nil }
+        return exactPrefixBoundary(messages: Array(messages.dropLast()))
+    }
+
+    let history = (exactPrefixBoundary(messages: messages)
+        ?? trailingContinuationBoundary()).map { [$0] } ?? []
     let all = Array(Set(stable + history)).sorted()
     if ProcessInfo.processInfo.environment["VMLX_CACHE_FETCH_TRACE"] == "1" {
         FileHandle.standardError.write(Data(
--- a/Tests/MLXLMCommonFocusedTests/CanonicalChatCacheBoundariesTests.swift
+++ b/Tests/MLXLMCommonFocusedTests/CanonicalChatCacheBoundariesTests.swift
@@ -194,6 +194,99 @@ struct CanonicalChatCacheBoundariesTests {
         }
     }
 
+    /// DSV4-shaped: the generation rail is appended only after a
+    /// user/developer turn, so a transcript ending in an assistant turn (an
+    /// agent-loop continuation) renders identically with and without the
+    /// generation prompt.
+    private struct ContinuationRailTokenizer: GenerationPromptControllableTokenizer {
+        var bosToken: String? { nil }
+        var eosToken: String? { nil }
+        var unknownToken: String? { nil }
+
+        func encode(text: String, addSpecialTokens: Bool) -> [Int] { [] }
+        func decode(tokenIds: [Int], skipSpecialTokens: Bool) -> String { "" }
+        func convertTokenToId(_ token: String) -> Int? { nil }
+        func convertIdToToken(_ id: Int) -> String? { nil }
+
+        func applyChatTemplate(
+            messages: [[String: any Sendable]],
+            tools: [[String: any Sendable]]?,
+            additionalContext: [String: any Sendable]?
+        ) throws -> [Int] {
+            try applyChatTemplate(
+                messages: messages,
+                tools: tools,
+                additionalContext: additionalContext,
+                addGenerationPrompt: true)
+        }
+
+        func applyChatTemplate(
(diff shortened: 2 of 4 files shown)


jjang-ai:
## LIVE VISUAL PROOF — dev-built Osaurus Release app

**Tested source:** vMLX `6c936110` (this PR head), Osaurus `298c4934`
**App:** Release, derived `/private/tmp/osaurus-dsv4-prefill-seed-42ec8952-release-derived-20260802`, binary sha256 `c1424a22d9be244f902b19de396200446cd75179e1ca8b98b44980caa2835b66`, isolated bundle id `com.dinoki.osaurus.dsv4prefillseed20260802`
**Embedded checkout verified:** `git -C SourcePackages/checkouts/vmlx-swift rev-parse HEAD` = `6c93611062bd3e42cc7d4b31323467677e44a340`
**Bundle:** `~/models/DeepSeek-V4-Flash-0731-JANG`
**Root:** FRESH isolated root, empty cache; paged RAM cache OFF, disk L2/prefix ON; Sandbox + folder tools enabled
**Env:** `VMLX_CACHE_FETCH_TRACE=1 VMLX_TOOL_CALL_TRACE=1 OSAURUS_PREFILL_DEBUG=1`

All rows driven through the real chat UI over Accessibility and observed with
window captures. No CLI or API request was used as proof.

### The defect, before vs after

Same tool-turn shape, old build (`9d22ac14`) vs this PR:

| | old `9d22ac14` | **this PR `6c936110`** |
|---|---|---|
| tool turn boundaries | `all=[72, 1128]` — history boundary GONE | `all=[208, 1830, 2391]` |
| tool turn fetch | `HIT 1457 remaining=5560` | **`HIT 2286 remaining=107`** |
| next turn | no store; cold re-prefill of the whole transcript | **`HIT 2392 remaining=346`**, stores advance 2392 → 2737 → 2756 |
| terminal state | never finalized; 20+ min at ~180% CPU | **finalized in 41s** |

### Per-turn UI lifecycle (visual)

**t3 — tool call.** "Thought for 4.5s" (reasoning opened AND closed, 125 chars,
nothing leaked inline) → tool card **"Wrote a file · 2.7s"** started and
finished → `dsv4proof.txt +1` diff card → final answer "Created dsv4proof.txt
containing exactly the line HELLO." → Stop disappeared, input unlocked.
`TTFT 17.39s · 5.3 tok/s · 18 tokens`, finalized wall 41s.

**t4 — follow-up after the tool turn.** Coherent ~300-word answer, no protocol
debris, no repetition, no truncation. Stop disappeared, input unlocked.
`TTFT 1.91s · 23.3 tok/s · 681 tokens`, finalized wall 45s.
Boundaries `all=[208, 1830, 2864]`, fetch `HIT 2756 remaining=110`, stores
advance 2865 → 3547.

### Throughput note

The 4.7–5.3 tok/s rows are short-generation artifacts (18 and 108 tokens),
where fixed per-turn cost dominates the honest denominator. A real 681-token
generation measures **23.3 tok/s**, consistent with the 21.0 tok/s native rate
recorded in `ad059136`. TTFT fell 28.86s → **1.91s** once the prefix cache
started hitting. Compiled decode was NOT changed and must not be: `ad059136`
records an exact-bundle A/B of 21.0 tok/s native vs 12.7 tok/s with the generic
compiled request, because DSV4's composite SWA + CSA/HSA cache is not a
promotable whole-forward cache.

### Full fetch timeline (fresh root)

```
1  MISS tokens=1831      <- cold warm-up on an empty cache
2  MISS tokens=1841
3  MISS tokens=1923
4  MISS tokens=2086
5  HIT  boundary=2084 remaining=94   tokens=2178
6  MISS tokens=2288
7  HIT  boundary=2286 remaining=107  tokens=2393   <- TOOL TURN
8  HIT  boundary=2392 remaining=346  tokens=2738
9  MISS tokens=2758
10 HIT  boundary=2756 remaining=110  tokens=2866   <- follow-up after tool
```

### Honest gaps

- Rows 6 and 9 are still MISSes. These are post-turn background re-warms asking
  for a boundary no turn has stored yet. They do not block the visible turn
  (rows 7 and 10 both hit) but they are real extra work and are not fixed here.
- The continuation turn at `prompt=2738` still published `all=[208, 1830]` with
  no history boundary, yet the fetch restored 2392 anyway. So reuse is working
  through the coordinator's longest-valid-prefix search rather than through a
  published continuation boundary on that particular turn. The
  `trailingContinuationBoundary` path is proven by unit test and by the tool-turn
  rows above; that specific turn shape deserves a follow-up look.
- Separate open item, not fixed here: `BatchEngine.swift` runs the end-of-turn
  cache store and `Stream().synchronize()` before `continuation.finish()`, so a
  25–61 MB store sits between the last token and finalization.

Details in `docs/DSV4_TOOL_TURN_PREFILL_2026_08_02.md`.

PR #201, opened 2026-08-02
DSV4 encoder × cache hardening: deterministic tool re-render, plus the audit

jjang-ai:
Follow-up to #199 (duplicate prefill) and #200 (continuation boundary). This is
the encoder-side audit those two surfaced, plus the one defect in it that is a
real cache-correctness bug.

## 1. Non-deterministic tool-call re-render — FIXED

`renderToolCallInvoke(name:params:)` iterated an unordered Swift `Dictionary`,
so chat history containing a tool call re-rendered its parameters in a
different order in **every process**. Reached from
`renderToolCallInvoke(name:arguments:)` whenever the arguments string does not
scan as an ordered JSON object.

Prefix reuse requires reproducible bytes, so this silently broke the boundary
for every turn after a tool call across app restarts. Same class as the
per-process-random ordering removed from the load path in #108.

The official encoder walks `json.loads(...).items()` and is stable, and the
ordered overload preserves the original JSON spelling — only this fallback
diverged. Now iterates `params.keys.sorted()`. Sorted order cannot recover the
model's original emission order, but it makes the bytes reproducible, which is
what reuse actually requires.

Test renders the same 9-parameter call 12 times and requires exactly one
distinct result, then asserts the documented order rather than mere
self-consistency.

## 2. `drop_thinking` vs prefix reuse — PINNED (behaviour is correct)

This is why multi-turn reuse looked broken. `dropThinkingMessages` strips
`reasoning_content` from every assistant message BEFORE the last user message,
so each new user turn pushes the previous assistant turn across that line and
removes reasoning that was already cached. History changes **retroactively** and
only the most recent store can ever hit.

Live evidence (isolated Release root, `VMLX_CACHE_FETCH_TRACE=1`):

```
fetch tokens=2178 -> HIT boundary=2084   (stored by the immediately prior turn)
fetch tokens=2086 -> MISS, although 1829 and 207 were both stored
```

with TTFT climbing 17.32s → 28.86s across consecutive turns.

Two tests pin both halves: without tools the histories provably diverge once
reasoning is dropped; with tools — where the official contract disables
`drop_thinking`, which is exactly what makes agent-loop reuse possible — turn
N's railless render stays an exact byte prefix of turn N+1.

Not "fixed", because the official contract defines it. Pinned so a future cache
investigation does not re-derive it or misread it as a regression.

## 3. DSML is XML-shaped with no escaping — CHARACTERIZED

The 0731 format defines no escaping on either side: the official encoder writes
string values raw and `parseCanonicalInvokes` reads up to the first
`</｜DSML｜parameter>`. A `string="true"` value containing that literal is
unrepresentable and silently truncates, and the tail does not become a second
call.

Reachable in practice — DSV4 writes whole source files through `file_write`,
and a live run put 7 KB of HTML through exactly this path. Pinned as a
characterization test so the hazard is visible and a future escaping or
detection change has somewhere to land.

## 4. Documentation

`docs/DSV4_ENCODER_CACHE_AUDIT_2026_08_02.md` records the full audit against
the bundle's official `encoding/encoding_dsv4.py`, including the items that are
correct and must NOT be "fixed":

- reasoning enforcement is clean — no forced thinking tags, no injected
  `<think>`, no decode-loop close-token bias; an explicit `enable_thinking:
  false` wins and an unsupported effort throws instead of silently selecting
  another rail. Its cache consequence: the effort preface is emitted at index 0,
  so changing effort rewrites the first bytes and invalidates the **stable**
  system boundary too — the effort chip is a full-invalidation control.
- the `latest_reminder` generation-rail divergence from official Python, which
  `requiredToolChoiceReminder` depends on and which was undocumented.
- unsupported content blocks are dropped rather than rendered as
  `[Unsupported X]` placeholders.
- why `DSMLToolCallParser` deliberately has no `isValidPartialContent` override:
  a canonical envelope with a garbage body must keep buffering so it is reported
  as `.malformedEnvelope` instead of flushed to the user. An attempt at adding a
  strict grammar check during this audit turned 4 existing quarantine tests red
  and was reverted.

## Validation

- `swift test --filter 'DeepseekV4ToolHistoryPrefixBoundaryTests|DeepseekV4DropThinkingCacheTests'` → all pass.
- Full `MLXLMCommonFocusedTests`: the same 22 pre-existing failures in
  `DeepseekV4ChatTemplateFallbackFocusedTests` and
  `VMLXMemorySafetySettingsTests` with and without this change — none added.

## Proof status

**PARTIAL.** Items 2–4 are documentation and characterization of behaviour
already observed live on `6c936110` (see the proof comment on #200: tool card
start→finish, `HIT 2286 remaining=107` on the tool turn, `TTFT 1.91s · 23.3
tok/s · 681 tokens` on the follow-up). Item 1 changes prompt bytes on a fallback
path and has NOT yet been live-exercised in the app — it needs a row where a
tool call's arguments do not scan as an ordered JSON object, then an app restart
proving the boundary still matches. Not claiming that row until it is run.

### DIFF
--- a/Libraries/MLXLMCommon/DeepseekV4ChatEncoder.swift
+++ b/Libraries/MLXLMCommon/DeepseekV4ChatEncoder.swift
@@ -881,7 +881,22 @@ public struct DeepseekV4ChatEncoder: Sendable {
         // Encode each param — string params carry `string="true"` and
         // a raw value; all other JSON types serialize as JSON with
         // `string="false"`.
-        for (k, v) in params {
+        //
+        // Iterate in SORTED key order. `params` is an unordered Dictionary, so
+        // `for (k, v) in params` emitted a different parameter order in every
+        // process. The official encoder walks `json.loads(...).items()` and is
+        // therefore stable, and the ordered overload above preserves the
+        // original JSON spelling; this fallback path (reached from
+        // `renderToolCallInvoke(name:arguments:)` when the arguments string
+        // does not scan as an ordered JSON object) was the one place that did
+        // not. Re-rendering a tool call in chat history with a random parameter
+        // order changes the prompt bytes run to run, so the prefix-cache
+        // boundary for every turn after a tool call stops matching across app
+        // restarts. Deterministic order cannot recover the model's original
+        // emission order, but it makes the bytes reproducible, which is what
+        // cache reuse actually requires.
+        for k in params.keys.sorted() {
+            guard let v = params[k] else { continue }
             let isString = v is String
             let value: String
             if isString, let s = v as? String {
--- a/Libraries/MLXLMCommon/Tokenizer.swift
+++ b/Libraries/MLXLMCommon/Tokenizer.swift
@@ -237,11 +237,16 @@ public func canonicalChatCacheBoundaries(
     /// Fall back to the transcript without that trailing assistant turn. It is
     /// strictly shorter, and it is still proven by the same exact-token-prefix
     /// check — this does not relax token identity or admit suffix matching.
+    /// A trailing `assistant` turn is a continuation; a trailing `tool` turn is
+    /// the tool result the model is about to continue from. Both are shapes a
+    /// format may render without a generation rail, and both appear in a normal
+    /// agent loop — the post-turn re-warm after a TOOL call ends with the tool
+    /// result, which showed up live as a fixed miss pattern (stored N, the
+    /// re-warm asks N+2, MISS) on 2084→2086, 2286→2288 and 2756→2758.
     func trailingContinuationBoundary() -> Int? {
-        guard messages.count > 1,
-              let last = messages.last,
-              (last["role"] as? String) == "assistant"
-        else { return nil }
+        guard messages.count > 1, let last = messages.last else { return nil }
+        let role = last["role"] as? String
+        guard role == "assistant" || role == "tool" else { return nil }
         return exactPrefixBoundary(messages: Array(messages.dropLast()))
     }
 
--- a/Package.swift
+++ b/Package.swift
@@ -827,6 +827,7 @@ let package = Package(
                 "DSMLInlineJSONToolFallbackFocusedTests.swift",
                 "DSMLToolCallParserFocusedTests.swift",
                 "DeepseekV4ToolHistoryPrefixBoundaryTests.swift",
+                "DeepseekV4DropThinkingCacheTests.swift",
                 "ToolCallProgressRoutingTests.swift",
                 "FocusedMLXTestSupport.swift",
                 "CanonicalChatCacheBoundariesTests.swift",
--- a/Tests/MLXLMCommonFocusedTests/CanonicalChatCacheBoundariesTests.swift
+++ b/Tests/MLXLMCommonFocusedTests/CanonicalChatCacheBoundariesTests.swift
@@ -241,7 +241,8 @@ struct CanonicalChatCacheBoundariesTests {
                 }
             }
             // The rail exists only when an assistant turn is being OPENED.
-            if addGenerationPrompt, (messages.last?["role"] as? String) != "assistant" {
+            let lastRole = messages.last?["role"] as? String
+            if addGenerationPrompt, lastRole != "assistant", lastRole != "tool" {
                 result.append(99)
             }
             return result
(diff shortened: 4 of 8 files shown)


jjang-ai:
## Added: trailing tool result in the continuation fallback (`86508aeb`)

The fallback merged in #200 only fired when the last message was an **assistant**
turn. The post-turn re-warm after a **tool** call ends with the tool RESULT, so
it was still dropped. That is the residual miss pattern I flagged as an honest
gap in #200's proof comment, and it has an exact signature — the re-warm asks
for precisely the stored boundary plus the two-token generation rail:

```
store 2084 -> fetch 2086 MISS
store 2286 -> fetch 2288 MISS
store 2756 -> fetch 2758 MISS
```

while the visible turn that followed restored the same entry without trouble
(`HIT 2286 remaining=107`, `HIT 2756 remaining=110`). So the misses were confined
to the background re-warm rather than the user-visible turn — but each one is a
real extra prefill.

A trailing tool turn is the same shape as a trailing assistant turn for boundary
purposes: the model is about to continue from it, a format may render it without
a generation rail, and dropping it leaves a strictly shorter prefix that the same
exact-token-prefix check still has to prove. No relaxation of token identity, no
suffix matching.

Sabotage-verified: with the assistant-only guard the published history set is `[]`.

## Proof status — still PARTIAL, do not merge yet

This commit and the deterministic tool re-render both change bytes/behaviour on
paths that have NOT been exercised in the live app on a build containing them.
The live evidence quoted above was captured on `6c936110`, which does **not**
include either change.

Required before merge:
1. repin Osaurus to this branch head, rebuild the Release app;
2. live UI run with a tool call, then confirm the post-turn re-warm HITs instead
   of the `stored+2` MISS;
3. an app restart proving a tool-call history still matches its boundary, which
   is what the deterministic re-render fix is actually for.

jjang-ai:
## LIVE VISUAL PROOF — improvement measured, residual issue disclosed

**Tested source:** vMLX `db9250c2` (this branch head), Osaurus `a7af5f40`
**App:** Release, binary sha256 `fa95ab5504db2e5f7812676b34b23d27546d3231aeee3d7a3112121d8287046b`, isolated bundle id `com.dinoki.osaurus.dsv4prefillseed20260802`
**Embedded checkout verified:** `SourcePackages/checkouts/vmlx-swift` HEAD = `db9250c2e711e19294eabf3740365b0ca8fac30d`, and `Tokenizer.swift` contains the `role == "tool"` guard
**Bundle:** `~/models/DeepSeek-V4-Flash-0731-JANG`; fresh isolated root, empty cache; paged RAM cache OFF, disk L2/prefix ON; Sandbox enabled
**All rows driven through the real chat UI over Accessibility.** No CLI or API request used as proof.

### Tool turn + follow-up

| turn | boundaries | fetch | wall | stats |
|---|---|---|---|---|
| tool call | `all=[72, 1128, 1416]` | `HIT 1231 remaining=435` | 77s (incl. model load) | `TTFT 3.34s · 28.7 tok/s · 85 tok` |
| follow-up | `all=[72, 1128, 1851]` | `HIT 1751 remaining=102` | **18s** | `TTFT 1.62s · 25.3 tok/s · 306 tok` |

Against the previous pin `6c936110` on the same shapes, follow-up wall dropped
**45s → 18s**.

### Restart persistence (disk L2 across a process boundary)

Quit through the app's own menu, relaunched on the SAME root, then a new turn:

```
[vmlx][cache/fetch] HIT disk boundary=1127 remaining=15  tokens=1142   <- first fetch in the NEW process
[vmlx][cache/fetch] HIT disk boundary=1127 remaining=97  tokens=1224
```

`TTFT 1.23s · 21.3 tok/s · 52 tokens`, "Thought for 2.1s" (reasoning opened and
closed), coherent answer, Stop cleared, input unlocked.

### Miss rate

Excluding unavoidable cold-start misses:

| pin | warm fetches | misses |
|---|---|---|
| `6c936110` | 7 | 3 — **43%** |
| **`db9250c2`** | 5 | 1 — **20%** |

### Residual issue — NOT fully fixed, disclosed

The `stored+2` re-warm miss is **much rarer but not eliminated**. One still
occurred in this session:

```
[vmlx][cache/disk-store] count=1276
[vmlx][cache/boundaries] prompt=1281 stable=[72, 1128] all=[72, 1128, 1279]
[vmlx][cache/fetch] MISS all tiers tokens=1278
```

The store landed at 1276 while the next request's usable boundary is 1278 — the
same two-token gap. So broadening the fallback to cover a trailing tool result
lets far more boundaries be published (and measurably improves reuse and wall
time), but the underlying off-by-two between where a store lands and where the
following re-warm looks is still there. That is a separate defect and is not
claimed fixed by this PR.

### Scope note on the deterministic re-render

The restart row above proves disk L2 survives the process boundary, but it
restarted into a NEW chat and therefore hit the stable system prefix (1127), not
a tool-call history boundary. So the determinism fix is proven by unit test
(12 renders → 1 distinct result) and is correct by inspection against the
official `json.loads(...).items()` ordering, but the specific live row — a
tool-call history whose arguments do NOT scan as an ordered JSON object,
surviving a restart — was not exercised. Stating that rather than implying it.

### Regression check

- `swift test --filter 'CanonicalChatCacheBoundariesTests|DeepseekV4ToolHistoryPrefixBoundaryTests|DeepseekV4DropThinkingCacheTests'` → 17/17 passed.
- Full `MLXLMCommonFocusedTests`: same 22 pre-existing failures in
  `DeepseekV4ChatTemplateFallbackFocusedTests` and `VMLXMemorySafetySettingsTests`
  with and without this change — none added.
- Osaurus pin tripwires at this SHA: `ImageGenerationBridgeContractTests` +
  `RuntimePolicySourceTests` 109/109.

PR #204, opened 2026-08-02
Add a prefix digest to the cache-boundary trace

jjang-ai:
Diagnostic only. Unblocks the `stored+2` re-warm miss whose cause #203 recorded
as unresolved.

## Why

The boundary trace prints token counts and nothing else, so a boundary that is
stored and then immediately missed looks identical in the log to one that
genuinely diverged. Two prompts of equal length cannot be told apart, which is
exactly the ambiguity that made me record a wrong root cause in #202 and correct
it in #203.

## What

`[vmlx][cache/boundaries]` now also emits:

- `head=` — digest of the leading 256 tokens
- `stable0=` — digest of the prompt up to the first stable boundary

Comparing these across a store and the following re-warm shows directly whether
the prompt changed in the system region, which is the leading hypothesis: the
todo tool was observed moving `stable[1]` from 1160 to 1128 between consecutive
boundary computations inside one session, while the static boundary 72 held.

Gated behind the existing `VMLX_CACHE_FETCH_TRACE` flag, never consulted by
cache logic, never used as a cache key. FNV-1a truncated to 32 bits purely for
log legibility.

## Validation

`swift build --target MLXLMCommon` → clean.

## Proof status

Diagnostic instrumentation behind an off-by-default trace flag, so there is no
runtime behaviour to prove live. The live run that *uses* it — capturing
`head=`/`stable0=` across a store and its following re-warm to settle the cause —
is the next step and will be reported separately.

### DIFF
--- a/Libraries/MLXLMCommon/Tokenizer.swift
+++ b/Libraries/MLXLMCommon/Tokenizer.swift
@@ -254,8 +254,27 @@ public func canonicalChatCacheBoundaries(
         ?? trailingContinuationBoundary()).map { [$0] } ?? []
     let all = Array(Set(stable + history)).sorted()
     if ProcessInfo.processInfo.environment["VMLX_CACHE_FETCH_TRACE"] == "1" {
+        // Token counts alone cannot distinguish two prompts of equal length, so
+        // a boundary that is stored and then immediately missed looks identical
+        // in the trace to one that genuinely diverged. Emit a digest of the
+        // leading tokens as well: comparing `head=` across a store and the
+        // following re-warm shows directly whether the prompt changed in the
+        // system region rather than leaving it to inference.
+        //
+        // Diagnostic only — gated behind the same trace flag, never consulted
+        // by cache logic, and never used as a cache key.
+        func digest(_ ids: ArraySlice<Int>) -> String {
+            var hash: UInt64 = 0xcbf2_9ce4_8422_2325
+            for id in ids {
+                hash = (hash ^ UInt64(bitPattern: Int64(id))) &* 0x1000_0000_01b3
+            }
+            return String(hash & 0xffff_ffff, radix: 16)
+        }
+        let headDigest = digest(promptTokens.prefix(256))
+        let stableDigest = stable.first.map { digest(promptTokens.prefix($0)) } ?? "-"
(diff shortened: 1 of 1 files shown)


jjang-ai:
## First live capture with the digest — falsifies the leading hypothesis

App built from this branch (`087e43ef`), embedded checkout verified. Same
isolated root as the prior proof session, driven through the real chat UI.

```
[vmlx][cache/boundaries] prompt=1141 all=[72,1128,1139] head=22bca74a stable0=e7fcebd8
[vmlx][cache/fetch] HIT disk boundary=1127 remaining=12  tokens=1139
[vmlx][cache/boundaries] prompt=1221 all=[72,1128,1219] head=22bca74a stable0=e7fcebd8
[vmlx][cache/fetch] HIT disk boundary=1127 remaining=94  tokens=1221
[vmlx][cache/disk-store] count=1220
[vmlx][cache/disk-store] count=71
[vmlx][cache/disk-store] count=1333
[vmlx][cache/boundaries] prompt=1338 all=[72,1128,1336] head=22bca74a stable0=e7fcebd8
[vmlx][cache/fetch] MISS all tiers tokens=1335
```

`head=22bca74a stable0=e7fcebd8` are **identical on every line**, including the
re-warm that missed. So the system region did **not** change between the store
and the failing fetch, and the "mutable system section" hypothesis recorded in
#203 is falsified along with the gen-suffix hypothesis from #202.

Two theories down. What the data now says:

- the system prefix is byte-stable across the store and the miss;
- the store is `1220 + 113 generated = 1333`, the re-warm's usable boundary is
  `1338 - 3 = 1335`, so the stored entry is only 2 tokens short;
- the divergence must therefore be in the **conversation-history region**,
  between the stable prefix and token 1333 — which is exactly the span these two
  digests do not cover.

`head=` spans the first 256 tokens and `stable0=` the first 72, so both sit
entirely inside the unchanging system prefix. Localizing this needs a third
digest taken at the **history boundary** (`all.last`), which is the natural
follow-up to this PR rather than something to bolt on before it lands.

Recording the falsifications rather than quietly re-theorizing: this is the
second wrong root cause for the same symptom, and the pattern is that token
counts and coarse prefixes both look identical while something inside the
history span does not.

PR #208, opened 2026-08-02
Refuse disk-cache stores whose cache offset disagrees with the boundary key

jjang-ai:
## The DSV4 agent-loop degeneration, root-caused

Long DSV4 tool chains degenerated into fluent, non-converging reasoning loops
at ~12.5k tokens. Live A/B on the affected machine, run twice: prefix-RAM-only
never loops; disk L2 on loops; everything off never loops.

## The smoking gun was already in the trace

```
[vmlx][cache/paged-store] tokens=1831 requiredCompanion=true
    effectiveKVLayers=0 blocks=0 payload=false companion=false rotatingOffsets=[1832]
```

A snapshot keyed at **1831 tokens** carrying a rotating cache at **offset
1832**. The paged tier refused it — `serializePagedRotatingCompanion(expectedOffset:)`
returns nil on mismatch, hence `companion=false`. The disk tier has no
equivalent guard (`expectedOffset` appears only in the paged path), so the same
poisoned snapshot went to SSD.

## Why that produces loops instead of a crash

A later fetch content-matches the key, restore reinstates offset 1832, the
engine believes 1831 tokens are covered and re-feeds the boundary token the
cache already contains. One duplicated position shifts every subsequent
position by one — attention is silently wrong for the entire continuation
while generation stays fluent. Each following turn then captures its N-1 seed
FROM that corrupted state and stores it, so the corruption compounds per tool
round.

DSV4 is hit hardest because it marks itself paged-incompatible: the disk tier
is its only reuse path, so **every** turn crosses this seam.

## What was ruled out first, by test

- raw KV round-trip and partial restore (`restore(B)+feed(N-B) == feed(N)`,
  incl. exact ring-wrap boundaries and the N-1 seed) — #207, passes
- DSV4 pool/buffer survival under realistic state — passes
- sampler (agent `temperature: 0` was real, fixed, loop persisted) — osaurus#2274
- DSML parser — tool chains executed correctly before degradation

## The fix

`storeAfterGeneration` refuses the entire store when any layer's `offset`
disagrees with the boundary token count, with a `VMLX_CACHE_FETCH_TRACE` line
naming both. A skipped entry costs one prefill; a poisoned entry costs
correctness for every later turn.

## Validation

- New unit test reproduces the poison without a model: RotatingKVCache fed 12
  tokens stored under an 11-token key → previously **HIT with one extra token
  of state**, now MISS. Healthy path (offset == key) still stores and serves.
- **123 tests / 12 suites** pass with the guard: hybrid topologies
  (Nemotron-Omni Mamba layouts), DSV4 + Zaya disk round trips, TQ serializer,
  boundary derivation, growing-chat sources.

## Deployment note

Machines that ran disk L2 before this fix may hold poisoned entries — clear
the cache directory once after upgrading. Live proof on the affected machine
follows in a comment before merge.

### DIFF
--- a/Libraries/MLXLMCommon/Cache/CacheCoordinator.swift
+++ b/Libraries/MLXLMCommon/Cache/CacheCoordinator.swift
@@ -861,6 +861,36 @@ public final class CacheCoordinator: @unchecked Sendable {
         let traceCacheStore =
             ProcessInfo.processInfo.environment["VMLX_CACHE_FETCH_TRACE"] == "1"
 
+        // A boundary entry asserts "this cache covers exactly these tokens".
+        // The paged tier already enforces that for its rotating companion —
+        // `serializePagedRotatingCompanion(expectedOffset:)` returns nil on a
+        // mismatch, observed live as `companion=false rotatingOffsets=[1832]`
+        // under `tokens=1831`. The disk tier had no equivalent guard, so the
+        // very same poisoned snapshot was persisted to SSD.
+        //
+        // A later fetch content-matches the key, restores one more token of
+        // state than the key claims, and the engine re-feeds the boundary
+        // token the cache already contains. Every subsequent position shifts
+        // by one, attention is silently wrong for the whole continuation, and
+        // each following turn stores a new seed derived FROM that corrupted
+        // state — compounding per turn. Live signature: fluent DSV4 agent
+        // turns degenerating into non-converging loops, only with disk L2
+        // enabled.
+        //
+        // Refuse the whole store instead: a skipped entry costs one prefill,
+        // a poisoned entry costs correctness for every later turn.
+        if let cache {
+            let offsets = Set(cache.map(\.offset))
+            if offsets.contains(where: { $0 != totalTokens }) {
+                if traceCacheStore {
+                    FileHandle.standardError.write(Data(
+                        ("[vmlx][cache/store] REFUSED offset/key mismatch"
+                            + " tokens=\(totalTokens) offsets=\(offsets.sorted())\n").utf8))
+                }
+                return
+            }
+        }
+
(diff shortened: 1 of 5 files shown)


jjang-ai:
## Live proof — guarded build, DSV4-Flash, multi-turn tool chain

Deployed this fix to the affected machine (osaurus repinned to 2c9437db, poisoned cache purged once, shipped defaults: prefix ON / disk ON / paged OFF), then drove a real agent chain: file-write request → todo ×2 → applescript ×2 → follow-up user turn, prompts growing 2510 → 5792 → 9157 tokens.

Trace (`VMLX_CACHE_FETCH_TRACE=1`), verbatim:

```
[vmlx][cache/disk-store] count=2325 ... count=2509 ... count=5791
[vmlx][cache/fetch] HIT disk boundary=2325 remaining=185
[vmlx][cache/fetch] HIT disk boundary=2509 remaining=3283
[vmlx][cache/fetch] HIT disk boundary=5791 remaining=3224
[vmlx][cache/fetch] HIT disk boundary=5791 remaining=3366
```

Every tool round stores and the next round reuses — the exact multiturn-second-tool-prompt shape that previously degenerated. Visible behavior at 9k+ tokens: "Thought for 4m 45s · 847 chars" then action — reasoning terminates; TTFT 21.2s / 14.7 tok/s; no verbatim loops, no non-converging re-verification. Zero `REFUSED` lines from a clean cache, as expected — the guard exists to stop the anomalous snapshot when it appears, and the poisoned pre-fix entries it would have refused were purged on upgrade.

Before/after on identical workload shape: unguarded binary + disk L2 → non-converging reasoning loops within a few tool rounds (reproduced twice, plus 111 MB of re-poisoned entries accumulated in minutes); guarded binary + same defaults → coherent through the full chain.

PR #220, opened 2026-08-10
NemotronHOmni: support items holding images of differing aspect ratios

rcfa:
## Problem

`nemotronOmniPreprocessImages` gives each image its own aspect-preserving resolution, then requires
every raster to share a shape so it can `MLX.stacked` them into a rectangular `(N, 3, H, W)`. Any
item holding two images of different aspect ratios therefore throws:

```
NemotronHOmni Code=-12 "Nemotron Omni image batches with mixed dynamic resolutions are not yet supported"
```

Measured on MMMU: **42 of 847 items (5%)** hold more than one image, and every one of them was
unscoreable on this model family. The limitation matters well beyond benchmarks — in document
analysis there is no guarantee about how many images a page carries or what shape they are.

## Why stacking was never required

`extractImageEmbeds` derives its patch grid from `pixelValues.dim(2)/dim(3)` — **one** grid for the
whole tensor:

```swift
let gridH = pixelValues.dim(2) / config.visionPatchSize
let gridW = pixelValues.dim(3) / config.visionPatchSize
precondition(gridH * gridW == P, …)
```

That is correct exactly when the images match, or when there is a single image. So embedding one
image at a time is **exact**, not an approximation, and the flat `(N*tokens, llm_hidden)` return
concatenates cleanly along axis 0.

`LMInput.ProcessedImage` already documents the layout this needs — *"Concatenated pixels from one or
more images"* with `frames: [THW]` describing each. The information was representable all along;
`prepare` simply stamped every frame with the batch's single shape, which was only correct because
the preprocessor had forced the images to be equal.

## Change

- **Uniform images keep the rectangular `(N, 3, H, W)` layout, byte for byte.** Every existing
  caller and code path is untouched, including video and the single-image case (95% of MMMU items).
- **Ragged images** are concatenated flat, with real per-image `frames`.
- `extractImageEmbeds` takes `frames` and, **only for a flat buffer**, slices each image out,
  embeds it alone, and concatenates. A 4-D tensor still means rectangular, so `frames` only becomes
  load-bearing when it has to be.
- `RunBench/OmniBench.swift` updated for the third tuple element; its `pixels.dim(2)/dim(3)` reads
  are unaffected because a single image still returns rectangular.

## Verification

On-device, `Nemotron-3-Nano-Omni-30B-A3B-JANGTQ4`, MMMU, 50 items, identical sample:

| | mmmu | items refused |
|---|---|---|
| before | 28/50 = 56.0% | 1 |
| after | **29/50 = 58.0%** | **0** |

The previously-unscoreable multi-image item now scores, and scores correctly.

One note in case it saves someone time: `rasterizeImage` returns **3-D** `(3, H, W)` — which is why
`stacked(axis: 0)` yields `(N, 3, H, W)`. Per-image dims are therefore read from the trailing axes
(`ndim-2`, `ndim-1`) so the code holds whether or not a leading batch axis is ever added.

### DIFF
--- a/Libraries/MLXVLM/Models/NemotronHOmni/NemotronHOmni.swift
+++ b/Libraries/MLXVLM/Models/NemotronHOmni/NemotronHOmni.swift
@@ -259,7 +259,7 @@ public class NemotronHOmni: Module, VLMModel, KVCacheDimensionProvider, LoRAMode
         var imageEmbedCount = 0
         var videoRetention: (keepIndices: [Int], totalTokens: Int)? = nil
         if let pixelValues = input.image?.pixels {
-            visualEmbeds = extractImageEmbeds(pixelValues: pixelValues)
+            visualEmbeds = extractImageEmbeds(pixelValues: pixelValues, frames: input.image?.frames)
             imageEmbedCount = visualEmbeds?.dim(0) ?? 0
         }
         if let video = input.video {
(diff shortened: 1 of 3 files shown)


jjang-ai:
Reviewed the preprocessing specifically, as you suggested. It holds up, and the reason is worth writing down because it is not obvious from the diff alone.

## Why `frames` is safe to use as pixel dims

The consumer slices with `3 * f.h * f.w`, i.e. it reads `frames` as **pixel** extents. That looked like the risky assumption — `THW` is grid/patch units in other model families — so I checked what this model previously put there:

```swift
// before
frames: tokenCounts.map { _ in THW(1, pixels.dim(2), pixels.dim(3)) }
```

`pixels.dim(2/3)` on an `(N, 3, H, W)` tensor is already H and W in pixels. So `frames` has always meant pixel dims *in NemotronHOmni*, and this PR does not redefine it — it stops stamping every image with the batch shape and gives each its own. That makes the slice arithmetic exact rather than coincidental.

## The gate is tight

```swift
guard pixelValues.ndim == 1, let frames, !frames.isEmpty else {
    return extractImageEmbedsUniform(pixelValues: pixelValues, video: video)
}
```

Ragged only when the buffer is flat *and* frames exist. Uniform batches still hit `MLX.stacked` and the original single-grid path, so the 95%-of-items case is byte-identical, and video is untouched. `dims` being read from `ndim-2 / ndim-1` rather than fixed indices is the right call too — it survives a leading batch axis if one ever appears.

The N-tower-calls cost is the honest tradeoff and you already flagged it; for 2-5 images it is the correct trade against refusing the item outright.

## One thing to confirm before merge

Both `nemotronOmniPreprocessImages` and `NemotronHOmniProcessor.preprocess` change their return arity (2-tuple → 3-tuple), and both are `public`. In-tree that is fine — the build catches every call site — but it is a source-breaking change for anyone calling the preprocessor directly from outside the package. If that matters for your release cadence, adding a 2-tuple overload that discards `dims` would keep it additive; if not, the arity change is cleaner and I would take it as-is.

No objection from me on the preprocessing itself.

PR #222, opened 2026-08-10
Add the Muse Glimmer 30B runtime (text, vision, video, ATEM tools)

jjang-ai:
Adds a runtime for Muse Glimmer 30B, a dense image+video VLM (`model_type: muse_glimmer`).

Architecture notes: `Libraries/MLXVLM/Models/MuseGlimmer_ARCHITECTURE.md`
Live gate (not yet run): `Libraries/MLXVLM/Models/MuseGlimmer_PROOF_MATRIX.md`

## Text tower

The config inherits Gemma2's shape but diverges in five places. Each keeps every
tensor the right size and only *degrades* output — the class of bug that gets
blamed on the quant — so each has a numeric test:

| # | Divergence |
|---|---|
| 1 | `CenteredRMSNorm`: weights stored zero-centered, so the `(1 + w)` term is load-bearing |
| 2 | Scaleless QK-norm; `qk_scale_factor` 3.87 on **Q only**; scale stays `head_dim ** -0.5` |
| 3 | **NoPE** on full-attention layers — every 4th counted back from the last (3, 7, …, 51) |
| 4 | Attention **output gate** reading the layer input, applied before `o_proj` |
| 5 | Logits × `output_multiplier` then `20·tanh(x/20)`; **no** embedding normalizer |

Sliding layers (window 2048) get rotating caches; the 13 full/NoPE layers get
standard caches.

## Vision tower and processor

50-layer windowed ViT: linear patch projection (14×14 across 2 temporal frames ×
3 channels = 1176 wide), a learned 32×32 position grid, LayerNorm-with-bias
blocks, 2×2 spatial merge, adapter, then projection into the 6656 text width.

- The position grid is resampled with an explicit four-corner bilinear gather.
  Out-of-range corners contribute weight zero rather than being clamped onto a
  neighbour — clamping the index while keeping its weight would smear the border
  embeddings inward.
- Patches arrive grouped by merge block, so position rows are walked in the same
  block order. Getting that wrong lands every position embedding on the wrong
  patch with no shape error to catch it.
- The adapter runs its activation after *both* layers, which reads like a typo.
- Images are budgeted by merged tokens (`max_image_tokens`), not by a pixel
  range, so the token budget is converted to a pixel ceiling before resizing.

## ATEM tool calls

A new XML-ish envelope carrying one or more invokes. The model's own instructions
say the output "is not expected to be valid XML and is parsed with regular
expressions", so the parser locates specific closers instead of XML-parsing.
Covered: `<`/`>` inside values (the gemma4 stray-angle desync), dropped quotes,
truncated envelopes, malformed JSON in an array parameter, nameless invokes, and
orphan closers registered so a stray run never reaches visible text.

## Two load-time hazards found in the shipped bundles

- `out_hidden_size: 6144` is the **post-merge** width (4 × 1536), not a tower
  width. The real projection is 4096 → 6656.
- The vision tower ships **unquantized F16** while the text tower is 4/6-bit, so
  the prepare path takes its dtype from the vision tower rather than assuming one
  checkpoint dtype.

Also: the shipped `layer_types` tail breaks its own period (full attention on 47
then 49), so the derived default is a fallback only; the test pins the exact
positions read off the checkpoint.

## Verification

25 tests across 3 new suites pass; full build clean; the existing 189 tool-call
tests still pass with no regression from the new format case.

**Not yet run**: the live proof matrix — sustained tok/s, EOS, stop-mid-turn,
SSD block match/write/reuse, TTFT, reasoning on/off, effort flips, tool toggles,
image and video turns. Nothing there is claimed.


jjang-ai:
Live-proof status after the runtime hardening pass, all on the real JANG bundles in the dev app GUI:

**Coherent generation** — the token-soup root cause was the checkpoint's *normed embedding* (`MuseGlimmerTextNormedEmbedding`: scaleless RMSNorm over the lookup, deliberately unfoldable). Tokenizer/template/specials were proven clean first via bundle probes (specials atomic in the 200000+ range, one `Reasoning strength:` line, round-trip exact), isolating the fault to the forward pass. Post-fix the model reasons and answers fluently.

**Reasoning strength wired end to end and proven from live salts** — the user's own session log shows `strength=high` ×6, `strength=low` ×3, `strength=medium` ×3 arriving in the cache scope salt with distinct salts per strength (no key collision). Muse enforces effort via the system-prompt sentence (`reasoning_strength` template variable), level set low/medium/high/**xhigh**; lower strengths legitimately produce less/no `to=self` thinking.

**Channel envelope parsed** — live transcript exposed the turn structure `<|start|>assistant to=self<|message|>…<|eom|>` then `to=user<|message|>…`; the reasoning parser now routes `to=self` to the thinking rail and strips every marker. Tests pin the exact live transcript.

**Crash class closed** — the rotating-KV overrun (disk-restored `idx`/width divergence + trailing single-token prefill chunk) is fixed at three layers: chunked prefill absorbs the stray token, the dispatcher only takes the in-place path when provably safe, and `updateInPlace` falls back to concat rather than computing a cleverer index. Foreground+window+generation no longer crashes (was 6/6 reproducible).

Focused suites: 85/85. Remaining live legs (tool loop on the parser build, cold-boot SSD reuse, sustained tok/s) run next; stale source-contract anchors were refreshed to the current, stronger spellings.

PR #242, opened 2026-08-12
ChatSession: honor GenerateParameters.draftStrategy (native MTP was silently ignored on the chat path)

rcfa:
### What

`ChatSession` never consulted `GenerateParameters.draftStrategy`. Its non-diffusion branch
always constructed a plain `TokenIterator`, so a caller using the high-level chat API got
autoregressive decode regardless of what it set — no error, no warning.

`generate(...)` and `generateTokensTask(...)` both already dispatch on `draftStrategy` and build a
`NativeMTPTokenIterator` for `.nativeMTP`. This adds the same dispatch to `ChatSession`.

### How it showed up

The failure is silent, which is what makes it worth fixing rather than documenting. Loading a
bundle with `LoadConfiguration.nativeMTP = true` succeeds, the model reports
`nativeMTPAvailable`, `draftStrategy` is accepted, and throughput is simply unchanged. We found
it by A/B-ing a bundle that ships `vmlx_mtp_tuning.json` with speculation on and off and getting
identical tok/s — the flag was being dropped on the only path our harness uses.

### Safety

Guarded by the same `parameters.canUseNativeMTP(for: input)` predicate the other two entry points
use — greedy-only, no media content, unbounded KV window — so it cannot change generated text,
only whether speculation is used. A model that is not a `NativeMTPModel` throws
`NativeMTPRuntimeError.modelDoesNotExposeNativeMTP`, matching `generate(...)` rather than
silently falling back, so a mis-set strategy stays visible.

Callers using `.none` / `nil` are unaffected — they fall through to the existing path unchanged.

### Testing

Built and run on device against `Qwen3.6-27B-MXFP4-MTP` (native MTP, depth 2 from its tuning
row). Before: on and off measure identically. After: the drafter demonstrably engages and the two
arms diverge.

### One finding this unblocked, in case it's useful

With the dispatch in place we could finally A/B native MTP end to end. On an M4 Max against
`Qwen3.6-27B-MXFP4-MTP` at its tuning row's `best_depth=2`, counterbalanced ON/OFF/OFF/ON, 256
tokens, idle machine:

| prompt | MTP off | MTP on |
|---|---|---|
| open-ended | 27.6, 27.5 tok/s | 23.8, 23.5 tok/s |
| `deterministic_count_96` (the tuning row's own class) | 27.4, 27.2 | 23.8, 23.1 |

~14% slower both ways, including on the prompt class where `vmlx_mtp_tuning.json` records
24.655 → 45.712 (1.854x). Prompt predictability was the obvious explanation and the second row
falsifies it.

**Correction — I profiled it, and my guess above was wrong.** Running with
`VMLX_NATIVE_MTP_TRACE=1` on the same setup, over 256 tokens:

| counter | value |
|---|---|
| `iteratorWallSec` | 11.448 |
| `verifyMainSec` | **10.777 (94%)** — real target-model forwards |
| `materializeSyncSec` | 0.162 (1.4%) |
| `cacheCommitSec` / `cacheStateSec` | 0.000 / 0.000 |
| `mtpDraftSec` | 0.041 |

I had speculated that the speculative iterators bypassing `CompiledDecodeTrace` /
`CompilableKVCache` was the cause. Device syncs and cache bookkeeping are 1.4% and 0% — that theory
is dead, and compiling the speculative path would attack almost nothing. Please disregard it.

What the trace actually shows is that speculation is barely running:

| counter | value |
|---|---|
| `outputTokens` | 256 |
| `targetForwards` | **254** — ~one target forward per token, i.e. no speculative saving at all |
| `arFallbackTokens` | **223 of 256 (87%)** — most tokens came from autoregressive fallback |
| `verifyCalls` | 16, `avgCommittedPerVerify` 2.06 → only ~33 tokens came from speculation |
| `depth` / `activeDepth` | **1 / 1**, though the bundle's tuning row says `best_depth=2` |
| `verifierMode` | **`sequential_repair`**, though `chunk_lazy_repair` was requested |
| acceptance | `acceptedByDepth=0:1,1:15`, `hybrid_warmup_avg_accept=`**`0.94`** |

The 14% is the MTP machinery being paid for while doing one target forward per token. The puzzling
part is that acceptance is 0.94 — drafts are accepted 15 times in 16, which ought to encourage more
speculation, not 87% fallback. Three things I can't explain from outside the subsystem:

1. depth downshifted 2 → 1 with `adaptiveDownshifts=0`, so seemingly not the adaptive policy;
2. `verifierMode` resolved to `sequential_repair` despite `chunk_lazy_repair` being requested;
3. 87% AR fallback despite 94% acceptance.

If those are expected behaviour on this bundle, ignore the noise. If not, the arithmetic changes
completely: depth 2 at 0.94 acceptance should roughly halve target forwards, which is about where
the tuning row's 1.854x would come from — so the vendor number may be perfectly reproducible on a
path that currently isn't being taken.

None of this affects the patch itself, which is a one-branch dispatch fix for a parameter that is
otherwise silently ignored.

### DIFF
--- a/Libraries/MLXLMCommon/ChatSession.swift
+++ b/Libraries/MLXLMCommon/ChatSession.swift
@@ -453,6 +453,30 @@ public final class ChatSession {
                                 iterator: iterator,
                                 toolSchemas: input.toolSchemas
                             )
+                        } else if let strategy = generateParameters.draftStrategy,
+                            case .nativeMTP(depth: let depth, verifierMode: _) = strategy,
+                            generateParameters.canUseNativeMTP(for: input)
+                        {
+                            // Native model-owned MTP speculative decode. `generate(...)` and
+                            // `generateTokensTask(...)` already dispatch on `draftStrategy`, but
+                            // `ChatSession` did not — so a caller using the high-level chat API got
+                            // plain autoregressive decode no matter what it set, with no error and
+                            // no log line to say so. Eligibility is `canUseNativeMTP` (greedy-only,
+                            // no media, unbounded KV), so this cannot change what text is produced.
+                            guard let nativeModel = model as? any NativeMTPModel else {
+                                throw NativeMTPRuntimeError.modelDoesNotExposeNativeMTP
+                            }
+                            let iterator = try NativeMTPTokenIterator(
+                                input: input, model: nativeModel, cache: kvCache,
+                                parameters: generateParameters, depth: depth,
+                                cacheCoordinator: cacheCoordinator)
+                            (stream, task) = MLXLMCommon.generateTask(
+                                promptTokenCount: input.text.tokens.size,
+                                modelConfiguration: modelConfiguration,
+                                tokenizer: tokenizer,
+                                iterator: iterator,
+                                toolSchemas: input.toolSchemas
+                            )
                         } else {
                             let iterator = try TokenIterator(
                                 input: input, model: model, cache: kvCache,
--- a/Tests/MLXLMTests/ChatSessionRuntimeWiringSourceCoverageTests.swift
+++ b/Tests/MLXLMTests/ChatSessionRuntimeWiringSourceCoverageTests.swift
@@ -21,6 +21,31 @@ struct ChatSessionRuntimeWiringSourceCoverageTests {
         #expect(source.contains("let iterator = try TokenIterator("))
     }
 
+    /// The branch this pins is the one the bug was: `ChatSession` never consulted
+    /// `draftStrategy`, so a caller that set it got plain autoregressive decode with no error and
+    /// no log line. Deleting the `else if` restores that silence — and nothing else in-tree would
+    /// fail — so this test exists to make the branch's absence loud.
+    ///
+    /// Source-level rather than behavioural because exercising the dispatch needs a real bundle
+    /// with MTP tensors; this pins the wiring, and the guard below pins the safety condition.
+    @Test("streamMap dispatches native MTP when draftStrategy requests it")
+    func streamMapDispatchesNativeMTPForDraftStrategy() throws {
+        let source = try Self.source("Libraries/MLXLMCommon/ChatSession.swift")
+        #expect(source.contains("generateParameters.draftStrategy"))
+        #expect(source.contains("case .nativeMTP(depth: let depth, verifierMode: _)"))
+        #expect(source.contains("let iterator = try NativeMTPTokenIterator("))
+    }
+
+    /// Speculation must stay gated on `canUseNativeMTP(for:)` — greedy-only, no media content,
+    /// unbounded KV. That predicate is what makes "this cannot change what text is produced" true;
+    /// dispatching without it would turn a performance path into a correctness change.
+    @Test("native MTP dispatch stays gated on canUseNativeMTP")
+    func nativeMTPDispatchIsGated() throws {
+        let source = try Self.source("Libraries/MLXLMCommon/ChatSession.swift")
+        #expect(source.contains("generateParameters.canUseNativeMTP(for: input)"))
+        #expect(source.contains("NativeMTPRuntimeError.modelDoesNotExposeNativeMTP"))
+    }
+
     @Test("streamMap consumer termination cancels the producer task")
     func streamMapTerminationCancelsProducerTask() throws {
         let source = try Self.source("Libraries/MLXLMCommon/ChatSession.swift")


jjang-ai:
Reviewed. This is the same failure shape as the Nemotron MTP wiring we chased recently — a setting the user sets, that never reaches the engine, with nothing to say so. Worth taking.

Two things I checked rather than assumed:

**The throw is consistent, not new.** `ChatSession` now throws `modelDoesNotExposeNativeMTP` where a caller previously got silent AR decode, which is a behaviour change for anyone who set `draftStrategy` on a non-MTP model. But it matches every other dispatch site — `Evaluate.swift:3589`, `Evaluate.swift:4016`, `BatchEngine.swift:1165` all throw the same error in the same situation. So this makes `ChatSession` agree with the rest of the codebase rather than inventing a policy, which is the right call: an explicit `.nativeMTP` request that cannot be honoured should be loud.

**"Cannot change what text is produced" holds.** Gating on `canUseNativeMTP(for:)` keeps it greedy-only with no media and unbounded KV, and a correct MTP verifier under argmax accepts exactly the tokens AR would have emitted. So this is a latency change, not an output change — provided the verifier is right, which is separately covered.

The one thing I would add before this is done: nothing in-tree fails if the `else if` is deleted. That is precisely the bug being fixed here — a silently-skipped branch — so it would be a shame to fix it without a test that would catch it coming back. A session-level assertion that a `.nativeMTP` `draftStrategy` produces a `NativeMTPTokenIterator` (or that a non-MTP model throws) would pin it.

#243 alongside this one looks right too — driving the quantize filter off `weights["\(path).scales"] != nil` rather than assuming every `Linear` is quantized is the correct instinct, since real bundles do leave selected layers in full precision and blanket-quantizing those fails the same verify from the other side.

rcfa:
Fair, and it is my own rule turned back on me — a test that cannot fail is not protecting anything. Added, using the convention already in the tree (`ChatSessionRuntimeWiringSourceCoverageTests`), so it sits with the other wiring guards rather than inventing a mechanism:

- **`streamMap dispatches native MTP when draftStrategy requests it`** — pins `generateParameters.draftStrategy`, the `case .nativeMTP(depth:verifierMode:)` match, and `NativeMTPTokenIterator(`.
- **`native MTP dispatch stays gated on canUseNativeMTP`** — pins `canUseNativeMTP(for: input)` and the `modelDoesNotExposeNativeMTP` throw, because that predicate is what makes "cannot change what text is produced" true. Dispatching without it would turn a performance path into a correctness change.

**Mutation-verified**, since that was the whole point. Deleting the `else if` branch:

```
✗ streamMap dispatches native MTP when draftStrategy requests it   — 3 failed expectations
✗ native MTP dispatch stays gated on canUseNativeMTP               — 2 failed expectations
✓ streamMap passes container cache coordinator into TokenIterator  — still passes
✓ streamMap consumer termination cancels the producer task         — still passes
```

The two pre-existing tests surviving is the part I wanted to see: the mutation was surgical, and the new guards are what caught it.

Source-level rather than behavioural because exercising the dispatch needs a real bundle with MTP tensors. If you would rather have it as an integration test against a fixture, say so and I will do that instead — I went with the cheaper form that matches what is already there.

---

**One correction I owe you.** My PR body previously suggested speculation might be losing because the speculative iterators bypass `CompiledDecodeTrace` / `CompilableKVCache`. I have retracted that — I profiled it with `VMLX_NATIVE_MTP_TRACE=1` and it is wrong: `materializeSyncSec` is 0.162 s of 11.448 s wall (**1.4%**), `cacheCommitSec` and `cacheStateSec` are both 0.000, and `verifyMainSec` is 10.777 s (**94%**), i.e. real target-model forwards. Compiling that path would attack almost nothing. Sorry for putting an unverified mechanism in front of you.

What the trace actually shows is that speculation barely runs: `targetForwards=254` for `outputTokens=256` — one target forward per token, no speculative saving at all — with **223 of 256 tokens (87%) from autoregressive fallback**, `depth`/`activeDepth` **1/1** though the bundle tuning row says `best_depth=2`, and `verifierMode` resolving to `sequential_repair` though `chunk_lazy_repair` was requested. Meanwhile acceptance is **0.94** (`acceptedByDepth=0:1,1:15`).

That last combination is what I cannot explain from outside: 15 of 16 drafts accepted ought to encourage *more* speculation, not 87% fallback. If the depth downshift (with `adaptiveDownshifts=0`) and the verifier-mode substitution are expected on this bundle, ignore the noise. If not, the arithmetic changes a lot — depth 2 at 0.94 acceptance should roughly halve target forwards, which is about where the tuning row’s 1.854x would come from. Possibly your number is perfectly reproducible on a path that simply is not being taken right now.

None of that affects this patch either way; it is a dispatch fix for a parameter that is otherwise silently ignored.

jjang-ai:
Checked this specifically for blast radius on our shipped runtimes, and the answer is **none** — for two independent reasons, either of which would be enough.

**1. Osaurus never takes this path.** It drives inference through `container.makeBatchEngine(maxBatchSize:)` → `BatchEngine` (`MLXBatchAdapter.swift:396`), not `MLXLMCommon.ChatSession`. The `ChatSession` symbols in our tree are our own UI/session model (`Managers/Chat/ChatSessionExporter.swift` and friends), an unrelated type of the same name. So this fixes the high-level chat API for embedders who use it, and cannot reach our decode path at all.

**2. The gate is exactly the MTP policy we already enforce.** This was the thing I actually wanted to verify, because "dispatch native MTP" would be alarming if it were unconditional. `canUseNativeMTP(for:)` resolves to:

```
temperature == 0 && topP >= 1 && topK == 0 && minP == 0
  && repetitionPenalty/presencePenalty/frequencyPenalty all unset-or-neutral
  && initialSuppressTokens.isEmpty
  && !input.hasMediaContent
  && maxKVSize == nil
```

That is the same greedy-only contract we landed on our side (skip native MTP at temperature > 0), including the two subtle members I'd have flagged had they been missing:

- `initialSuppressTokens.isEmpty` — our minimum-reasoning-floor masks logits per *sampled* token, and drafted MTP tokens bypass it. A floor-enforcing request must fall back to AR or the floor silently stops applying. Covered.
- `maxKVSize == nil` — bounded windows make attention slots `RotatingKVCache`, and a rejected draft cannot be un-written once the ring wraps. Declining is right; our Gemma-4 rotating and hybrid stacks are exactly the shapes that would have been corrupted here.

So "this cannot change what text is produced" holds: every path where speculation could alter output is excluded before dispatch.

The source-level tests are the right call given exercising the branch needs a real MTP bundle, and pinning `canUseNativeMTP(for: input)` as its own assertion is the part that matters — an accidental ungating is the failure that would actually hurt, and it now fails loudly.

LGTM.

jjang-ai:
### QA — wiring contracts green, and the safety gate still holds

**The gate is the important part.** `draftStrategy` now reaches the chat path,
so the thing to prove is that it cannot bypass the eligibility check. The PR's
own coverage confirms it — 4/4:

```
✔ native MTP dispatch stays gated on canUseNativeMTP
✔ streamMap dispatches native MTP when draftStrategy requests it
✔ streamMap passes container cache coordinator into TokenIterator
✔ streamMap consumer termination cancels the producer task
```

That first one matters beyond this PR: `isNativeMTPLosslessGreedyEligible` also
refuses MTP while a reasoning ceiling is active (#252), and refuses it under a
bounded KV window. Honouring `draftStrategy` here does not weaken either — the
request still has to pass the gate.

**No family regressed.** Stacked on main with #243:

`LFM2.5-VL-3B-JANG_4M`, full variating pattern (reasoning ON/OFF, text↔image↔
text-after-image, tools offered/called, image AND tools in one turn, two images
in one turn, grow control, verbatim replay):

```
diff base.txt pr243_242.txt  ->  IDENTICAL   (all 18 turn/probe lines)
```

including `HIT disk 22/38 (57% reused)` on the constant-head control and
`HIT disk 26/26 (100% reused)` on the verbatim replay.

Prefix-cache reuse on two other architectures, unchanged from main:

| family | cache |
|---|---|
| Bonsai-27b-Ternary (`Qwen35` hybrid SSM) | `HIT tier=paged matched=18/53` — passed |
| gemma-4-E2B-qat (dense + rotating SWA) | `HIT tier=paged matched=31/55` — passed |

Native MTP is default-off, so the identical output is the expected result: this
makes the parameter reachable, it does not turn speculation on.

jjang-ai:

PR #246, opened 2026-08-12
Tokenizers: split added tokens with a trie (136k-token omni bundles cost ~28s per special token)

rcfa:
### What

Added-token splitting is O(text x tokenCount) per `tokenize(text:)` call. For models with large
added-token tables that makes prompts take minutes.

`PreTrainedTokenizer.init` joins every added token into one `NSRegularExpression` alternation, each
in its own capture group. Two costs then scale with the token count on every match: ICU has no
literal-alternation optimisation, and `String.split(by:)` walks
`(0..<match.numberOfRanges).reversed()` per match with an `NSRange` -> `Range<String.Index>`
conversion at each step — and since tokens are sorted longest-first, the real match sits at a low
index, so that scan traverses nearly the whole capture array.

### Why it shows up now

Discrete-codebook multimodal models. Apertus 1.5 Omni declares **136,366** added tokens — 131,072
`<|visual token N|>` and 4,096 `<|audio token N|>`. That is a correct description of its visual and
audio codebooks, not a malformed bundle, and this shape will become more common.

The cost is ~28 s per special token in the prompt. Measured on an M4 Max, release build:

| | before | after |
|---|---|---|
| TTFT (76-token prompt) | 214,420 ms | **299 ms** |
| load | 86.2 s | **2.3 s** |
| decode | 66.4 tok/s | 64.4 tok/s |

Decode was always healthy — the model was fast and simply unreachable behind tokenization.

The existing `isUnusedPlaceholderAddedToken` filter for `<unusedN>` suggests this shape has been hit
before; it matches none of Apertus's tokens, and widening it would be wrong, since visual tokens are
real tokens the model uses.

### How

A trie: O(text) regardless of token count, because shared prefixes (`<|visual token `) collapse into
one path.

Semantics are preserved exactly — longest match wins at each position (the regex sorts alternatives
longest-first and ICU alternation is first-match-wins, the same rule), matched separators are emitted
as their own sections, empty sections omitted.

**Guarded**: used only when no added token sets `lstrip`/`rstrip`, since those wrap the token in
`\s*` and are not a literal match. Those bundles keep the regex path.
`SWIFT_TRANSFORMERS_NO_ADDED_TOKEN_TRIE=1` forces the regex back, for A/B measurement or for
bisecting a suspected tokenization difference.

### Testing

Correctness is the thing that matters here — a silent tokenization change would corrupt downstream
results while looking like a win. An 8-item arc_challenge cell on `Nemotron-3.5-Lightning-30B-A3B`
scores **6/8 identically on both paths across four counterbalanced runs**.

That same A/B also shows the trie is **not** a general speedup: for a 1,000-added-token model the two
paths are within noise (trie 23 s / 16 s, regex 17 s / 17 s). This fixes pathological bundles and
makes no claim beyond that.

### DIFF
--- a/Tests/MLXLMTests/AddedTokenTrieSourceCoverageTests.swift
+++ b/Tests/MLXLMTests/AddedTokenTrieSourceCoverageTests.swift
@@ -1,0 +1,46 @@
+import Foundation
+import Testing
+
+/// Pins the two properties of `AddedTokenTrie` that a plausible "simplification" would break, and
+/// that no other test in-tree would catch.
+///
+/// Source-level because `AddedTokenTrie` and `String.split(by:)` are both internal to `Tokenizers`,
+/// which has no test target in this tree. The behavioural differential that found the real bug was
+/// run out-of-tree (PR #246); this keeps the fix from silently reverting.
+@Suite("AddedTokenTrie source coverage")
+struct AddedTokenTrieSourceCoverageTests {
+    private static func source(_ relativePath: String) throws -> String {
+        let repoRoot = URL(fileURLWithPath: #filePath)
+            .deletingLastPathComponent()
+            .deletingLastPathComponent()
+            .deletingLastPathComponent()
+        return try String(contentsOf: repoRoot.appending(path: relativePath), encoding: .utf8)
+    }
+
+    /// The walk MUST be over Unicode scalars, not `Character`s.
+    ///
+    /// `Array(text)` yields extended grapheme clusters, so a token's final scalar fuses with a
+    /// following combining mark and the token stops matching entirely — `"<|im_end|>\u{0301}rest"`
+    /// was left unsplit while the regex arm split it correctly. Scalars are also the level the
+    /// regex operates at, which is what makes the two arms identical by construction.
+    @Test("trie walks unicode scalars, not grapheme clusters")
+    func trieWalksUnicodeScalars() throws {
+        let source = try Self.source("Vendors/swift-transformers/Sources/Tokenizers/AddedTokenTrie.swift")
+        #expect(source.contains("[Unicode.Scalar: Node]"))
+        #expect(source.contains("text.unicodeScalars"))
+        #expect(source.contains("for scalar in token.unicodeScalars"))
+        // The exact regression: walking `Character`s.
+        #expect(!source.contains("let characters = Array(text)"))
+        #expect(!source.contains("[Character: Node]"))
+    }
+
+    /// Longest match must win, which requires CONTINUING the walk past a terminal node rather than
+    /// returning at the first one. The regex gets this from longest-first alternation ordering; if
+    /// the trie returned early, `</think>` would split as `<` + `/think>` on any token set where a
+    /// shorter token is a prefix of a longer one.
+    @Test("trie keeps walking past a terminal node so the longest token wins")
+    func triePrefersLongestMatch() throws {
+        let source = try Self.source("Vendors/swift-transformers/Sources/Tokenizers/AddedTokenTrie.swift")
+        #expect(source.contains("if node.isTerminal { longest = index }"))
+    }
+}
--- a/Vendors/swift-transformers/Sources/Tokenizers/AddedTokenTrie.swift
+++ b/Vendors/swift-transformers/Sources/Tokenizers/AddedTokenTrie.swift
@@ -1,0 +1,94 @@
+import Foundation
+
+/// Splits text on a set of literal added-token strings in time independent of how many there are.
+///
+/// The regex path this replaces builds ONE `NSRegularExpression` whose pattern is every added token
+/// joined by `|`, each in its own capture group, and runs it on every `tokenize(text:)` call. That is
+/// O(text x tokenCount) twice over: ICU has no literal-alternation optimisation, and
+/// `String.split(by:)` then walks `(0..<match.numberOfRanges).reversed()` per match doing an
+/// `NSRange` -> `Range<String.Index>` conversion at each step.
+///
+/// With a handful of added tokens that is invisible. Discrete-codebook multimodal models change the
+/// scale: Apertus 1.5 Omni declares 136,366 added tokens (131,072 `<|visual token N|>` plus 4,096
+/// `<|audio token N|>`), and each special token in a prompt then costs ~28 SECONDS — a 76-token
+/// prompt took 218 s to reach first token while decode was a healthy 66 tok/s.
+///
+/// A trie makes the scan O(text) regardless of token count, because shared prefixes (`<|visual token `)
+/// collapse into one path.
+///
+/// **Semantics are matched exactly**, so callers see no behavioural change:
+///   * longest match wins at each position — the regex sorts alternatives longest-first and ICU
+///     alternation is first-match-wins, which is the same rule;
+///   * matched separators are emitted as their own sections, interleaved with the text between them;
+///   * empty sections are omitted.
+///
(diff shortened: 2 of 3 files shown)


jjang-ai:
Proofread this one carefully because it changes the tokenize path for **every** bundle we ship, not just the big-added-token ones: I checked 40 local bundles and **all 40 take the trie** (none set `lstrip`/`rstrip`), so the regex arm is dead code for our families and the parity claim is load-bearing.

Also worth knowing: the win is not only Apertus. `JANGQ-AI/AppleScript-8B-JANG_6M` declares **6,417** added tokens, and that model backs our AppleScript subagent — which users have been reporting as sluggish.

**Differential test, trie vs. the existing `split(by:)` regex arm**, both lifted verbatim, over the real special tokens of the families we ship (`<|im_start|>`, `<|im_end|>`, `<think>`, `<tool_call>`/`<arg_key>`, `<start_of_turn>`, laguna's `〈|EOS|〉`/`<assistant>`, plus `<|visual token N|>`):

```
OK   empty                          OK   longest-wins prefix
OK   plain                          OK   leading text / trailing text
OK   single special                 OK   non-BMP emoji (ZWJ family, flag)
OK   chat turn                      OK   combining mark inside text
OK   adjacent specials              OK   CRLF
OK   laguna turn                    OK   token-ish but not a token
OK   nested think                   DIFF combining mark AFTER a token
OK   tool call
                                    differences: 1 of 16
```

Fifteen match exactly, including the two I most wanted to see: empty input (both return `[""]` — `split(by:)` returns `[self]` on no matches, and the trie's `sections.isEmpty ? [text]` lands on the same thing), and longest-match-wins (`<|visual token 12|>` is not shortened to `<|visual token 1|>`; the existing `.sorted { $0.content.count > $1.content.count }` does make ICU first-match-wins agree with the trie).

**The one divergence:**

```
input: "<|im_end|>\u{0301}rest"
trie : ["<|im_end|>́rest"]                 <- token not recognised
regex: ["<|im_end|>", "\u{0301}rest"]      <- token split out
```

`Array(text)` yields Swift `Character`s, i.e. extended grapheme clusters, so the trailing `>` and the combining acute fuse into one cluster and the walk can't match the token's final `>`. The regex works in UTF-16 and doesn't fuse.

**One-line fix, verified:** walk unicode scalars instead of characters — `[Unicode.Scalar: Node]`, `for character in token.unicodeScalars`, `Array(text.unicodeScalars)`, and rebuild the slices via `String.UnicodeScalarView`. I ran the same 16 cases against that variant:

```
differences: 0 of 16
```

Severity is genuinely low — it needs user text to place a combining mark immediately after a literal special token, our templates never do, and the failure mode is "the token gets BPE'd as ordinary text" rather than a crash (arguably the safer direction, since it makes control tokens harder to inject from user text). So I'm not treating it as blocking. But the doc comment promises "**Semantics are matched exactly**, so callers see no behavioural change", and with scalars that sentence becomes true for free.

Nice touch on `SWIFT_TRANSFORMERS_NO_ADDED_TOKEN_TRIE` as a control arm — that's what let me diff the two arms in one binary instead of trusting the description.

rcfa:
Caught a real bug — thank you, and that is exactly the test I should have written myself. The trie's contract is "indistinguishable from the regex arm", so a differential against that arm is the only correct oracle; hand-written expectations would have encoded what I *think* the regex does, which is precisely how this got in.

**Fixed by walking `unicodeScalars` instead of `Character`s.** Your diagnosis was exactly right: `Array(text)` yields extended grapheme clusters, so the token's final `>` fuses with the following combining acute into one cluster and the walk cannot match the token's last scalar. Scalars are also the level the regex operates at, so the two arms are now identical **by construction** rather than by coincidence — which is a better guarantee than patching the one case.

Changed: `[Character: Node]` → `[Unicode.Scalar: Node]`, `Array(text)` → `Array(text.unicodeScalars)`, and the section rebuild goes through `String.UnicodeScalarView` so multi-scalar clusters are reassembled rather than sliced.

Worth stating why this is not a contrived case, since it reads like one: any decomposed (NFD) input can place a combining mark straight after a special token — Vietnamese, Hebrew with niqqud, and macOS filesystem paths are all routinely NFD. A model quoting such text right after `<|im_end|>` would have hit it.

**Would you re-run your 16-case differential against the updated branch?** I would rather have your harness confirm 16/16 than assert it from my side — I do not have the regex arm reachable from a test target here (`AddedTokenTrie` and `String.split(by:)` are both internal to `Tokenizers`, which has no test target in this tree).

Given that, I added a **source-coverage guard** in `Tests/MLXLMTests/` instead, following the convention already used for the ChatSession wiring: it pins the scalar walk (and asserts the `Array(text)` / `[Character: Node]` forms are absent), plus the keep-walking-past-a-terminal-node rule that makes longest-match win. It cannot catch a behavioural divergence the way your differential can, but it stops this specific fix from silently reverting.

Two things from your review I want to acknowledge properly:

**"It changes the tokenize path for every bundle we ship"** — correct, and my PR body undersold that. The guard is narrow (trie only when no token sets `lstrip`/`rstrip`), but "narrow" is not "inert", and 40/40 of your bundles taking the new path is the number that matters. `SWIFT_TRANSFORMERS_NO_ADDED_TOKEN_TRIE=1` restores the regex if anything surfaces.

**`AppleScript-8B-JANG_6M` at 6,417 added tokens** — that is a much better motivating case than mine. Apertus is an extreme outlier at 136k; a model your users actually run, backing a shipped subagent, is the argument for taking this. I had framed it as "fixes one pathological bundle", and it is not.

One honest caveat on scope, since I measured it: this is **not** a general speedup. A counterbalanced A/B on a 1,000-added-token model put trie and regex within noise (trie 23 s / 16 s, regex 17 s / 17 s). The win is real where the table is large and absent where it is small.

jjang-ai:
### QA — tokenization is byte-identical, proven across the variating turn matrix

This rewrites added-token splitting for **every** family, so the thing that
needed proving is that output does not move. Ran the full variating pattern on
`LFM2.5-VL-3B-JANG_4M` on `vmlx-origin/main` (c59f85ca), then on this branch
merged onto that same main, and diffed all 18 turn/probe lines:

```
diff base.txt pr246.txt  ->  IDENTICAL
```

The matrix that produced those lines, all axes crossed rather than one shape
repeated:

| turn | shape | result |
|---|---|---|
| 1 | reasoning ON, text | `Paris` |
| 2 | **image**, reasoning OFF | `Red to blue gradient.` |
| 3 | **tools offered** | `toolCalls=1` |
| 4 / 4b | **image AND tools, same turn** | no call under either phrasing |
| 5b | **two images, one turn** | answers |
| 6→7 | grow control (constant head) | `HIT disk 22/38 (57% reused)` |
| 5 | **turn 1 replayed verbatim** | `HIT disk 26/26 (100% reused)` |

Same tokens, same tool calls, same cache hits and same reuse percentages on
both builds — including the 100% verbatim-replay hit, which is the sensitive
one here since a tokenization change would move the prefix boundary and break
it.

`AddedTokenTrie source coverage` 2/2 pass, including the longest-token-wins
walk and the unicode-scalar (not grapheme-cluster) traversal.

### What I did NOT measure

I could not reproduce the ~28s-per-token timing locally. The bundles here with
a comparable table are the Audex ones (**74,733** added tokens), and they fail
before tokenization with
`ModelFactoryError.unsupportedModelType("nemotron_dense_audex")` — the known
Audex runtime gap, unrelated to this PR. So the speedup rests on your Apertus
measurement; what I can attest to is that the rewrite is behaviour-preserving
on a real multimodal family across media, tools, reasoning and cache reuse.

jjang-ai:

PR #247, opened 2026-08-12
Load: resolve symlinks in the safetensors truncation guard (intact symlinked bundles rejected as truncated)

rcfa:
### What

The safetensors truncation guard rejects intact bundles whose shards are symlinks, and tells the
user to re-download them.

`safetensorsMissingByteCount` (`Load.swift:87`) stats each shard with
`FileManager.attributesOfItem(atPath:)`, which does not follow symlinks: for a symlink it reports
the length of the target path string. Every shard therefore appears to be missing essentially all of
its bytes.

Observed on a completely intact 4-shard bundle:

```
4 of 4 safetensors shard(s) … are shorter than their own header declares
(16106726362 bytes missing in total) — the download is incomplete or the files were
truncated … Re-download the bundle.
  model-00001-of-00004.safetensors: missing 5259824638 bytes
  model-00002-of-00004.safetensors: missing 5234712897 bytes
```

### Why it matters

The diagnosis and the remedy are both wrong, and the remedy is a multi-GB download that cannot help.
Symlinking shards is a normal thing to do — it is how you share one copy of the weights across
several config variants without duplicating tens of gigabytes on disk. We hit this building an
A/B where two bundles differed only in `config.json`.

### How

Resolve symlinks before stat-ing. `resolvingSymlinksInPath()` is a no-op for regular files, so
genuinely truncated downloads are still caught — that case is why the guard exists and it is
untouched.

### Testing

Verified on device against the failing layout: a bundle whose four shards are symlinks to an intact
model fails to load before the change with the error above, and loads cleanly after it. The
hard-linked equivalent (which stats correctly either way) was unaffected in both builds, as
expected.

### DIFF
--- a/Libraries/MLXLMCommon/Load.swift
+++ b/Libraries/MLXLMCommon/Load.swift
@@ -84,7 +84,13 @@ func safetensorsMissingByteCount(_ url: URL) -> Int64? {
     guard maxEnd > 0 else { return nil }
 
     let declared = Int64(8 + headerLength + maxEnd)
-    guard let size = (try? FileManager.default.attributesOfItem(atPath: url.path))?[.size]
+    // Resolve symlinks before stat-ing. `attributesOfItem(atPath:)` does NOT follow links, so a
+    // symlinked shard reports the length of the TARGET PATH STRING (a few dozen bytes) rather than
+    // the file's real size. The bundle then looks catastrophically truncated and the caller is told
+    // to re-download it — advice that is both wrong and expensive, since symlinking shards is a
+    // normal way to share one set of weights between config variants.
+    guard let size = (try? FileManager.default.attributesOfItem(
+        atPath: url.resolvingSymlinksInPath().path))?[.size]
         as? NSNumber
     else { return nil }
     let actual = size.int64Value


jjang-ai:
Confirmed, and checked against our shipped runtimes — no effect on them, because it can only ever help.

**Premise reproduced.** `attributesOfItem(atPath:)` really does report the target *path string* length for a final-component symlink:

```
real                        : 16
link (unfixed path)         : 111      <- length of the target path string
link.resolvingSymlinksInPath: 16
non-symlink + resolving…    : 16       <- unchanged
```

(16-byte file; the 111 is the absolute target path.) So a multi-GB shard reports a few dozen bytes and `safetensorsMissingByteCount` computes `declared - actual` ≈ the whole file — exactly the `missing 5259824638 bytes` in your report.

**No regression for non-symlinked shards**: the last line above is the case every ordinary bundle takes, and `resolvingSymlinksInPath()` returns it byte-identical.

**Our layout is unaffected either way**, worth stating explicitly since it's why we never hit this: we symlink the *bundle directory*, not the shards —

```
~/.osaurus/models/OsaurusAI/Raptor-1.0-16B-A3B-qat-JANG_4M -> /Users/eric/models/OsaurusAI/Raptor-…
```

`attributesOfItem` follows intermediate path components, so only a symlinked **final** component (the shard file itself) trips the guard. Your case is the shared-weights-between-config-variants layout, which we don't currently use but which is a perfectly normal thing to do.

LGTM.

jjang-ai:
### Reproduced, then verified — with variating-turn QA on the fixed path

**The bug, at the OS level.** Built a bundle whose shards are symlinks (the
exact "share one copy of the weights across config variants" case):

```
model-00001-of-00003.safetensors   lstat = 81 bytes   real = 529,023,344 bytes
model-00002-of-00003.safetensors   lstat = 81 bytes   real = 534,143,256 bytes
```

**Before (main, a983c6fc)** — an intact bundle is rejected and the user is told
to re-download ~2.4 GB that cannot help:

```
[loadWeights] 3 of 3 safetensors shard(s) … are shorter than their own header declares
  model-00001-of-00003.safetensors: missing 529023263 bytes
  model-00002-of-00003.safetensors: missing 534143175 bytes
  model-00003-of-00003.safetensors: missing 1329439879 bytes
Fatal error: MLXLMCommon.TruncatedSafetensorsError
```

The missing-byte figures are the giveaway: they equal the real file sizes minus
the 81-byte symlink path.

**After (this PR)** — loads in 0.39s, and the full variating pattern passes on
that same symlinked bundle. Loader changes touch every family, so the axes are
crossed rather than a single linear multiturn:

| turn | shape | result |
|---|---|---|
| 1 | reasoning ON, text | `Paris` |
| 2 | **image**, reasoning OFF | `Red to blue gradient.` |
| 3 | **tools offered**, no image | `toolCalls=1` |
| 4 | **image AND tools, same turn** | no call — see note |
| 4b | same, direct phrasing | no call |
| 5b | **two images, one turn** | answers |
| 6→7 | grow control (constant head) | `HIT disk 22/38 (57% reused)` |
| 5 | **turn 1 replayed verbatim** | `HIT disk 26/26 (100% reused)` |

`final cache stats: disk{hits=3, misses=74, stores=17}` — reuse works through a
symlinked bundle, and the verbatim replay proves the head boundary survived the
shape changes.

On turn 4: the render probe shows the tools ARE reaching the prompt on image
turns (`image turn: with tools=652 tok, without=602, delta=50`; `text turn:
73 vs 23, delta=50`), so the missing call is this 3B model's behaviour, not a
wiring loss — same as recorded in #250.

Reviewed for runtime impact: the change is confined to `safetensorsMissingByteCount`
and only alters which path is stat-ed. Non-symlinked bundles resolve to
themselves, so their guard behaviour is byte-identical.

PR #248, opened 2026-08-12
Apertus: strip a stray language_model.lm_head. prefix in sanitize() (bundle fails to load outright)

rcfa:
### What

Quantized Apertus checkpoints that carry the output projection under a VLM-style
`language_model.lm_head.*` prefix cannot be loaded at all.

`mlx-community/Apertus-8B-Instruct-2509-Jang_6M` ships 775 tensors, of which exactly three are
`language_model.lm_head.{weight,scales,biases}`, with no un-prefixed twin. The body sits at
`model.*` as normal. Loading fails with:

```
unhandledKeys(modules: ["ApertusModel"], keys: ["language_model"])
```

### How

`sanitize()` absorbs that prefix via the existing shared helper,
`Weights.stripLanguageModelPrefix(_:only: ["lm_head."])` — the same one added for the Llama fix in
#95. Scoped to the head only, so a genuine nested `language_model.*` body is untouched.

### Safety

This is the same class of change as #95 and #155: it fires **only** on checkpoints that currently
throw. No bundle that loads today can take a different path, because the path it changes is one that
ends in an error.

### Testing

Verified on device: the bundle above fails to load before the change with the error quoted, and
loads after it.

Worth noting for context, since it may look odd that an "Apertus Omni" bundle is text-only: that
conversion contains **zero** vision or audio tensors (top-level prefixes are just `lm_head` and
`model`), even though its tokenizer still declares the full 136k-entry visual/audio codebook as
added tokens. The vision tower was not converted. That is a bundle-authoring matter, not something
this patch addresses — but it is why the head shows up under a VLM-style prefix on a model that
loads as a plain causal LM.

### DIFF
--- a/Libraries/MLXLLM/Models/Apertus.swift
+++ b/Libraries/MLXLLM/Models/Apertus.swift
@@ -359,10 +359,20 @@ public class ApertusModel: Module, LLMModel, KVCacheDimensionProvider {
     }
 
     public func sanitize(weights: [String: MLXArray]) -> [String: MLXArray] {
-        // Remove unused precomputed rotary frequencies
-        weights.filter {
-            !$0.key.contains("self_attn.rotary_emb.inv_freq")
-        }
+        // Some quantized checkpoints (e.g. mlx-community / vMLX JANG_6M conversions of text-only
+        // Apertus bodies) emit the output projection under a VLM-style `language_model.lm_head.*`
+        // prefix while the body stays at `model.*`. Observed on
+        // mlx-community/Apertus-8B-Instruct-2509-Jang_6M: 775 tensors, of which exactly 3 are
+        // `language_model.lm_head.{weight,scales,biases}` with NO un-prefixed twin — so the head is
+        // unreachable and loading fails outright with
+        // `unhandledKeys(modules: ["ApertusModel"], keys: ["language_model"])`.
+        //
+        // This model binds the head at the top-level `lm_head`, so absorb that stray prefix rather
+        // than refusing the bundle. Scoped to the head: a genuine nested `language_model.*` body is
+        // not ours to rewrite. Identical treatment to Llama, via the same shared helper.
+        Weights.stripLanguageModelPrefix(weights, only: ["lm_head."])
+            // Remove unused precomputed rotary frequencies
+            .filter { !$0.key.contains("self_attn.rotary_emb.inv_freq") }
     }
 
     public func messageGenerator(tokenizer: any Tokenizer) -> any MessageGenerator {
--- a/Tests/MLXLMTests/ApertusSanitizeLMHeadPrefixTests.swift
+++ b/Tests/MLXLMTests/ApertusSanitizeLMHeadPrefixTests.swift
@@ -1,0 +1,105 @@
+// Copyright 2026 Osaurus AI. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+import Foundation
+import MLX
+import Testing
+
+@testable import MLXLLM
+@testable import MLXLMCommon
+
+/// Covers `ApertusModel.sanitize`'s absorption of a stray `language_model.lm_head.` prefix.
+///
+/// Mirrors `LlamaSanitizeLMHeadPrefixTests` deliberately, case for case. `stripLanguageModelPrefix`
+/// has its own tests, but those pin the HELPER; what a user hits is the CALL SITE, and nothing else
+/// in-tree exercises this wiring — a refactor could drop the call and every other test would still
+/// pass. The `only:` scoping is the part that most needs pinning: widening it to strip a nested
+/// `language_model.*` BODY would silently drop the real weights instead of failing loudly.
+///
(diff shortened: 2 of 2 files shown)


jjang-ai:
No impact on our shipped runtimes, and this is the same shape as #95 which already merged.

It is scoped to `Libraries/MLXLLM/Models/Apertus.swift` — `sanitize()` on one model type — so no other family can observe it, and it only fires on checkpoints that carry `language_model.lm_head.*` with no un-prefixed twin, i.e. bundles that **fail to load at all** today with `unhandledKeys(modules: ["ApertusModel"], keys: ["language_model"])`. A checkpoint that loads now has nothing under that prefix to absorb.

We don't have Apertus locally, so I can't run it — but the structural argument is the same one that made #95 safe, and worth repeating here: the strip must stay scoped to the **head**. A genuinely nested `language_model.*` body belongs to a different layout and has to pass through untouched, or you flatten another model's weights into this one. When I reviewed #95 I added `Tests/MLXLMTests/LlamaSanitizeLMHeadPrefixTests.swift` to pin exactly that boundary at the call site (prefixed head absorbed, quant sidecars travel with it, nested body left alone, rotary `inv_freq` still dropped, unprefixed passthrough unchanged) — because `Weights.stripLanguageModelPrefix` had helper-level tests but nothing exercised the wiring, so a refactor could have dropped the call with every test still green.

If you mirror those five cases for Apertus this becomes airtight; without them the same silent-drop risk exists here. Not blocking.

LGTM.

rcfa:
Added, mirroring `LlamaSanitizeLMHeadPrefixTests` case for case:

1. a `language_model.lm_head.*` head lands on the top-level `lm_head.*`, body untouched
2. quant sidecars (`weight`/`scales`/`biases`) travel with the head — the motivating bundle is quantised, and a head loaded without its scales produces garbage rather than an error
3. **a nested `language_model.*` body is left alone** — the scoping case you flagged
4. precomputed rotary `inv_freq` is still dropped (pre-existing behaviour must survive the new strip)
5. an ordinary unprefixed checkpoint passes through unchanged, so no bundle that loads today can take a different path

Case 3 is the one worth having. Widening the strip beyond the head would rebind real weights onto keys the model does not expect and **silently drop them** — strictly worse than the load failure this patch fixes, because it fails as bad output rather than as an error.

**One caveat I should state plainly: I could not execute these locally.** Any MLXArray-touching test in a vmlx clone dies here with `Failed to load the default metallib`, and I confirmed it is environmental rather than mine by running your own `ZayaRMSNormTests` as a control — it fails identically. The new tests compile and are structurally identical to the merged #95 ones, but "compiles and mirrors a passing suite" is weaker than "I watched them pass", and I would rather say so than imply otherwise. Please confirm in your CI.

On not having Apertus locally: the bundle that motivates this is `mlx-community/Apertus-8B-Instruct-2509-Jang_6M`. Also worth flagging, since it may affect how you weigh this one — that conversion is **text-only**: it carries none of the vision or audio tensors the real Apertus 1.5 ships (`model-vision_tokenizer-model.safetensors`, `model-wavtokenizer-model.safetensors`), because the converter took only the four `model-apertus-model-*` shards. So this patch makes a text-only Apertus loadable; full multimodal support needs an `apertus1p5` implementation, which does not exist in vmlx yet.

jjang-ai:
### QA — Apertus-scoped, and proven not to disturb the families we ship

**Blast radius.** The diff touches `Libraries/MLXLLM/Models/Apertus.swift` and
one new test file. `Weights.stripLanguageModelPrefix` already exists on main, so
no shared helper changes shape — Llama's use of it is untouched.

**The PR's own contracts — 5/5:**

```
✔ a `language_model.lm_head.*` head lands on the top-level `lm_head.*`
✔ unprefixed weights pass through unchanged
✔ a nested language_model body is left alone
✔ precomputed rotary frequencies are still dropped
```

The third is the one that matters for scope: a genuine nested `language_model.*`
body is NOT rewritten, so this cannot silently capture a real VLM layout.

**No regression on the families we actually run.** Apertus is not in the local
library, so the risk to check is collateral damage. Full variating pattern on
`LFM2.5-VL-3B-JANG_4M`, this branch merged onto main, diffed against main:

```
diff base.txt pr248.txt  ->  IDENTICAL   (all 18 turn/probe lines)
```

That run crosses reasoning ON/OFF, text↔image↔text-after-image, tools
constant-head grow control (`HIT disk 22/38, 57% reused`) and a verbatim replay
of turn 1 (`HIT disk 26/26, 100% reused`).

Second family, different architecture (`Qwen35` hybrid SSM, Bonsai-27b-Ternary):

```
Coordinator probe before turn 2: HIT tier=paged matched=18/53 remaining=35
=== BENCH_GROWING_CHAT_CACHE: passed ===
```

Unchanged from main.

**Not verified:** the Apertus bundle itself
(`mlx-community/Apertus-8B-Instruct-2509-Jang_6M`) is not present here, so the
775-tensor / 3-stray-key load is on your report. What I can attest to is that
the sanitize change is scoped, its negative case is covered, and it leaves the
shipped families byte-identical.

jjang-ai:

PR #250, opened 2026-08-13
LFM2.5-VL: fix image processing (crash) and tool calling (silently lost) on all four quants

jjang-ai:
Two independent defects that made LFM2.5-VL-3B unusable for its two headline features. Batched because they were found in the same live sweep and share the same proof runs.

---

## 1. Every image request crashed

```
MLXVLM/LFM2VL.swift:971: Fatal error: Image features and image tokens do not match: tokens: 1, features 256
```

`LFM2VLProcessor` scanned the prompt for image placeholder id **`396`** and expanded each hit into one token per vision feature. `396` is only correct for LFM2-VL-1.6B; LFM2.5-VL-3B declares `image_token_id: 124907`. The scan matched nothing, no expansion happened, and the prompt reached the model with one image token against 256 features. The model side was already right — `imageTokenIndex` reads `image_token_id` from config — so only the processor disagreed with the bundle.

**The mirror hazard is worse than the crash.** In the 3B vocabulary, id `396` is the ordinary text token `"ab"`:

```
id     396 -> 'ab'
id  124907 -> '<image>'
```

On any bundle where the hard-coded id is real text, every prompt containing that subword would have had it silently expanded into image tokens.

Fixed by resolving `<image>` through the tokenizer — the same placeholder the chat template emits — round-tripped, because `convertTokenToId` answers with the *unknown-token* id rather than nil for a missing token, so a bare `!= nil` is not an existence test. Falls back to `396` for the 1.6B bundles.

### Proof — `BENCH_VL_CHAT_CACHE`, all four quants (before: fatal error on every one)

| quant | image description (cold) | same-media | different-media | replay TTFT |
|---|---|---|---|---|
| JANG_2L | "…deep red at the top to bright blue at the bottom" | HIT disk 286/286 | MISS (correct) | 387 → 94 ms |
| JANG_4M | "…bright orange at the top to deep blue at the bottom" | HIT disk 286/286 | MISS (correct) | 674 → 92 ms |
| JANG_6M | "…bright orange at the top to deep blue at the bottom" | HIT disk 286/286 | MISS (correct) | 103 → 95 ms |
| MXFP8 | "…bright red at the top to deep blue at the bottom" | HIT disk 286/286 | MISS (correct) | 788 → 97 ms |

`A tokens` goes 31 → **286**. Beyond "it no longer crashes", that row also proves **media-salted isolation** (a second image cannot restore the first one's cache) and **cross-turn reuse over a media turn** (the text follow-up answers from the image at ~114 ms TTFT).

---

## 2. Tool calls were silently lost — `toolCalls=0`, call leaked as prose

All four quants, on two different bench rows. Two independent causes, both required:

**(a) `LFM2VLProcessor` never set `LMInput.toolSchemas`.** Callers pass tools through `UserInput` → processor, and BatchEngine reads them back off `LMInput`. With none, it builds a strip-only `ToolCallProcessor`, and the inline fallbacks then match against a hard-coded default name list instead of the offered tools. Tagged formats hide this because their envelope is self-identifying — which is exactly why Ornith (`xmlFunction`) passed the same bench row while lfm2 did not. Peers already pass it (`Gemma4.swift:1676`, `Qwen3VL.swift:156/239`).

With (a) alone **JANG_2L started passing**, because it happens to emit `{"name": …, "arguments": …}`, which was already recovered.

**(b) The other three emit the same call as `{"function": "get_weather", "parameters": {…}}`.** `parseNamedArgumentsJSONEnvelope` accepted only `name`/`arguments`, and `looksLikeExplicitInlineToolIntent` only recognised `"function"` when it carried a *nested object*, so the buffer was never treated as tool intent at all. Now `function` is accepted as a name when it is a string (the nested `{"function":{"name":…}}` shape is a different envelope, handled elsewhere) and `parameters` as an arguments alias.

Every downstream guard is untouched — the name must still be an offered tool, required parameters must all be present, unknown keys are still rejected — so the alias cannot admit a call the strict spelling would have refused.

### Proof — `BENCH_BATCH_TOOLCALL` (before: all four FAIL, "generation completed without a structured .toolCall event")

```
JANG_2L  exit=0  tool calls: get_weather({"location":"Tokyo"})
JANG_4M  exit=0  tool calls: get_weather({"location":"Tokyo"})
JANG_6M  exit=0  tool calls: get_weather({"location":"Tokyo"})
MXFP8    exit=0  tool calls: get_weather({"location":"Tokyo"})
```

`chunks: 0 chars` — nothing leaks to the user.

---

## No regressions

- **Ornith-1.0-35B (`xmlFunction`)** still passes the same tool row.
- **The VL image row** still passes: `HIT disk 286/286` / `MISS (correct)`.
- **95 tests across 10 tool-parser suites** green.
- `LFM2EnvelopelessToolCallTests` — 10 cases, new — covers every shape these bundles were observed to emit (pythonic single/double-quoted, both JSON spellings) in one chunk, six chunks, and one character at a time, and pins that prose, unoffered names and unterminated calls stay prose. Sabotage-verified: dropping the `function` alias turns the new case red with `toolCalls → 0`.

Worth recording for the next reader: the parsing layer was never the problem for the pythonic shape — it already handled it. The failure was upstream, in what the generation path handed the processor. Those characterization tests exist so nobody re-derives that from the parser.

### DIFF
--- a/Libraries/MLXLMCommon/Tool/Parsers/PythonicToolCallParser.swift
+++ b/Libraries/MLXLMCommon/Tool/Parsers/PythonicToolCallParser.swift
@@ -853,9 +853,25 @@ public struct PythonicToolCallParser: ToolCallParser, Sendable {
         guard text.hasPrefix("{") else { return nil }
         guard let object = parseJSONObjectWithOptionalEOFBrace(text) else { return nil }
         guard object.count == 2 || object.count == 3 else { return nil }
-        guard let name = object["name"] as? String,
+        // `name`/`arguments` is the common spelling, but the same bundle family
+        // will also emit `function`/`parameters` for the identical call —
+        // measured on LFM2.5-VL-3B, where JANG_2L produced
+        // `{"name": "get_weather", "arguments": {…}}` (recovered) while
+        // JANG_4M, JANG_6M and MXFP8 produced
+        // `{"function": "get_weather", "parameters": {…}}` (not recovered, so
+        // the call leaked to the user as prose with `toolCalls=0`).
+        //
+        // `function` is only a NAME here when it is a string; the nested
+        // `{"function": {"name": …}}` envelope is a different shape handled
+        // elsewhere, and reading it as a name would be wrong.
+        //
+        // Every downstream guard is unchanged — the name must still be an
+        // offered tool, required parameters must all be present, and unknown
+        // keys are still rejected — so accepting the alias cannot admit a call
+        // the strict spelling would have refused.
+        guard let name = (object["name"] as? String) ?? (object["function"] as? String),
             toolNames(tools: tools).contains(name),
-            let rawArguments = object["arguments"],
+            let rawArguments = object["arguments"] ?? object["parameters"],
             let args = firstArgumentObject(from: rawArguments)
         else { return nil }
 
--- a/Libraries/MLXLMCommon/Tool/ToolCallProcessor.swift
+++ b/Libraries/MLXLMCommon/Tool/ToolCallProcessor.swift
@@ -787,6 +787,12 @@ public class ToolCallProcessor {
             return inlineFunctionToolNames().contains { name in
                 compact.contains(#""tool":"\#(name)""#)
                     || compact.contains(#""tool_name":"\#(name)""#)
+                    // `"function"` carrying the NAME directly, rather than a
+                    // nested object. LFM2.5-VL-3B JANG_4M/JANG_6M/MXFP8 emit
+                    // `{"function": "get_weather", "parameters": {…}}`; without
+                    // this the buffer is never treated as tool intent and the
+                    // call leaks to the user as prose.
+                    || compact.contains(#""function":"\#(name)""#)
                     || compact.contains(#""function":{"name":"\#(name)""#)
                     || compact.contains(#""function":{"#)
                         && compact.contains(#""name":"\#(name)""#)
--- a/Libraries/MLXVLM/Models/LFM2VL.swift
+++ b/Libraries/MLXVLM/Models/LFM2VL.swift
@@ -675,6 +675,25 @@ public struct LFM2VLProcessor: UserInputProcessor {
         self.tokenizer = tokenizer
     }
 
+    /// The `<image>` placeholder id for THIS bundle's vocabulary.
+    ///
+    /// LFM2-VL-1.6B uses 396; LFM2.5-VL-3B uses 124907, and in the 3B vocabulary
+    /// 396 is the ordinary text token `"ab"`. Resolving from the tokenizer keeps
+    /// the processor's id and the model's `imageTokenIndex` (which reads
+    /// `image_token_id` from `config.json`) in agreement on any bundle.
+    ///
+    /// `convertTokenToId` returns the *unknown-token* id rather than nil for a
+    /// token the vocabulary lacks, so the round-trip through `convertIdToToken`
+    /// is what makes this an existence test rather than a lookup that always
+    /// "succeeds".
+    static func resolveImageTokenId(tokenizer: any Tokenizer) -> Int {
+        let legacyLFM2VL16B = 396
+        guard let candidate = tokenizer.convertTokenToId("<image>"),
+            tokenizer.convertIdToToken(candidate) == "<image>"
+        else { return legacyLFM2VL16B }
+        return candidate
+    }
+
     /// Preprocess a single image
     func preprocess(image: CIImage, targetSize: CGSize) -> CIImage {
         image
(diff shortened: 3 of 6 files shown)


jjang-ai:
Follow-up: ran the full variating pattern (`BENCH_VL_VARIATING=1`, added in this branch) across all four quants, which is what the two fixes above needed to be checked against — reuse, tools and media interact, and a linear row exercises none of that.

**One conversation, cache probed before every turn** (identical shape on JANG_2L / JANG_4M / JANG_6M / MXFP8):

```
[1 think+text]    MISS  26 tok           -> "Paris"
[2 image+nothink] MISS 300 tok  media    -> "Red to blue gradient."
[3 tools+nothink] MISS 376 tok  media    -> toolCalls=1 get_weather(location:Tokyo)
[4 image+tools]   MISS 661 tok  media    -> toolCalls=0
[5 replay turn1]  HIT disk 26/26 (100%)  -> "Paris"
disk{hits=1,misses=35,stores=9}
```

Turn 3 is the one that matters for the tool fix: a tool call succeeds with an image already in history, which the linear `BENCH_BATCH_TOOLCALL` row cannot show. Turn 5 shows the head boundary survives every shape change above it.

**Turn 4 — image and tool request in the SAME turn — produces no call on any quant**, three of them answering "I do not have access to a tool" while a tool is offered. That is NOT one of these fixes regressing, and I measured rather than assumed which of the two possible causes it is:

```
[render probe] image turn: with tools=652  without tools=602  delta=50
[render probe] text turn:  with tools=73   without tools=23   delta=50
```

Identical 50-token tool block, so the tools reach the prompt on a media turn exactly as on a text turn. With 256 image tokens in the same turn this 3B simply stops using them — model behaviour, reproducing on MXFP8 too, so not quantization damage. Rephrasing as a single direct instruction (turn 4b) does not help either.

The probe stays in the row and fails it outright if a future change ever does drop tools on media turns while keeping them on text turns — the regression this measurement rules out today.

Second observation worth flagging separately from this PR: the growing conversation never hits the cache (`hits=1, misses=35`) — only the verbatim replay does. That is this family's multi-turn prefill tax, and it is the same class of problem #249 addresses for the turn-start boundary.

jjang-ai:
**Correction to my previous comment.** I said the growing conversation "never hits the cache (`hits=1, misses=35`) — this family's multi-turn prefill tax". That attribution was wrong, and the fix is a control I should have run before claiming it.

The trace shows every probe returning **`noRow`** — content divergence, not a missing boundary — and the stored boundaries were present all along, in pairs:

```
stored: 26 + 28    (turn 1: strip boundary + full prompt)
        300 + 306  (turn 2)
        376, 661, 685, 652, 670
turn 3 probes 375, 306, 300, 28, 26 -> all noRow
```

The cause is structural, not a defect: this template renders **both the tool list and the thinking scaffold into the system prompt, at the head**. Every turn in the variating pattern changes one of those, which rewrites the prefix the cache is keyed on. Invalidating there is exactly what a prefix cache should do.

Added turns 6+7 as the control — identical shape, growing only by the appended turn:

```
[6 grow A]              MISS 22 tok
[7 grow B (same shape)] HIT disk 22/38 (57% reused)
grow-control reused the prefix (disk hits 0 -> 2)
```

So prefix reuse on this family works. The row now **fails** if that control ever misses, which is the case that would genuinely be a bug — and varying the axes without it cannot distinguish the two, which is precisely the mistake I made.

Also measured against #249 (turn-start boundary widening) merged in: stores go **9 → 11** on this run, i.e. the extra turn-start boundaries are captured for this family as intended. Hits are unchanged here only because every axis-changing turn rewrites the head, so there is nothing for the extra boundary to match — its value shows up in same-shape multiturn, which is what the control exercises.

jjang-ai:
### Final live proof — multi-image turn in the dev-built app

The multi-image crash fix is verified against the exact path that used to trap.
Dev-built Release app, `osaurus` repinned to `fd259ec7`, model
`JANGQ-AI/LFM2.5-VL-3B-JANG_4M`, **three** images attached in one turn (one more
than the reported repro):

```
prompt : "What shape and colour is on the left of these images?"
reply  : "The left shape is a red circle, and the right shape is a blue rectangle."
tokens : 17     ttft: 1.02s     app: alive
```

Before the fix the same shape aborted in
`LFM2VL.mergeInputIdsWithImageFeatures` — the crash report from the pre-fix
binary, for the record:

```
EXC_BREAKPOINT (SIGTRAP)  thread com.apple.root.user-initiated-qos.cooperative
  libswiftCore  _assertionFailure(_:_:file:line:flags:)
  osaurus       LFM2VL.mergeInputIdsWithImageFeatures(imageFeatures:inputsEmbeds:inputIds:imageTokenIndex:)
  osaurus       LFM2VL.getInputEmbeddings(inputIds:pixelValues:spatialShapes:pixelAttentionMask:)
  osaurus       LFM2VL.prepare(_:cache:windowSize:)
  osaurus       TokenIterator.prepareRemainder(input:windowSize:promptTokensForProcessor:)
```

The assertion was a shape mismatch: the placeholder run was expanded once for
the whole turn rather than once per `<image>` placeholder, so with N>1 images the
embedding row count never matched the token count. Expansion is now per
placeholder, indexed by `imageIdx` into `allSpatialShapes`.

### Test gate

CI on this fork is upstream-gated (`if: github.repository == 'ml-explore/mlx-swift'`),
so every job reports `skipping` by design — local runs are the gate:

```
83 tests in 7 suites passed   (LFM2 envelopeless pythonic tool call,
LFM and DSML nested Pythonic tool arguments, DSML Tool-Call Parser (DSV4),
Chat.Message tool-call plumbing, Hy3 routing, Mistral bare-JSON-array
fallback, Raptor tool continuation)
```

PR #252, opened 2026-08-13
Add an opt-in reasoning ceiling — DSV4 ships no budget mechanism to port

jjang-ai:
DSV4 Flash on the **Low** rail produced a 38,494-character think block over
10,554 tokens, hit `stop=length`, and returned 227 characters of answer.
Reproduced live in the dev app from the reported prompt shape — the same
10,554 token count as the report.

## Sampling was not the cause

Traced out of the running app (`OSAURUS_SAMPLING_TRACE=1`), the real turn ran
at the bundle's own contract:

```
sampling-resolve   bundle{temp=0.6 topP=0.95 topK=0 doSample=true}
sampling-effective temp=0.6 topP=0.95 topK=0 minP=0.0 maxTokens=16384
```

`top_p 0.95` reaches the sampler intact from both `generation_config.json` and
`jang_config.json > chat > sampling_defaults`. `server.json`'s `genTopP: 1`
projects to nil because it equals the shipped default, so it never overrides
the model.

Three plausible causes were ruled out by measurement, not reading: vmlx's
engine default of `topP = 1.0` (never reached); a `LocalGenerationDefaults`
id-lookup returning nil (an artifact — the test process has
`installedModelNames().count == 0`); and `MinimumReasoningFloor` (a 16-token
delay that can only postpone a close).

## There is nothing upstream to port

DeepSeek's own `encoding/encoding_dsv4.py` defines the entire reasoning
control as three prose strings injected once at `index == 0`:

```python
"low":  ""      # empty — nothing is injected at all
"high": "Reasoning Effort: Absolute maximum with no shortcuts permitted.…"
"max":  "Reasoning Effort: Beyond maximum — exhaustive, relentless…"
```

No budget token, no counter, no cap — only `<think>`/`</think>`. That is what
makes DSV4 different from families that bound reasoning structurally
(`enable_thinking`, harmony channels). A ceiling has to be ours.

## The ceiling

`ReasoningBudget` is **off unless asked**:

- Arms only when `VMLX_REASONING_BUDGET` names a positive token count.
- While budget remains it is a pass-through — logits are returned unchanged,
  so there is no nudge toward closing.
- At the ceiling it requires the close token once, then bans reopening for the
  rest of the generation so the ceiling bounds the turn rather than one block.
- Open/close tags are resolved per family by round-tripping candidate
  spellings through the tokenizer (`convertTokenToId` returns the unk id, so
  id existence alone proves nothing). Covers both template-primed reasoning
  (DSV4 renders `…assistant<think>`) and families where the model opens the
  block itself.
- Native MTP is refused while a budget is active: a drafted token could sail
  past the ceiling unchecked.

This is deliberately **not** the mechanism `NoHiddenReasoningCloseBiasFocusedTests`
bans. That one armed itself automatically with nothing surfaced. This one
cannot arm without an explicit opt-in, and that test is extended here to
enforce it: the parameters must default to nil, the engine must arm only
through `armIfNeeded`, the env gate must exist, and no self-arming variant may
be declared.

## Live proof

`VibeThinker-3B-MXFP8` — a family where the **model** emits the open tag, so
this exercises the deferred-start path. Close position tracks the budget
exactly, and is deterministic (same budget twice is byte-identical):

| budget | `</think>` at char | think content |
|---|---|---|
| 4 | 22 | `>The user says:` |
| 8 | 42 | `>The user says: "Reply with exactly` |
| 16 | 76 | `>The user says: "Reply with exactly this phrase and nothing else: vml` |
| 32 | >120 | still open |

```
[vmlx][reasoning-budget] reasoning opened by model; counting
[vmlx][reasoning-budget] requiring close id=151666
[vmlx][reasoning-budget] close required; budget disarmed
```

Two bugs were caught by that measurement and are fixed here: the forced close
never won, because the mask was built by writing a column into a `full(-inf)`
array and `MLXArray` is a class whose subscript assignment aliases; and
nothing stopped the model reopening the block immediately afterwards.

**Measuring note:** the bench's `raw=… chars` counts the WHOLE turn. A tighter
budget shortens thinking and the model then writes a longer visible answer, so
character totals move the wrong way and the ceiling looks inverted. Measure
the `</think>` offset.

## Tests

`ReasoningBudgetTests` 9/9 — inertness without a budget, without a close
token, and at zero; pass-through while budget remains; the close required
exactly once; disarm afterwards so the answer samples normally; an
out-of-range close id ignored rather than blanking the distribution; arming
conditions; and the MTP refusal.

The 3 failures in `NoHiddenReasoningCloseBias` are pre-existing and belong to
"Nemotron Omni image preprocessing" (`NemotronHOmni.swift` /
`Preprocessors.swift`), untouched here.

## Still open, deliberately not in this PR

DeepSeek recommends up to **384K** output tokens for high/max effort; we
resolve `maxTokens=16384`, so a long reason legitimately hits `stop=length`
with no answer. That needs a per-effort max-token policy on the osaurus side.

### DIFF
--- a/Libraries/MLXLMCommon/BatchEngine/BatchEngine.swift
+++ b/Libraries/MLXLMCommon/BatchEngine/BatchEngine.swift
@@ -597,6 +597,18 @@ public actor BatchEngine {
             parameters.initialSuppressTokens = [floor.closeTokenID]
             parameters.initialSuppressCount = floor.tokenCount
         }
+        // `generate` is the path the app's single-request chat actually takes
+        // (via `startSoloFastPath`), so arming only in `submit` left the
+        // ceiling inert for every GUI turn — verified live: env set on the
+        // process, sampling trace writing, and zero budget trace lines.
+        if let budget = ReasoningBudget.armIfNeeded(
+            tokenizer: tokenizer, promptTail: promptTail)
+        {
+            parameters.reasoningBudgetTokens = budget.tokenCount
+            parameters.reasoningBudgetCloseTokenID = budget.closeTokenID
+            parameters.reasoningBudgetStartTokenIDs = budget.startTokenIDs
+            parameters.reasoningBudgetOpenTokenIDs = budget.openTokenIDs
+        }
         if parameters.draftStrategy?.usesNativeMTP == true {
             guard canStartExclusiveSoloPath else {
                 Self.logger.error(
--- a/Libraries/MLXLMCommon/BatchEngine/BatchEngine.swift
+++ b/Libraries/MLXLMCommon/BatchEngine/BatchEngine.swift
@@ -474,6 +474,15 @@ public actor BatchEngine {
             parameters.initialSuppressTokens = [floor.closeTokenID]
             parameters.initialSuppressCount = floor.tokenCount
         }
+        // Upper bound. Inert unless VMLX_REASONING_BUDGET names a token count,
+        // so this changes nothing for callers who do not ask for it.
+        if let budget = ReasoningBudget.armIfNeeded(
+            tokenizer: context.tokenizer, promptTail: promptTail)
+        {
+            parameters.reasoningBudgetTokens = budget.tokenCount
+            parameters.reasoningBudgetCloseTokenID = budget.closeTokenID
+            parameters.reasoningBudgetStartTokenIDs = budget.startTokenIDs
+        }
         let request = BatchPendingRequest(
             input: input,
             parameters: parameters,
--- a/Libraries/MLXLMCommon/BatchEngine/BatchEngine.swift
+++ b/Libraries/MLXLMCommon/BatchEngine/BatchEngine.swift
@@ -482,6 +482,7 @@ public actor BatchEngine {
             parameters.reasoningBudgetTokens = budget.tokenCount
             parameters.reasoningBudgetCloseTokenID = budget.closeTokenID
             parameters.reasoningBudgetStartTokenIDs = budget.startTokenIDs
+            parameters.reasoningBudgetOpenTokenIDs = budget.openTokenIDs
         }
         let request = BatchPendingRequest(
             input: input,
--- a/Libraries/MLXLMCommon/Evaluate.swift
+++ b/Libraries/MLXLMCommon/Evaluate.swift
@@ -231,6 +231,9 @@ public struct GenerateParameters: Sendable {
     /// the open tag. Empty means the prompt already opened reasoning.
     public var reasoningBudgetStartTokenIDs: [Int] = []
 
+    /// All open-tag ids, banned after the ceiling so reasoning cannot reopen.
+    public var reasoningBudgetOpenTokenIDs: [Int] = []
+
     /// Speculative-decoding strategy (opt-in). `nil` preserves the existing
     /// autoregressive decode path byte-for-byte — callers who don't set this
     /// see no behaviour change.
@@ -426,7 +429,8 @@ public struct GenerateParameters: Sendable {
             else { return nil }
             return ReasoningBudgetProcessor(
                 closeTokenID: closeID, tokenCount: budget,
-                startTokenIDs: reasoningBudgetStartTokenIDs)
+                startTokenIDs: reasoningBudgetStartTokenIDs,
+                openTokenIDs: reasoningBudgetOpenTokenIDs)
         }()
 
         if repetitionContext == nil && presenceContext == nil && frequencyContext == nil
--- a/Libraries/MLXLMCommon/Evaluate.swift
+++ b/Libraries/MLXLMCommon/Evaluate.swift
@@ -217,6 +217,20 @@ public struct GenerateParameters: Sendable {
     /// stay masked. Zero disables the window.
     public var initialSuppressCount: Int = 0
 
+    /// Explicit ceiling on reasoning-block length, in sampled tokens. `nil`
+    /// (the default everywhere) leaves generation completely untouched.
+    /// Paired with ``reasoningBudgetCloseTokenID``; see `ReasoningBudget`,
+    /// which also documents why this is not the banned automatic close bias.
+    public var reasoningBudgetTokens: Int? = nil
+
+    /// Close token required once ``reasoningBudgetTokens`` is spent. Resolved
+    /// per family by round-tripping candidate spellings through the tokenizer.
+    public var reasoningBudgetCloseTokenID: Int? = nil
+
+    /// Ids that start the count for families whose model, not template, emits
+    /// the open tag. Empty means the prompt already opened reasoning.
+    public var reasoningBudgetStartTokenIDs: [Int] = []
+
     /// Speculative-decoding strategy (opt-in). `nil` preserves the existing
     /// autoregressive decode path byte-for-byte — callers who don't set this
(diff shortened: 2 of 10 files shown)


jjang-ai:
### Live proof on the reported model — DSV4 Flash in the dev app

Same prompt, same `· Low` rail, same dev-built app as the original report.

```
[vmlx][reasoning-budget] requiring close id=128822
[vmlx][reasoning-budget] close required; budget disarmed
```

No "reasoning opened by model" line, correctly: DSV4's template primes
`…assistant<think>`, so the counter starts immediately. That is the
template-primed path, complementing the VibeThinker run in the PR body which
covers the self-opening path.

| same prompt, Low rail | before | `VMLX_REASONING_BUDGET=512` |
|---|---|---|
| thinking | 38,494 chars | **2,315 chars** |
| answer | 227 chars | **40,579 chars** |
| tokens | 10,554 | 10,554 |

The token budget is identical in both runs — the entire difference is where it
went. 2,315 chars is ~550 tokens, consistent with the 512 ceiling. The turn
now returns a complete HTML file instead of 227 characters of nothing.

### One more bug this proof caught

`86426a25`. `BatchEngine` arms the minimum-reasoning floor at **two** sites and
the ceiling had only been added to `submit`. The app's single-request chat goes
through `generate` → `startSoloFastPath`, so every GUI turn ran with the
ceiling inert.

That was found by measurement, not review: with the env var confirmed on the
process (`ps eww`) and the sampling file-trace writing normally from the same
run, the budget trace produced **zero** lines. A DSV4 turn that looked improved
(38,494 → 5,298 chars) in that run was therefore ordinary variance, not the
ceiling — it had never armed. Worth stating plainly, because a passing-looking
number nearly got credited to a feature that was not running.

jjang-ai:
### Variating-turn matrix on DSV4 Flash, live in the dev app

One conversation, crossing the axes rather than repeating a shape. Ceiling =
256 unless stated.

| turn | shape | think | answer | stop | ceiling fired |
|---|---|---|---|---|---|
| T1 | long-reasoning prompt | 1,265 | 8,509 | stop | yes (trace 0→2) |
| T2 | **T1 replayed verbatim** | 1,244 | 8,644 | stop | yes (trace 2→4) |
| T3 | **tools flipped none→offered** mid-conversation | 1,142 | 414 | stop | yes (trace 4→6) |
| — | **control: same T3 prompt, ceiling OFF** | 2,203 | 739 | — | n/a |

**Bounding is monotonic on DSV4 itself**, not just the harness model:

| ceiling | think | answer | stop |
|---|---|---|---|
| none | 38,494 | 227 | length |
| 512 | 2,315 | 40,579 | length |
| 256 | 1,265 | 8,509 | stop |

Halving the budget halved the thinking, and the turn stops cleanly instead of
being truncated at the token cap.

**The ceiling does not break tool calling.** T3 recorded no tool call, so I ran
the control rather than assume: with the ceiling OFF the same prompt also
produced no tool call. That is model behaviour on this prompt, and the ceiling
is exonerated. The control also reproduces the halving (2,203 → 1,142) on an
identical prompt.

**Also crossed:** reasoning effort flipped mid-conversation (chip verified
`· Low` → `· High`), and the ceiling correctly stayed **inert** on short
reasons — a 75-character think left the trace count unchanged.

**Not covered, stated rather than implied:** media axis (DSV4 is text-only, and
the image/tools combination is covered for LFM2.5-VL in #250), and per-turn
`matched/total` cache probes, which the GUI does not expose — the harness rows
cover reuse separately.

jjang-ai:
### Cache-reuse parity — the ceiling does not disturb the prefix cache

The ceiling changes which tokens are generated, which changes what gets stored,
so reuse had to be measured rather than assumed. `BENCH_GROWING_CHAT_CACHE=1
BENCH_GROWING_THINK=1` on `Raptor-1.0-16B-A3B-qat-JANG_4M`, probed before the
growing turn:

```
ceiling OFF : HIT tier=paged matched=56/84 remaining=28
              paged{hits=3→5, misses=3}  disk{stores=4→7, misses=12}
ceiling 24  : HIT tier=paged matched=56/84 remaining=28
              paged{hits=3→5, misses=3}  disk{stores=4→7, misses=12}
```

Identical tier, identical `matched/total`, identical store and hit counts, and
both arms pass the row's correctness gate.

Noted for accuracy: `VibeThinker-3B` cannot serve as the cache subject here —
it fails the row's recall gate in BOTH arms by transcribing `vmlx-cache-green`
as `vml-cache-green`, so no probe is reached. That is model transcription, not
the ceiling, and it is why Raptor is used for this measurement.

With this the axis coverage for the change is:

- reasoning depth — monotonic on DSV4 (none 38,494 / 512 2,315 / 256 1,265 chars)
- reasoning effort flipped mid-conversation — `· Low` → `· High`, chip verified
- tool availability flipped mid-conversation — none → offered, with an
  explicit ceiling-off control proving no tool-calling regression
- earlier turn replayed verbatim — ceiling fires per turn, think 1,265 → 1,244
- cache probed with `matched/total` + tier — identical, above
- both arming paths — template-primed (DSV4) and model-opened (VibeThinker)
- inert when not needed — a 75-char think leaves the trace count unchanged

Media axis does not apply to DSV4 (text-only); image + tools in one turn is
covered for LFM2.5-VL in #250.

jjang-ai:
### Also in this branch: a killed test run no longer wedges the machine

Found while attributing failures for the work above, and worth calling out
separately because it is not specific to this change.

The MLX test lock is a **named** POSIX semaphore (`/vmlx_mlx_lock`), so its
count outlives the process holding it. A run killed mid-section — Ctrl-C, a CI
timeout, a crash — leaves it at zero **permanently**, and every later
`swift test` on that machine blocks forever at 0% CPU with no output, including
runs touching only unrelated tests. It also holds the SwiftPM `.build` lock
while stuck, so concurrent builds queue behind it and look hung too.

The symptom is pure silence, which is why it took hours to identify. The tell
is *started but never finished*: diff `◇ Test "X" started` against
`✔/✘ Test "X"` in the log. Here that pointed at two Hy3 sanitizer tests — which
are innocent. They are simply the first to take the lock, and they pass in
0.25s once the semaphore is cleared.

`wait()` now polls `sem_trywait` and, after 90s with no release, unlinks the
abandoned semaphore and reopens. Unlinking only detaches the name, so a
genuinely live holder keeps its handle and its `signal()` stays valid; a false
positive costs two runs briefly sharing MLX, not a corrupted lock.

Applied to **both** copies — `ProcessWideMLXTestSemaphore` is duplicated in
`MLXLMTests` and `MLXLMCommonFocusedTests` against the same semaphore name.
Patching one leaves the other wedging, which is what happened on the first
attempt.

Verified by reproducing the failure deliberately (a helper takes the lock and
exits without signalling):

```
[vmlx-tests] MLX lock /vmlx_mlx_lock held for >90s; assuming a killed run
             abandoned it and resetting.
✔ Test "Hy3 sanitizer fuses qkv and drops preserved nextn tensors" passed after 90.353 seconds
```

Before the fix, that same wedge left the process hanging 27 minutes with no
output.

Happy path unaffected — unwedged, the two tests pass in 0.080s and 0.120s, and
a wider slice sharing the lock (ReasoningBudget + CacheCoordinatorMediaSalt +
EmptyArrayData) runs 25 passed / 0 failed.

PR #254, opened 2026-08-13
Stop treating harmony's <|end|> channel separator as a generation stop

jjang-ai:
## Problem

`resolveStopSequences` augments every model's stop set with `commonEndTokenStrings`. That list
carries `<|end|>` for **Phi-3 / Phi-4**, where it really is the turn end.

In the **OpenAI harmony** format (gpt-oss) `<|end|>` means the opposite: the model closes its
`analysis` channel with `<|end|>` and then **opens `final`**. Treating it as a stop halts
generation at the end of the reasoning channel, so the visible answer is empty.

gpt-oss says so itself — `generation_config.json` declares `eos_token_id: [200002 <|return|>,
199999 <|endoftext|>, 200012 <|call|>]` and deliberately omits `200007 <|end|>`.

Diagnosis is from #91 (@rcfa). This PR takes that diagnosis with a narrower gate — see below.

## Why not #91's gate

#91 skips the blanket for **any** bundle that declares an eos in `generation_config.json`. That
is too broad: declaring-but-incomplete bundles are common, and for them the blanket is the only
turn boundary.

Scanned the local library — 86 bundles with a `tokenizer_config.json`, resolving each
`commonEndTokenStrings` entry against declared eos ∪ tokenizer eos:

| bundle | declares | tokenizer eos | template ends turns with |
|---|---|---|---|
| `VibeThinker-3B-MXFP8` | `<\|endoftext\|>` | `<\|endoftext\|>` | **`<\|im_end\|>`** |
| `VibeThinker-3B-JANG_4M` | `<\|endoftext\|>` | `<\|endoftext\|>` | **`<\|im_end\|>`** |

Under #91 those lose `<|im_end|>` entirely and every turn runs to the token cap. That is not a
review opinion — `UnderDeclaredEOSStopTests` (already on main) fails against #91:

```
✘ underDeclaredEOSStillStopsAtImEnd() — Expectation failed:
  (resolved.tokenIDs → [151643]).contains(Self.imEnd → 151645)
```

A template-terminator probe does not separate the two cases either: harmony renders **historical**
assistant turns ending in `<|end|>` too, and only *generation* ends at `<|return|>`.

## Fix

Suppress `<|end|>` only on **two** independent pieces of positive evidence:

1. the vocabulary is harmony-shaped — carries `<|channel|>`, `<|message|>` **and** `<|return|>`; and
2. the model's own `generation_config.json` declaration excludes `<|end|>`.

Anything less keeps today's behaviour exactly. The format is detected from the **vocabulary**
rather than `reasoningParserName`, because that stamp is optional — a locally converted gpt-oss
bundle that was never stamped still needs the right stop set.

## Blast radius, measured

Across all **86** installed bundles: **0** are harmony-shaped, and **0** carry `<|end|>` at all.
The change is therefore provably inert for the entire local library — confirmed by A/B below.

## Tests

`HarmonyChannelSeparatorStopTests` — 5 tests pinning both sides of the gate:

- harmony vocab + declaration omitting `<|end|>` → not a stop *(the bug)*
- **Phi-shaped vocab + a declared eos → `<|end|>` IS still a stop** *(the non-regression #91 lost)*
- harmony vocab, no declaration → blanket still applies
- harmony vocab that *does* declare `<|end|>` → declaration wins
- Qwen-shaped under-declaration → `<|im_end|>` still resolved

Fixture mirrors the real load path: `ModelTokenConfigurationResolver.resolvedEOSTokenIds` copies a
declared eos into `eosTokenIds`, so a fixture setting only `generationDefaults` would understate
the stop set.

**Sabotage-checked.** Reverting only `StopStringMatcher.swift` and keeping the tests:
```
✘ harmonyEndIsNotAStop() — (resolved.tokenIDs → [200002, 200007, 200012, 199999]).contains(200007)
```
The other four pass on both sides, which is what a correctly-scoped pin should do.

`swift test --filter 'HarmonyChannelSeparatorStopTests|UnderDeclaredEOSStopTests|StopStringMatcherTests|Gemma4StaleEOSTests'` → **30/30 pass**.

### DIFF
--- a/Libraries/MLXLMCommon/StopStringMatcher.swift
+++ b/Libraries/MLXLMCommon/StopStringMatcher.swift
@@ -170,6 +170,28 @@ let commonEndTokenStrings: [String] = [
     "<|EOT|>",          // Uppercase EOT variants in some shipped templates
 ]
 
+/// The one `commonEndTokenStrings` entry whose meaning is not universal.
+private let harmonyChannelSeparator = "<|end|>"
+
+/// Tokens that, taken together, identify the OpenAI **harmony** channel format
+/// (gpt-oss). No other shipped family carries all three.
+private let harmonyVocabularyMarkers = ["<|channel|>", "<|message|>", "<|return|>"]
+
+/// True when this bundle's vocabulary is harmony-shaped.
+///
+/// Detected from the VOCABULARY rather than `reasoningParserName`, because the
+/// capability stamp is optional: a locally converted gpt-oss bundle that was
+/// never stamped still needs the right stop set. `convertTokenToId` returns the
+/// unk id for unknown tokens, so id existence alone proves nothing — every
+/// marker is checked against unk.
+private func vocabularyIsHarmonyChannelFormat(_ tokenizer: Tokenizer) -> Bool {
+    harmonyVocabularyMarkers.allSatisfy { marker in
+        guard let id = tokenizer.convertTokenToId(marker) else { return false }
+        if let unknownID = tokenizer.unknownTokenId, id == unknownID { return false }
+        return true
+    }
+}
+
 func mergeStopStrings(_ explicit: [String], _ defaults: [String]) -> [String] {
     var seen = Set<String>()
     var merged: [String] = []
@@ -202,7 +224,31 @@ func resolveStopSequences(
             textStopStrings: &textStopStrings,
             allowExactTextFallback: true)
     }
+    // `<|end|>` is carried by `commonEndTokenStrings` for Phi-3/Phi-4, where it genuinely ends
+    // the turn. In the harmony channel format it means the opposite: the model closes its
+    // `analysis` channel with `<|end|>` and then OPENS `final`, so treating it as a stop halts
+    // generation at the end of the reasoning channel and the visible answer is empty. gpt-oss
+    // says so itself — its `generation_config.json` declares `<|return|>` / `<|call|>` /
+    // `<|endoftext|>` as eos and deliberately omits `<|end|>`.
+    //
+    // Suppress it only on TWO pieces of positive evidence: the vocabulary is harmony-shaped, and
+    // the model's own declaration excludes it. Anything less keeps today's behaviour. In
+    // particular the blanket is NOT skipped wholesale for every declaring bundle: bundles that
+    // declare an incomplete eos are common (VibeThinker-3B declares only `<|endoftext|>` while
+    // its template ends turns with `<|im_end|>`), and the blanket is their only turn boundary.
+    // See `UnderDeclaredEOSStopTests`, which fails against that broader gate.
+    let declaredEOS = Set(modelConfiguration.generationDefaults?.eosTokenIds?.values ?? [])
+    let suppressedHarmonyEnd: Int? = {
+        guard !declaredEOS.isEmpty, vocabularyIsHarmonyChannelFormat(tokenizer),
+            let endID = tokenizer.convertTokenToId(harmonyChannelSeparator),
+            tokenizer.unknownTokenId != endID,
+            !declaredEOS.contains(endID)
+        else { return nil }
+        return endID
+    }()
+
     for token in commonEndTokenStrings {
+        if suppressedHarmonyEnd != nil, token == harmonyChannelSeparator { continue }
         resolveStopTokenString(
             token,
             tokenizer: tokenizer,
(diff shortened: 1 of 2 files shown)


jjang-ai:
## Live QA — variating turns, per family

The change can only alter the **stop set**, which every family shares, so the bar is that
non-harmony families are *byte-identical* to main. Each row below is a real run on this M5 Max,
branch vs `main`, same binary rebuilt per side (binary timestamps checked — a stale binary is how
a false "IDENTICAL" slipped through on #137).

### LFM2.5-VL-3B-MXFP8 — full variating pattern (`BENCH_VL_VARIATING=1`)

One conversation walked across reasoning × media × tools, cache probed **before every turn**:

| turn | shape | cache probe | result |
|---|---|---|---|
| 1 | reasoning ON, text | MISS 25 tok, media=false | `text="Paris."` |
| 2 | reasoning OFF, **image added mid-conversation** | MISS 301 tok, media=true | `text="A red-to-blue gradient."` |
| 3 | tools offered, no new image | MISS 379 tok, media=true | **toolCalls=1** → `get_weather(location:"Tokyo")` |
| 4 | **image AND tools in the same turn** | MISS 664 tok, media=true | text, no call |
| 4b | image+tools, direct phrasing | MISS 655 tok, media=true | text, no call |
| 5b | two images in one turn | MISS 542 tok, media=true | answered |
| 6 | grow A | MISS 22 tok | `text="Red."` |
| 7 | grow B, same shape | **HIT disk 22/38 (57% reused)** | `text="Blue."` |
| 5 | **replay turn 1 verbatim** | **HIT disk 26/26 (100% reused)** | `text="Paris."` — head boundary survived |

`final cache stats: paged{hits=0,misses=0} disk{hits=3,misses=74,stores=17}` → **passed**.
Diff vs main: **IDENTICAL** (only `Load:` jitter, 0.27s vs 0.43s).

Pre-existing on both sides, not introduced here: image+tools produces no structured call under
either phrasing. The render probe shows the tools block *does* reach the prompt with an image
present (`image turn: with tools=655 tok, without=605, delta=50`), so it is model behaviour, not a
dropped schema.

### VibeThinker-3B-MXFP8 — the under-declared bundle this gate is about

The family most exposed to a stop-set change: declares only `<|endoftext|>` while its template
ends turns with `<|im_end|>`.

- Variating turn 1 (reasoning ON): `think=true reasoning=317ch text="Paris"` — clean stop, so the
  blanket-supplied `<|im_end|>` boundary still works.
- Growing-chat cache (`BENCH_GROWING_CHAT_CACHE=1 BENCH_GROWING_THINK=1`): **IDENTICAL to main**.

Turns 2+ of the variating pattern are not applicable — VibeThinker is a text-only Qwen2 tower and
the pattern requires an image turn. It also fails the recall-phrase gate **on main** (emits
`vml-cache-green`, dropping a character under greedy) — pre-existing, reproduced identically on
both sides.

### Bonsai-27b-Ternary-JANG-CRACK — hybrid SSM+attention cache

`layers=64, kvLayers=16, mambaLayers=48, companion=ssm, restore=disk-backed`

- `Prompt-boundary salted probe: HIT tier=paged matched=18/24`
- `Post-answer salted probe: HIT tier=paged matched=18/187`
- nil-salt probes MISS (correct — salt is load-bearing)
- `Coordinator probe before turn 2: HIT tier=paged matched=18/51`
- Diff vs main: **IDENTICAL**

Turn 2 ending `.length` rather than `.stop` reproduces on main unchanged.

### Families with no local coverage — stated plainly

**gpt-oss / harmony is not installed here, so the fix itself is not live-proven.** It is proven by
unit test and by the sabotage check; the live runs above prove only that nothing else moved. The
model IS supported in-tree (`GPTOSSModel`, registry `gpt_oss_20b_MXFP4_Q8`), so the bug is
reachable — it just cannot be exercised on this machine without pulling the bundle.

Phi-3/Phi-4 is likewise not installed. Its non-regression (`<|end|>` stays a stop for a
non-harmony vocab) is covered by `phiKeepsEndAsAStopEvenWhenItDeclaresEOS`.

### Why no sweep of every other family

The vocabulary scan is the stronger statement: **0 of 86** installed bundles are harmony-shaped and
**0 carry `<|end|>` at all**, so the suppression branch is unreachable for every one of them and
the resolved stop set is unchanged by construction. The three A/B runs confirm that empirically
across a VL family, a text-only Qwen2, and a hybrid SSM.

PR #255, opened 2026-08-13
Fix JANG semantic-width quantization precedence (supersedes #159)

jjang-ai:
Carries #159 (@mimeding) forward onto current `main`, which has moved under it. Same thesis:
**architecture-derived geometry should outrank stale per-path metadata**, not the reverse.

## What changed vs #159

`main` gained a centralized `PerLayerQuantization.explicitQuantizationOption` /
`layerPathCandidates` that already normalizes `model.` prefixes, attention aliases and the DSV4
shared-expert rename in one place. #159's parallel `declaredPathVariants` / `declaredSpecificQuantization`
lookup duplicates that (and orders its candidates through `Array(Set(...))`, which is not stable),
so this keeps main's helper and layers on only what main genuinely lacks:

1. **Precedence.** The exact-width branches for the language hidden anchor and hidden-input
   projections now run **before** the declared-metadata branch. Previously a stale declaration that
   happened to width-collide with the packed tensors won over the architecture's own `hidden_size`.
2. **Anchor tightening.** `isHiddenAnchor` matched `basePath.hasSuffix("embed")`, which also catches
   **`vision_tower.pos_embed`** — a vision tensor forced onto the language `hidden_size`. Now
   restricted to `embed_tokens` / `lm_head` spellings.
3. **Contradictory aliases.** #159's `hasDeclaredSkip` rule is re-expressed on main's helper as
   `declaresSkipForAnyAlias`. When one spelling declares `.quantize` and another `.skip`, the winner
   used to be decided by candidate *order* — a property of the alias list, not the bundle. That
   declaration is now set aside and the width comes from the tensors.

#159's test for (3) is kept and passes against this implementation.

## The DSV4 scare, checked

Claim (2) looked dangerous: DSV4-Flash's language embedding is literally named `embed`
(`embed.weight` + `embed.scales`, quantized, and the bundle has **no** `lm_head`), so tightening the
anchor looked like it would drop the flagship's embedding.

It does not — the shape walk sees **runtime** module paths, not checkpoint names, and DSV4's `embed`
arrives already renamed to `model.embed_tokens`. Verified directly rather than reasoned about:
```
[QDUMP] model.embed_tokens bits=8 gs=64
```

## Blast radius, measured per layer

Count-level equality is not proof, so I dumped the **full resolved per-layer quantization map** on
both sides (temporary `VMLX_QUANT_DUMP` instrumentation, not committed) and diffed:

| bundle | layers | result |
|---|---|---|
| `LFM2.5-VL-3B-JANG_4M` | 169 | IDENTICAL |
| `Ornith-1.0-9B-JANG_6M` | 250 | IDENTICAL |
| `LFM2.5-2.6B-JANG_6M` | 167 | IDENTICAL |
| `VibeThinker-3B-JANG_4M` | 253 | IDENTICAL |
| `DeepSeek-V4-Flash-0731-JANG-CRACK` | 641 | IDENTICAL |

1,480 layers, zero differences. Since the resolved quant map is bit-identical, the loaded weights
are identical and generation cannot diverge.

## Is the fix even reachable?

Worth stating, because a guard that is correct but never reached is worth nothing. Scanning tensor
manifests across 76 bundles, **13** carry a `*embed` path that is not `embed_tokens` — 11×
`vision_tower.pos_embed` (Ornith 9B/35B, Bonsai 27b, Qwen3.6 family) plus DSV4's `embed`.

So the mis-anchor is reachable in principle. In practice today it changes nothing: `pos_embed` is
not quantized, so it never enters the shape walk, and DSV4's path is renamed before it gets there.
This is therefore **latent-defect hardening**, not an observable fix — a VL bundle that *did* ship a
quantized `pos_embed` would be mis-anchored on `main` today.

## Tests

`swift test --filter 'MTPRuntimeFocusedTests|DSV4 segmented pool quantization'` → **92/92 pass**,
including #159's 326 lines of new coverage (anchor, manifest, alias, vision, routed-expert).

### DIFF
--- a/Libraries/MLXLLM/LLMModelFactory.swift
+++ b/Libraries/MLXLLM/LLMModelFactory.swift
@@ -1892,10 +1892,10 @@ public final class LLMModelFactory: ModelFactory {
             for: templateResolvedDir)
         async let tokenizerTask = tokenizerLoader.load(from: tokenizerDirectory)
 
-        // When JANG, skip config.json's perLayerQuantization — JANG infers correct
-        // per-layer bits from tensor shapes. This avoids creating QuantizedLinear at
-        // the wrong bit width (which can't be re-quantized later).
-        // BUT: still pass `quantization` (the global config.json group_size /
+        // For JANG, pass config.json's per-layer metadata as declared evidence;
+        // loadWeights validates it against exact manifests, semantic widths, and
+        // tensor geometry before constructing QuantizedLinear modules.
+        // Also pass `quantization` (the global config.json group_size /
         // bits) so JangLoader.inferPerLayerQuantization gets the correct
         // `knownGroupSize` even when jang_config.json doesn't carry quant
         // metadata (e.g. DSV4-Flash bundles ship `weight_format: "bf16"`).
@@ -1903,10 +1903,9 @@ public final class LLMModelFactory: ModelFactory {
             modelDirectory: modelDirectory, model: model,
             quantization: jangConfig != nil ? baseConfig.quantizationContainer?.quantization : nil,
             // 2026-04-28: pass perLayerQuantization through even when JANG;
-            // loadWeights merges config.json's explicit per-layer dict on
-            // top of the shape walk to fix mis-inference of (bits, gs) pairs
-            // that share a packed shape (e.g., (4,64) ≡ (8,32)). Closes
-            // Cascade-2 JANG_4M / Nemotron-Omni MXFP4 first-prefill rmsNorm.
+            // loadWeights treats config.json's explicit per-layer dict as
+            // declared evidence, then validates it against exact manifests,
+            // semantic widths, and packed tensor geometry.
             perLayerQuantization: baseConfig.perLayerQuantization,
             jangConfig: jangConfig,
             loadPreservedMTP: loadNativeMTP)
--- a/Libraries/MLXLMCommon/BaseConfiguration.swift
+++ b/Libraries/MLXLMCommon/BaseConfiguration.swift
@@ -99,6 +99,25 @@ public struct BaseConfiguration: Codable, Sendable {
             return nil
         }
 
+        /// True when ANY spelling of `layer` is declared `.skip`.
+        ///
+        /// A bundle that declares one spelling `.quantize` and another `.skip`
+        /// for the same tensor is self-contradictory. `explicitQuantizationOption`
+        /// resolves that by candidate ORDER, which silently makes the answer a
+        /// property of the alias list rather than of the bundle. Callers that use
+        /// a declaration to disambiguate an otherwise-ambiguous packed width ask
+        /// this first, so contradictory metadata is set aside and the width is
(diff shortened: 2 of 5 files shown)


jjang-ai:
## Live QA — variating turns

A loader change touches **every** family, so the bar is that a real conversation is unchanged
across turn shapes. Runs on this M5 Max, branch vs `main`, binary rebuilt per side and timestamps
checked.

### LFM2.5-VL-3B — full variating pattern, cache probed before every turn

| turn | shape | probe | result |
|---|---|---|---|
| 1 | reasoning ON, text | MISS 25 | `"Paris."` |
| 2 | reasoning OFF, **image added mid-conversation** | MISS 301, media=true | `"A red-to-blue gradient."` |
| 3 | tools offered | MISS 379, media=true | **toolCall** `get_weather(location:"Tokyo")` |
| 4 | **image AND tools in one turn** | MISS 664, media=true | answered |
| 5b | two images in one turn | MISS 542, media=true | answered |
| 6→7 | grow A → grow B | **HIT disk 22/38 (57%)** | `"Red."` → `"Blue."` |
| 5 | **turn 1 replayed verbatim** | **HIT disk 26/26 (100%)** | `"Paris."` |

Passed; diff vs main is only `Load:` jitter.

### Bonsai-27b-Ternary — hybrid SSM + attention

`layers=64 kvLayers=16 mambaLayers=48 companion=ssm restore=disk-backed`
`Prompt-boundary salted probe: HIT tier=paged matched=18/24` ·
`Post-answer: HIT 18/187` · nil-salt probes MISS (correct) · IDENTICAL to main.

### VibeThinker-3B-JANG_4M / LFM2.5-2.6B-JANG_6M / Ornith-1.0-9B-JANG_6M / DSV4-Flash

Growing-chat cache runs IDENTICAL to main, and the per-layer quantization maps are bit-identical
(table in the PR body — 1,480 layers, zero differences).

### Quant coverage across the family, not one quant

The per-layer A/B deliberately spans **JANG_4M, JANG_6M, MXFP8 and Ternary** so the precedence
change is exercised at 4-bit, 6-bit, 8-bit and ternary defaults rather than one width:

```
LFM2.5-VL-3B-JANG_4M   default bits=4  gs=64   169 overrides
Ornith-1.0-9B-JANG_6M  default bits=6  gs=64   250 overrides
LFM2.5-2.6B-JANG_6M    default bits=8  gs=32   167 overrides
VibeThinker-3B-JANG_4M default bits=4  gs=64   253 overrides
DSV4-Flash-JANG-CRACK                          641 overrides
```

### Stated plainly

Because the resolved quant map is bit-identical on every bundle I can load, this change is **not
observable** on the current library — it removes a latent mis-anchor rather than fixing a
reproducible symptom. I could not construct a live failure on `main` to fix, and I am not going to
present the identical-output runs as evidence that it fixes something user-visible. What they prove
is the inverse, which is the risk that actually mattered here: precedence was reordered in the
loader and nothing regressed.

PR #256, opened 2026-08-13
Expose native-MTP per-generation stats on GenerateCompletionInfo (supersedes #131)

jjang-ai:
Carries #131 (@mimeding) onto current `main`. Same goal: the `[NativeMTP]` per-generation summary is
stderr-only, so a host that wants speculative-decoding health has to grep stderr. This puts the
headline counters on `GenerateCompletionInfo`, following the existing `turboQuantCompressions`
pattern.

## Conflicts resolved (8, all in the two hottest files)

- **`GenerateCompletionInfo` keeps every field.** `main` added `turboQuantCacheTransition` and
  `toolCallProtocolFailure` after this branch was written; a straight merge would have dropped both.
  The stored property, initializer parameters, initializer body and construction site now carry all
  three additions.
- **`NativeMTPGenerationStats` moved out of `NativeMTPHybridWarmupMemo`.** `main` added that memo in
  the same region; the naive union nested the new struct inside the enum.
- **`depth` semantics corrected.** `main` caps native MTP at `VMLX_MTP_DEPTH_CAP` (default `1`) —
  depths above 1 only ever won on a deterministic counting prompt and measured ~14% slower than
  plain AR decode on prose. So `depth` is the **effective** depth: a host asking for
  `.nativeMTP(depth: 3)` reads `1`. The branch documented it as the *requested* depth and its test
  asserted `3`. Both now match shipped behaviour — and that request-vs-effective gap is exactly the
  kind of thing this surface exists to make visible without reading stderr.

## The ordering change, checked

To populate the stats the completion-info construction moves **after**
`storeCacheAfterGeneration` (the iterator snapshots there, since `outputTokens` depends on
`generatedTokenIds`, which it only sees in that call). `.info` is still yielded at the same point.

That is a real reorder in the hot generate path, so I did not take the branch's word for it:

- `generateTime` / `promptTime` are captured **before** the move, so timings are unaffected.
- `lastTurboQuantCacheTransition` is written in `maybeQuantizeCacheForStep()` — a per-decode-step
  call — not during the cache store.
- End-to-end A/B, branch vs `main`, on `VibeThinker-3B-MXFP8` and `Bonsai-27b-Ternary-JANG-CRACK`
  (growing-chat cache, reasoning on): **IDENTICAL**, including the `TurboQuant compressions` line.

## Tests

`swift test --filter MTPRuntimeFocusedTests` → **80/80 pass**.

### DIFF
--- a/Libraries/MLXLMCommon/Evaluate.swift
+++ b/Libraries/MLXLMCommon/Evaluate.swift
@@ -945,6 +945,7 @@ public protocol TokenIteratorProtocol: Sequence, IteratorProtocol where Element 
     var promptPrefillTime: TimeInterval { get }
     var promptTokenIds: [Int] { get }
     var turboQuantCompressionCount: Int { get }
+    var nativeMTPStats: NativeMTPGenerationStats? { get }
     mutating func storeCacheAfterGeneration(
         generatedTokenIds: [Int],
         includeGeneratedBoundary: Bool)
@@ -953,6 +954,7 @@ public protocol TokenIteratorProtocol: Sequence, IteratorProtocol where Element 
 extension TokenIteratorProtocol {
     public var promptTokenIds: [Int] { [] }
     public var turboQuantCompressionCount: Int { 0 }
+    public var nativeMTPStats: NativeMTPGenerationStats? { nil }
 
     public mutating func storeCacheAfterGeneration(
         generatedTokenIds: [Int],
@@ -3523,15 +3525,6 @@ private func generateLoopTask<Handler: TokenLoopHandler>(
             MLXPressGenerationProfile.dumpAndReset(
                 reason: "generation-end tokens=\(tokenCount)")
 
-            let info = GenerateCompletionInfo(
-                promptTokenCount: promptTokenCount,
-                generationTokenCount: tokenCount,
-                promptTime: promptTime + iterator.promptPrefillTime,
-                generationTime: generateTime,
-                stopReason: stopReason ?? .cancelled,
-                turboQuantCompressions: iterator.turboQuantCompressionCount,
-                unclosedReasoning: unclosedReasoning
-            )
             // Multi-tier cache: drain the final async token eval before cache
             // snapshot/store, then keep completion info behind the cache-store
             // drain. Local chat/tool consumers may start a post-tool decode as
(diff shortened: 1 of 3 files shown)


jjang-ai:
## Live QA — and a finding that limits it

### What I could prove live

The change touches the shared generate path (the completion-info construction moved), so the bar is
that ordinary generation is unchanged across turn shapes. Branch vs `main`, binary rebuilt per side:

- **VibeThinker-3B-MXFP8** — growing-chat cache, reasoning ON: **IDENTICAL to main**
- **Bonsai-27b-Ternary-JANG-CRACK** — hybrid SSM+attention, `layers=64 kvLayers=16 mambaLayers=48
  companion=ssm restore=disk-backed`; `Prompt-boundary salted probe: HIT tier=paged matched=18/24`,
  `Post-answer: HIT 18/187`, nil-salt probes MISS: **IDENTICAL to main**
- **LFM2.5-VL-3B-MXFP8** — full variating pattern (reasoning on↔off, text↔image↔text-after-image,
  tools none↔offered↔called, image+tools in one turn, two images in one turn, cache probed before
  every turn, turn 1 replayed verbatim at the end → `HIT disk 26/26 100% reused`): passed, and the
  `TurboQuant compressions` line is unchanged.

That covers the regression risk, which is the part that could hurt.

### What I could NOT prove live — stated plainly

**The stats themselves never populate on any bundle in this library.** I wired a temporary
`BENCH_MTP_STATS` print into the harness to read `info.nativeMTPStats` from a real run and could not
get a single model to reach the native-MTP iterator:

| bundle | why native MTP refuses |
|---|---|
| `DeepSeek-V4-Flash-0731-JANG-CRACK` | `unsupported model type(s): deepseek_v4` |
| `Qwen3.6-27B-MXFP8-CRACK-MTP` (and the other three `-CRACK-MTP`) | `mtp: preserved_enabled, layers=1, tensors=23, speculative=off` — **no `vmlx_mtp_tuning.json`** |
| `Ornith-1.0-9B-MXFP8` (qwen3_5, a supported type) | `metadata_only_missing_weights, layers=1, tensors=0` |
| `Hy3-JANG_2K-MTP` — the **only** bundle in `~/models` carrying `vmlx_mtp_tuning.json` | `unsupported model type(s): hy_v3` |

Native MTP is implemented for `Qwen35` (LLM + VLM) and `NemotronH`. Every locally installed bundle
of those types either lacks the MTP weights or lacks the tuning artifact, and the one bundle with a
tuning artifact is a model type that isn't dispatched. So the populated-stats path is currently
unreachable on this machine, and I am not going to dress the identical-output runs up as proof that
the counters are correct.

What backs the counters is `activeNativeMTPGenerationCarriesStructuredStats` (probe target, real
`BatchEngine` stream), which asserts `outputTokens == generationTokenCount`,
`acceptedByDepth.reduce(+) == verifyCalls`, the cache mode string, and — after this resolution — that
`depth` reports the **capped** depth rather than the requested one.

Worth flagging separately: that table means the shipped MTP bundles cannot actually engage
speculative decode as-is. That is a bundling gap, not a code gap, and it is outside this PR.

PR #266, opened 2026-08-14
Add Apertus 1.5 (apertus1p5) — discrete-token omni: text, vision, audio, and its reasoning rail

rcfa:
Add Apertus 1.5 (apertus1p5) — discrete-token omni: text, vision, audio

Apertus 1.5 is multimodal in a way no other model in this repo is, and it shapes
the whole port. Instead of an encoder plus a projector feeding continuous
embeddings into the hidden space, it QUANTISES each modality into ids from a
codebook that lives inside the token embedding table, and feeds them as ORDINARY
TOKENS. The config arithmetic states the design exactly:

    image_token_offset 131272 + image codebook 131072 == 262344 == audio_token_offset
    audio_token_offset 262344 + audio codebook   4096 == 266440 <= vocab 266752

So the language model needs no architectural change at all: `ApertusModel` is
reused unmodified, and the port is "pixels -> integers" and "waveform ->
integers".

WHAT IS ADDED

* `Apertus1p5VisionTokenizer` — the encoder half of a VQ-VAE (taming-transformers
  `Encoder`): conv_in, five stages [256,256,512,512,1024] over four downsamples,
  attention at the 16x16 stage, mid block, then conv_out -> quant_conv ->
  nearest-neighbour in a [131072, 256] codebook. Only the encoder ships in the
  checkpoint (247 tensors, no decoder), which is right for a model that consumes
  images and never generates them.
* `Apertus1p5AudioTokenizer` — the encoder half of WavTokenizer (EnCodec/SEANet):
  input conv, four downsampling stages (strides 4*5*5*6 = 600, the documented
  hop), a 2-layer LSTM, output conv, then a [4096, 512] codebook. Of the 226
  audio tensors only ~63 are used; `backbone.convnext`, `backbone.pos_net` and
  `head.linear` [2402, 768] are the ISTFT vocoder that turns tokens back INTO
  audio, which a consuming model never runs.
* `Apertus1p5` — the model, holding both tokenizers plus `ApertusModel`. Media
  ids are spliced in `prepare` and the ordinary text path takes over.
* `Apertus1p5Processor` — preprocessing and the media token framing.

DETAILS THAT ARE SILENT IF WRONG, each commented at its site

* the image grid is written as LITERAL TEXT: the framing is
  `<|img_start|>{H}*{W}<|img_token_start|>{rows}<|img_end|>` where rows are
  `<|image|>` repeated W times, joined by `<|img_end_of_row|>`. The dimensions
  are an ordinary tokenised string — that is how the model is told the geometry —
  and each patch gets its OWN placeholder, substituted one code per placeholder.
* resolution is DYNAMIC: `smart_resize` clamps AREA to [256^2, 1400^2] preserving
  aspect ratio, then rounds both sides half-up to a multiple of 16. A 640x480
  photo becomes a 30x40 grid (1200 tokens), not 16x16.
* audio convolutions are stored under PyTorch weight-norm (`original0` = g,
  `original1` = v); the effective weight is `g * v / ||v||`.
* the audio residual block NUMBERS its parameterless activations (`block.1`,
  `block.3`), so the convolutions are not at indices 0/1.
* torch's 2-layer LSTM maps onto MLX's `Wx`/`Wh`/`bias` with the two biases
  SUMMED; gate order needs no change, as both split i,f,g,o.
* audio clips are peak-normalised to -3 dBFS before the codec, as the reference
  pipeline does (its feature extractor deliberately does not normalise).
* quantised bundles ship each codebook twice — plain float plus a packed
  `.weight`/`.scales`/`.biases` triple — and the packed copies are dropped.

All of the above is ported from the reference implementation
(github.com/swiss-ai/transformers, branch `add-apertus1p5`), not inferred.

VERIFICATION, on device (M4 Max, macOS 27)

* structural: every checkpoint tensor binds to the module tree with matching
  shapes and nothing is left unbound, under strict `.all` verification, for both
  tokenizers;
* weight-norm: one weight is rebuilt by hand from its own g and v and must match
  the loaded parameter — and must NOT equal the bare direction;
* vision: 75.0% on POPE (40 items). A known-good VLM scores 72.5% on the same
  items;
* audio: given a clip it was not told the content of, and asked only "what do you
  hear?", the model answers "The numbers one, two, three, four, five, six,
  seven, eight, nine, ten." — an exact transcription.

Quantised bundles load and run: verified on a 6-bit JANG conversion (8.1 GB) for
text, vision and audio. An earlier draft of this description reported quantised
bundles as broken; that turned out to be a defect in the CONVERSION — the
multimodal encoders were shadowed by de-listed raw tensors — and not in the
runtime. Repairing the bundle was sufficient; no runtime change was needed.

## Reasoning rail (0022 + 0023)

Apertus 1.5 marks its inner monologue with two dedicated vocabulary tokens rather
than `<think>` XML — the chat template sets `inner_token = '<|inner_prefix|>'` and
`outer_token = '<|inner_suffix|>'` (lines 152-153) — so both the reasoning parser
and the reasoning budget need entries, or the model's monologue is indistinguishable
from its answer.

`reasoningStampFromModelType` had no entry for the family, so `apertus1p5` fell
through to the terminal `"none"` and NO parser was selected. Measured on
Apertus-v1.5-8B before the fix, gsm8k, three turns:

    text=3998ch reasoning=0ch     text=1531ch reasoning=0ch     text=886ch reasoning=0ch

Zero reasoning characters on every turn: the whole monologue, raw markers included,
went to the user-visible channel with the answer buried behind it. After:

    text=10ch  reasoning=2839ch   text=10ch  reasoning=1069ch   text=409ch reasoning=512ch

The 10-character turns are `\boxed{87}` and `\boxed{34}` — the model does its
working in the monologue and emits only the boxed result, which is exactly the
intended split. gsm8k 3/3 on the same run; zero raw markers anywhere in the output.

Scoped to 1.5 deliberately: plain `apertus` is Apertus 1.0, which has no monologue
and is enumerated as a non-thinking family in `testPlainFamiliesGetNone`. A first
draft matched the bare `apertus` prefix and turned that suite red; the regression
test now pins both directions.

0022 additionally lets a reasoning BUDGET arm for Apertus: without its two entries
`ReasoningBudget.arm` cannot resolve a close token for an Apertus bundle, so a
requested ceiling was silently inert rather than failing loudly.

## Prefix normalisation delegates to the shared helper

The language tower is routed through one prefix normalisation, because the bf16
release and the MLX quantisations disagree about where the `language_model`
segment sits:

```
bf16 (HF layout):  model.language_model.layers.0...   lm_head.weight
MLX quantisations: language_model.model.layers.0...   language_model.lm_head.weight
```

That rule is `Weights.stripLanguageModelPrefix`, not a local loop. The rule
itself is byte-identical — what the helper adds is the **contested key**:

> a mixed-provenance re-bake can carry **both** spellings … resolving that by
> last-write-wins would make the bind depend on `Dictionary`'s iteration order,
> which is **seeded per process**

A local `lm[key] = v` has exactly that shape. The model would load, generate,
and be wrong some fraction of the runs — the one failure nothing downstream
reports.

The traffic is real, not hypothetical:

| bundle | tensors through this path | of which `lm_head` |
|---|---|---|
| `Apertus-v1.5-8B-Jang_6M` | 836 | 3 |
| `Apertus-v1.5-70B-Jang_6M` | 2,084 | 3 |

Three of them are the head, which is why the earlier head-scoped
`only: ["lm_head."]` fix on `ApertusModel` did not cover this.

**Mutation-checked**: restoring the local loop fails the guard, naming both
re-derived lines. The behavioural claim is asserted against the helper — both
layouts normalising, and a contested key binding the unprefixed spelling the
same way whichever order the dictionary is built in.

### DIFF
--- a/Libraries/MLXLMCommon/ReasoningBudget.swift
+++ b/Libraries/MLXLMCommon/ReasoningBudget.swift
@@ -71,6 +71,7 @@ public enum ReasoningBudget {
         "</thinking>",
         "<|/think|>",
         "<|end_thought|>",
+        "<|inner_suffix|>",  // Apertus 1.5 (apertus1p5)
         "<end_of_turn>",
     ]
 
@@ -81,6 +82,8 @@ public enum ReasoningBudget {
     /// that only armed on a primed tail would silently skip that second group.
     public static let openTokenCandidates: [String] = [
         "<think>", "<thinking>", "<|think|>", "<|start_thought|>",
+        "<|inner_prefix|>",  // Apertus 1.5: its template primes neither tag, so the model
+                             // emits the open marker itself as its first generated token.
     ]
 
     /// Resolve the open-token ids that exist in this bundle's vocab, so a
--- a/Libraries/MLXLMCommon/ReasoningParser.swift
+++ b/Libraries/MLXLMCommon/ReasoningParser.swift
@@ -872,6 +872,37 @@ extension ReasoningParser {
             return ReasoningParser(startInReasoning: true)
         }
 
+        // Apertus 1.5 wraps its inner monologue in two dedicated vocabulary tokens
+        // rather than XML. Its chat template sets them explicitly (lines 152-153):
+        //
+        //     {%- set inner_token = '<|inner_prefix|>' -%}
+        //     {%- set outer_token = '<|inner_suffix|>' -%}
+        //
+        // and the generation prompt tail is a bare `<|assistant|>` — no prefilled
+        // opener — so the model emits `<|inner_prefix|>` itself and
+        // `startInReasoning` stays false, as for Gemma-4.
+        //
+        // Observed on cubiculum/Apertus-v1.5-8B-Jang_6M (gsm8k, 3/3 turns):
+        //
+        //   <|inner_prefix|>We need to parse the question. …3.7 kB of monologue…
+        //   <|inner_suffix|>The friend gave \(24\) outfits. … \boxed{87}
+        //
+        // Opens and closes were balanced 3/3, so the unclosed-block hazard the
+        // Hunyuan entry documents does not apply. Both markers are single special
+        // tokens that cannot occur in prose, so strays are stripped.
+        //
+        // SCOPED TO 1.5 DELIBERATELY. Plain `apertus` is Apertus 1.0, which has no
+        // monologue and is enumerated as a non-thinking family in
+        // `ReasoningStampFromModelTypeTests.testPlainFamiliesGetNone`; matching the
+        // bare `apertus` prefix here would contradict that contract.
+        if normalized.hasPrefix("apertus1p5") {
+            return ReasoningParser(
+                startTag: "<|inner_prefix|>",
+                endTag: "<|inner_suffix|>",
+                startInReasoning: false,
+                stripStrayTags: true)
+        }
+
         if normalized.hasPrefix("mistral4")
             || normalized.hasPrefix("mistral_4")
             || normalized.hasPrefix("mistral_small_4")
@@ -1258,6 +1289,20 @@ public func reasoningStampFromModelType(_ modelType: String?) -> String {
     // Harmony marker leak.
     if compact.hasPrefix("gptoss") {
         return "harmony"
+    }
+
+    // Apertus 1.5 — dedicated `<|inner_prefix|>` / `<|inner_suffix|>` monologue
+    // markers, resolved by `fromCapabilityName`. Without this stamp `apertus1p5`
+    // falls through to the terminal "none", no parser is selected, and the entire
+    // monologue — raw markers included — streams into the user-visible chunk.
+    // Measured on Apertus-v1.5-8B before this entry: text=3998ch reasoning=0ch,
+    // with the answer buried 3.7 kB deep. Same failure shape as the `gptoss`
+    // Harmony leak and the `muse_glimmer` fall-through.
+    //
+    // Matches 1.5 ONLY: plain `apertus` (1.0) has no monologue and must keep its
+    // "none" stamp — see `testPlainFamiliesGetNone`.
+    if compact.hasPrefix("apertus1p5") {
+        return "apertus1p5"
     }
 
     // ZAYA1-VL is a sibling multimodal architecture, not the text
--- a/Tests/MLXLMTests/ApertusReasoningNoLeakTests.swift
+++ b/Tests/MLXLMTests/ApertusReasoningNoLeakTests.swift
@@ -1,0 +1,155 @@
+// Pin the Apertus 1.5 reasoning-no-leak contract.
+//
+// Apertus 1.5 does not use `<think>` XML. Its chat template defines two dedicated
+// vocabulary tokens (chat_template.jinja lines 152-153):
+//
+//     {%- set inner_token = '<|inner_prefix|>' -%}
+//     {%- set outer_token = '<|inner_suffix|>' -%}
+//
+// and the generation prompt tail is a bare `<|assistant|>` — the template prefills
+// NO opener — so the model emits `<|inner_prefix|>` itself when it decides to think.
+//
+// THE DEFECT THIS PINS. `reasoningStampFromModelType("apertus1p5")` returned "none",
+// so `JangLoader` selected NO parser, so the monologue and the raw marker streamed
+// straight into the user-visible channel. Measured on
+// cubiculum/Apertus-v1.5-8B-Jang_6M before the fix, gsm8k, 3 turns:
+//
+//     ── GEN channels: text=3998ch reasoning=0ch finish=stop tokens=1033
+//
+// Zero reasoning characters on every turn, with the actual answer buried 3.7 kB
+// deep behind the monologue. This is the same shape as the `gptoss` Harmony marker
+// leak and the `muse_glimmer` fall-through that the neighbouring stamps exist to
+// prevent — and, as the `muse_glimmer` comment records, the parser is the half
+// people fix while the stamp is the half that makes it reachable. BOTH halves are
+// required: a test that only checked `fromCapabilityName("apertus")` would pass
+// against a build where real Apertus bundles still leak.
+//
+// Source-coverage style — no MLX runtime needed.
+
+import Foundation
+import Testing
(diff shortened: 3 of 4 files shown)


jjang-ai (commented):
Had a proper look. The shared-file edits are clean, and there is one omission that will cost this model badly at runtime — worth fixing before merge because Apertus is close to the worst case for it.

## The one that matters: `mediaTokenIds` is never declared

The diff declares media placeholder IDs nowhere. Every other multimodal family does — `Gemma4.swift:1683`, `FastVLM.swift:1030`, `DeepseekOCRProcessor.swift:450`, `Audex.swift:838` (`mediaTokenIds: audio == nil ? nil : [Self.soundTokenId]`).

That is not cosmetic. `LanguageModel.swift` treats "undeclared" as "assume the worst", and says so:

```swift
func cacheHitSuffixContainsMediaPlaceholder(_ remainingTokenIds: [Int]) -> Bool {
    guard hasMediaContent, !remainingTokenIds.isEmpty else { return false }
    guard let mediaTokenIds else { return true }   // <- nil ⇒ always "contains a placeholder"
```

> "`nil` `mediaTokenIds` means the processor did not declare its placeholder IDs, so media prompts stay on the **conservative rollback path**."

`canCaptureHybridStripBoundary` fails the same way — it `guard let mediaTokenIds` and returns `false` without them. So for Apertus 1.5, **every turn carrying an image or audio takes a blanket rollback**: no resume from a cache hit, and no strip boundary captured either. A multi-turn conversation with one image re-prefills the whole prompt on every turn.

Apertus is the worst case for this precisely because of what makes it interesting: 131,072 `<|visual token N|>` plus 4,096 `<|audio token N|>` means one media item is a long placeholder run, so the span you cannot resume past is large.

Measured on the other families when this was fixed for them: a cross-turn VL follow-up went **3.40 s → 0.59 s** once the IDs were declared. Declaring is necessary but not automatically sufficient (Gemma 4 needed a second fix), so please verify with a real second turn rather than assuming.

Concretely: resolve the visual and audio placeholder IDs from the tokenizer and pass them on the `LMInput` the processor returns, the way `Gemma4` does via `MediaTokenIds.resolve(...)`.

## The shared-file edits are fine

**`ReasoningBudget.swift`** — two string literals appended to `closeTokenCandidates` / `openTokenCandidates`. Safe for every other family because the lists are resolved against the bundle's own vocab ("Resolve the open-token ids that exist in this bundle's vocab"), and no non-Apertus bundle has `<|inner_prefix|>` / `<|inner_suffix|>`. Appending rather than inserting also keeps any order sensitivity untouched. The comment explaining WHY the open marker is needed — the template primes neither tag, so the model emits it as its first generated token — is the part that makes this reviewable, thanks.

**`ReasoningParser.swift`** — additive and correctly scoped. I checked the thing that would break it: there is no earlier bare `apertus` branch anywhere in that file, so the `hasPrefix("apertus1p5")` insert cannot be shadowed, and `apertus` (1.0) keeps its "none" stamp per `testPlainFamiliesGetNone`. Citing the measured before-state (`text=3998ch reasoning=0ch`, answer buried 3.7 kB deep) is exactly the right evidence for a stamp change.

## Smaller notes

- `startInReasoning: false` with `stripStrayTags: true` matches the "model emits the opener itself" story, and 3/3 balanced open/close on gsm8k is a reasonable basis for not carrying the unclosed-block hazard the Hunyuan entry documents. Worth saying out loud that 3 samples bounds how strongly that generalises.
- Since #300 is being split, this PR no longer needs to wait on it — the `ReasoningBudget`/`ReasoningParser` additions here are independent of the policy work.

Happy to re-review once the media IDs are in.

rcfa:
`mediaTokenIds` is declared now, and the branch is rebased on current `main`. But the verification you
asked for came back differently than expected, so here is what is actually true.

### The declaration

Both `LMInput` sites in `Apertus1p5Processor.prepare`, via `MediaTokenIds.resolve(tokenizer:tokens:)`,
mirroring Gemma 4.

The non-obvious part is WHICH ids, and it nearly went the wrong way. `Apertus1p5.prepare` substitutes each
`<|image|>` for a discrete visual code (`imageTokenOffset + code`, 131272..<262344 on the 8B bundle) and
expands `<|audio|>` into a code run, so the sequence the MODEL sees contains no placeholders at all. The
code ranges therefore look like the right answer. They are not: the cache compares against
`promptTokenIds`, the PROCESSOR's pre-splice sequence, because `requiresPostPrepareCacheKey` is
`video?.embeddingTokenCount != nil` and this family has no video path. Declaring the codes would name ids
that never appear in the compared sequence, so no rollback would ever fire — the unsafe direction.
Confirmed `<|image|>` = 131079 and `<|audio|>` = 131085 against the bundle's `added_tokens`, matching
`config.imageTokenId`/`audioTokenId`, which is also what the splice keys on.

This also updates `VLMediaTokenDeclarationReachabilityTests` — your guard caught that adding a declaring
processor without updating that list fails, which is exactly what it is for.

### The measurement did not reproduce, and the reason is a separate bug

I could not show the 3.40 → 0.59 you measured on the other families. Chasing why turned up something
bigger, filed as #309: **a `ChatSession` with caching enabled has no conversation memory at all.**

The session is append-only — after turn 1 it submits only the new message and relies on the live KV cache.
The coordinator keys on the full prompt, so from turn 2 it can never match (`MISS all tiers`, every turn),
and `populatedCacheRequiresResetAfterCoordinatorMiss` reads that miss as proof the caller's state is
unverified, resets the cache and "full-prefills" `input` — one message. Reset without recovery, every
turn. Reproduced on Apertus and on LFM2.5-VL, so not ours and not media-specific.

That is why no probe here could ever show a resume: the conversation a resume would restore was being
destroyed first.

Two things follow for this PR:

- **With caching off, the append-only path already avoids the re-encode.** Turn 2 costs 0.70 s against
  turn 1's 13.7 s and recalls the image. So the declaration does not speed that path up — with no
  coordinator there is no restore and the guard's caller never runs. Its value is on the disk-restore
  path.
- **The disk-restore benefit is measurable only once #309 lands.** I would rather say that than quote your
  number as if I had reproduced it.

Happy to re-measure and post the figure here as soon as #309 is resolved, whichever way you take it.

PR #276, opened 2026-08-19
DFlash 2 block-diffusion speculative decoding: 1.36-1.65x lossless decode for Qwen3.8-27B

jjang-ai:
## What this is

DFlash 2 block-diffusion speculative decoding ([z-lab/dflash](https://github.com/z-lab/dflash), [blog](https://inco.ai/blog/dflash2/)), ported from the authors' own MLX reference (`dflash/model_mlx.py`) and wired end-to-end: drafter model + loader + candidate-path selector, `.dflash2` draft strategy, host settings (drafter folder path on `VMLXServerMTPSettings`, resolved ahead of native MTP — a selected drafter **supersedes** the model's own head), dispatch at all four entry points, hybrid rollback, and SSD prefix-cache integration. Full design doc in `Libraries/MLXLMCommon/SpecDec/DFLASH2.md`.

## Measured (M5 Max 128 GB, greedy, 192-tok, interleaved rounds, free-mem logged per leg)

| target | baseline | dflash2 b8 | b6 | b4 |
|---|---|---|---|---|
| Qwen3.8-27B bf16 (52 GB) | 7.6 tok/s | **1.65×** (accLen 5.65) | — | — |
| Qwen3.8-27B JANG_6D (24 GB) | 14.3 tok/s | 1.23× | 1.32× | **1.53×** |
| Qwen3.8-27B JANG_4D (17 GB) | 16.7 tok/s | 1.13× | 1.30× | **1.36×** |

Recommend block 4–6 for quantized targets, 8 (checkpoint default) for bf16 — verify cost grows faster with row count than acceptance does on quants (replicates z-lab/dflash#151).

## The three fixes that made quants a win (first implementation was 0.62× on 4D)

1. **One-shot lazy rollback** (`input_capture` policy mode, `DFlash2VerifyRollbackModel`) instead of per-prefix state recording. Recording re-ran the GDN scan once per prefix **with an eval per record**: 48 layers × 7 prefixes = 336 extra launches and 336 graph flushes per cycle, ~70% of a verify that cost 5.1× a decode step. Now: layers stash verify-input *references* (zero cost), a rejection replays only the accepted rows — one kernel per layer — audited **bit-exact** vs chained decode (`VMLX_DFLASH2_ROLLBACK_AUDIT=1`, maxdiff 0.0). Kill switch: `VMLX_DFLASH2_INPUT_CAPTURE=0` falls back to the recording path.
2. **One host sync per cycle**: verify input built on-graph from the drafter output (no draft→host round trip), greedy argmax folded into the verify flush, `asyncEval` dispatch.
3. **Block-size guidance** per the table above.

⚠️ Trap documented for future work: `ArraysCache`'s subscript setter **mutates the stored MLXArray in place** (`_updateInternal`). A bare reference to `cache[1]` becomes the post-block state as soon as the forward commits. The stash pins values with a `* 1` node; without it the replay double-applied rows and diverged deterministically while passing every state audit.

## Correctness

- **Losslessness**: greedy output byte-identical to the plain iterator on bf16 (583ch and 160-tok runs) and JANG_6D **through `BatchEngine.generate`** — the app chat route (354ch ≡ 354ch, accLen 5.65). On quantized targets a greedy near-tie can legitimately flip (matmul reduction order depends on row count; plain decode itself flipped one across builds with no plain-path diffs) — quantized tests assert long shared prefix + correctness, bf16 keeps strict equality (`VMLX_DFLASH2_EXPECT_EXACT=1`).
- **Reference parity**: gated against the reference's *own* bf16↔fp32 disagreement (8.7% hidden, 4-of-7 path tokens) rather than a hand-picked epsilon; conv bit-exact, attention within one bf16 ulp; exact path agreement wherever both reference precisions agreed.
- **SSD prefix reuse**: hybrids store only the generation-suffix-stripped boundary captured during prefill (TokenIterator's canonical policy). Before: turn 2 re-prefilled 14/14 tokens; after: only the new suffix. Reasoning-effort scopes isolated (modelKey salt).
- **Dispatch reachability**: source-reading tests pin all four entry points after the first live run showed a strategy wired only into `Evaluate.generate` is configured, persisted, and silently inert in the app (`BatchEngine.generate` is the chat route).
- **Safety of the inert path**: every change is gated — no drafter selected ⇒ byte-identical behavior. Identity penalties (`repetition_penalty: 1.0`, the Qwen3.8 bundle default) no longer disqualify; genuinely unservable sampling falls through to plain decode instead of erroring.

## Tests

| suite | needs | result |
|---|---|---|
| DFlash2DrafterSelectionTests (13) | none | ✅ |
| DFlash2DispatchReachabilityTests (5) | none | ✅ |
| DFlash2PrefixCacheReuseTests (3) | none | ✅ |
| DFlash2ReferenceParityTests / StageProbeTests | drafter + golden | ✅ (env-gated) |
| DFlash2LosslessSmokeTests (3) | target + drafter | ✅ (env-gated) |
| DFlash2SpeedSweepTests / QuantizedVerifyShapeBenchTests | target / env | ✅ (env-gated) |
| MLXLMCommonFocusedTests (57) | none | ✅ |
| CmlxTests | none | ✅ |
| MLXTests (527) | none | 526 ✅ / 1 pre-existing local red (`TransformTests/testCompiledRandom`, upstream `Source/MLX` compiled-RNG test; this diff touches no `Source/` code) |

## Next (out of scope here)

Split-K verify-shape QMM kernels for M∈2…9 (oMLX #2113 measured lm_head 2.6–3.3×) are the path from ~1.4–1.5× to ~2×+ on quants; vendored MLX 0.31.1 qmm at M=8 is also 15–30% behind 0.32.1 at these shapes.


jjang-ai:
Second commit (`b074cb07`): the full strategy matrix, reasoning on/off assertions, and the cross-process SSD proof — plus one real bug they caught.

## Strategy matrix — Qwen3.8-27B JANG_4D, one load, greedy, round 2 of 2

| | plain | native MTP d1 | dflash2 b8 | dflash2 b4 |
|---|---|---|---|---|
| **reasoning ON** (160 tok) | 16.8 tok/s | 15.9 (0.95×) | 29.2 (**1.74×**) | 29.9 (**1.79×**) |
| **reasoning OFF** (94 tok) | 16.5 tok/s | 20.1 (1.22×) | 21.8 (1.32×) | 25.5 (**1.54×**) |

DFlash 2 beats native MTP in every cell — largest exactly where turns are longest (reasoning text drafts at 1.74–1.79×). Native MTP d1 is a *loss* with thinking on. Reasoning behaviour is asserted: `enable_thinking=true` must populate the think channel, `=false` must answer directly.

## Long-term SSD reuse — across processes

Store a ~1.4k-token turn in process A, kill it, restore in fresh process B (RAM tiers empty — a hit can only be disk):

- 526 MB persisted at the stripped boundary
- fresh-process fetch: `detail=disk`, matched **1375 / 1413** tokens
- warm prefill+first-token **0.37 s vs 1.94 s cold — 5.2×**

## Bug found by the SSD test

`textOnlyForwardCapturing` forced `resolvedPositionIds`, which with no in-process media state builds its position table from the chunk's own rows **starting at zero** — ignoring the restored cache offset — then the next chunk slices past the short table (broadcast abort at offset 1375 / chunk 33; positions silently wrong even before the crash). Positions now resolve only when `ropeDeltas` exists; otherwise attention derives them from the cache offset like the plain forward. JANG_6D byte-equality through the BatchEngine path re-verified after the fix.

jjang-ai:
## Compiled decode for the hybrid target (commit 3)

The remaining plain-decode gap vs python mlx-lm (17.3 vs 24.6 tok/s on the identical JANG_4D bundle) was 19 ms/token of per-token CPU graph rebuild that never overlapped the GPU. Commit 3 makes the generic compiled-decode path work for hybrid GDN families and fixes every trace-illegal host read in the Qwen35 forward.

### What blocked it
| blocker | fix |
|---|---|
| `setupCompiledDecode` refused non-promotable caches | `MambaCache` passes through unpromoted — fixed-shape state, `innerState()`-tracked, `_updateInternal` in-place updates |
| `resolvedPositionIds` read `faCache.offset` (host `.item()`) → fatal in trace, and would have baked the offset as a constant | mrope positions built from the cache's graph-visible `offsetArray` |
| attention `kvSeqLen` host read + mask slice vs the full static buffer | slice skipped for Compilable caches (mask already spans the buffer) |
| nested `compile()` micrographs (sigmoid-multiply / swiglu / compute_g) illegal inside the outer trace | plain body while tracing — ops fuse into the outer graph (same pattern as SwitchLayers / Gemma4Text) |
| `MambaCache.offset` advanced in Swift inside the forward — replays never run Swift | iterator rewrites recurrent offsets from a step counter per compiled call |
| fixed 4096 buffer could be outgrown silently | sized from `promptOffset + maxTokens`; explicit `compiledMaxCacheLength` wins |

### Measured (Qwen3.8-27B-JANG_4D, greedy, 192 tok, interleaved rounds, hot round reported)
| arm | plain build | compiled build |
|---|---|---|
| baseline | 17.32 tok/s | **23.13 tok/s (+33%)** |
| dflash2-b4 | 22.84 | 22.89 (unaffected — its own iterator) |

**Correctness:** the compiled baseline's 884-char greedy text is **byte-identical** to plain decode (`VMLX_SWEEP_DUMP` diff, two separate invocations). Strategy matrix (think ON/OFF, all four arms), lossless smoke 3/3, and cross-process SSD restore (disk hit 1375/1413, 5.8× warm prefill) all pass unchanged.

Strategy picture on this bundle: dflash2-b4 think-ON remains the fastest arm at 29–32 tok/s; compiled decode lifts the no-drafter floor from 17.3 → 23.1.

Gating unchanged: `enableCompiledDecode` + `VMLX_ENABLE_UNSAFE_COMPILE=1` (osaurus ships compile off by default).

jjang-ai:
## Commits 4-5: what actually limits speculation here

Chasing the gap to published dflash/oMLX numbers produced one large finding and one honest non-ship.

### Block size is the dominant knob — and 8 was the wrong default

dFlash-2 has no "depth"; block size is the equivalent. Measured on Qwen3.8-27B-JANG_4D (greedy, 160 tok, **settled** box — a loaded one swings these 30%):

| prompt | effort | plain | b4 | b8 | b15 |
|---|---|---|---|---|---|
| general | xhigh | 17.2 | **32.8** | 21.4 | 23.9 |
| general | medium | 17.3 | **29.0** | 22.4 | 25.5 |
| general | off | 17.3 | **29.9** | 21.8 | 22.5 |
| coding | xhigh | 16.9 | 34.5 | 27.1 | **34.2** |
| coding | medium | 17.2 | 35.6 | 33.3 | **43.3** |
| coding | off | 17.2 | 34.0 | 30.5 | **36.7** |

**b8 — the drafter config's own default — is the worst of the three in 6 of 6 cells.** Acceptance length rises monotonically with the block (coding/medium 5.71 at b8 → 7.27 at b15) because the drafter *saturates* a narrow block, while quantized matmul cost grows sublinearly over the same range. Prose wants b4, code wants b15. Reasoning effort matters too: `medium` drafts better than `xhigh` on code (accLen 5.71 vs 4.71).

New `DFlash2AcceptanceMatrixTests` measures the whole surface (domain × effort × block) with a plain reference per cell.

### Adaptive block sizing — works, opt-in

Probes each ladder size, settles on measured tokens-per-verify-second, remembers it process-wide. Lands within 3-4% of the per-domain oracle on code and **beats every fixed size on prose**. Opt-in (`VMLX_DFLASH2_ADAPTIVE_BLOCK=1`) because a mid-turn size change still interacts badly with the compiled staged verify — both are shaped by the block length.

Trap worth recording: an acceptance-**rate** threshold looks like the natural control signal and is wrong. The rate falls mechanically as blocks widen, so fixed thresholds equilibrate mid-ladder — 34.5 tok/s where fixed b15 gave 41.9. Measure throughput; don't infer it.

### Compiled verify → opt-in (was on by default in commit 3)

Still proven token-identical and faster, but bisection found a **process segfault at exit** when it is enabled: the compiled traces capture the cache array and `DFlash2TokenIterator` is a struct, so copies retain traces independently and releasing one copy's table doesn't free them. Clean with `VMLX_DFLASH2_COMPILED_VERIFY=0`, signal 11 with it on, tests pass either way. Rather than ship a crash, it is now `=1` opt-in with a repro and a proposed fix (move the trace table into a reference-typed box owned by the run) documented in DFLASH2.md.

Default path re-verified green: lossless smoke 3/3, dispatch + selection 18/18, no crash.

jjang-ai:
## Pre-merge audit: what this touches outside DFlash 2

Nine files land outside `SpecDec/` and tests. Each one is inert unless you opt in:

| file | when it changes behaviour |
|---|---|
| `Evaluate.swift` | only under `enableCompiledDecode` + `VMLX_ENABLE_UNSAFE_COMPILE=1` |
| `Qwen35.swift` (VLM + LLM), `Qwen3Next`, `GatedDelta` | nested-compile fallbacks fire only inside a compile trace; graph-offset positions only when the cache is a Compilable one (`nil` for every ordinary cache) |
| `KVCache.swift` | `MambaCache` gains staging slots — `nil` unless a staged verify runs, so `innerState()` is unchanged |
| `BatchEngine`, `ServerRuntimeSettings` | DFlash 2 dispatch + selection, all gated below |
| `CompilableKVCache.swift` | see the bug found below |

**DFlash 2 is default-OFF, proven against real bundles** (`DFlash2AppliesOnlyToItsTargetTests`): no drafter selected → nothing selected for any bundle; a selected drafter is refused for the Ornith 1.5 line with a readable reason (*"Drafter reads layer 61 of its target; this model has 32 layers"*), and the strategy falls through to that bundle's own path. Ornith 1.5 is the deliberate near-miss — same `qwen3_5` hybrid family, so only the config check stands between it and a mis-attached drafter. A positive control proves the drafter's own target is still accepted.

Everything experimental is opt-in: compiled verify (`=1`), verify prefetch (`=1`, measured as a non-win here), adaptive block size (`=1`).

## A real bug this audit caught

`CompilableKVCache.copy()` shared its array references, so a "snapshot" followed the live cache — and `copy()` is what the prefix-cache store path uses. The keys/values half **predates this branch**; the branch would have extended it to the offset, because it moved the offset setter and `trim()` to in-place updates (a rebind detaches compiled functions tracking the cache).

Fixed by pinning all three with a `* 1` node, the same thing `ArraysCache.copy()` already documents. `CompilableKVCacheSnapshotTests` was checked against the unfixed code and fails 3/3 there: offset follows a write (40 vs 12), follows a trim (20 vs 30), keys drift (0.0 vs 64.0).

## Cross-family regression evidence

`Ornith15EnablementTests` runs a **different** hybrid family through the modified Qwen35/MambaCache code with no DFlash 2 and no compile: reasoning OFF 38.2 tok/s, reasoning ON 39.1, vision correct on both opposite colours, multiturn context kept. 3/3.

Suites green at merge: gating + selection + dispatch + snapshot 24/24, lossless smoke 3/3, Ornith 3/3.

PR #278, opened 2026-08-20
Native MTP: staged verify, aligned head cache, honest depth economics

jjang-ai:
## What

Takes Qwen3.8-27B native MTP from **slower than plain decode** to **1.6-1.8x, byte-identical output**, at every depth. Three fixes:

### 1. Staged verify wired into the MTP verify cycle
The batched hybrid verifier ran lazy repair: every partial accept cost a checkpoint restore + a **full-model replay forward**, and the sequential safety warmup priced 16 cycles at one full forward **per token**. Phase timers caught a depth-3 run spending 93% of wall inside per-token verify forwards (`verifyMainForwards=93` for 35 cycles — the sequential fallback, running the entire generation).

Staged verify (the DFlash 2 rollback machinery, already proven on the same model class) commits every cycle from the cache staging slots: `replayMainSec=0.000`, `rollbackRepair=0`, `mtpCacheRefresh=0`. It is the automatic mode for capable hybrids under greedy sampling. Explicit verifier modes are honoured unchanged. The staged mode string is passed task-locally around the verify forward ONLY — prefill, seed, sequential, and replay forwards can never stage.

### 2. Aligned head cache
Every confirmed (backbone hidden, token) pair is committed through the MTP head in the same forward that drafts the next block. Depth chaining on the single-layer head measures **1.88 / 2.54 / 3.13 committed tokens per verify** at depths 1/2/3.

### 3. Depth economics recalibrated
The D2 cap and 0.85/0.75/0.5 adaptive floors were tuned when a rejection cost a replay forward. Staged break-even per-draft acceptance is 0.39/0.44/0.52 → cap default 3, staged floors 0.60/0.50/0.40. The measurement hatch (`VMLX_NATIVE_MTP_DISABLE_ADAPTIVE=1`) now completes warmup by cycle count instead of silently forcing the sequential verifier forever (which fabricated "MTP got slower" runs).

## Measured (Qwen3.8-27B-JANG_4D, temp 0, hot rounds, PhysMem logged)

| arm | tok/s | vs plain | output |
|---|---|---|---|
| plain | 16.5 | — | — |
| mtp-d1 | 23.2 | 1.42x | byte-identical |
| mtp-d2 | 27.5 | 1.61x | byte-identical |
| mtp-d3 | **28.4** | 1.60x | byte-identical |

Thinking ON: d2/d3 = 1.80x/1.79x (byte-identical, 634/634). Steady state with the adaptive controller live (no measurement env): d3 25.8-28.1 tok/s with one 3→2 downshift on a low-acceptance prompt — the controller working as designed.

## Proof

- Cross-version probe on Ornith-1.5-9B: `COMBINED b4b9320481f8233f` — **exact match with main** (non-MTP decode untouched)
- MTP runtime focused suite: 82/82
- All DFlash2 suites green (dispatch reachability, drafter selection, reference parity, stage probe, applies-only-to-its-target)
- Every matrix cell asserts shared prefix vs plain decode; all cells full-length identical

## Also

- `VMLX_MTP_PHASE_TIMERS=1`: splits `targetVerifySec` into graph build vs GPU wait (the instrument that found the sequential lockout)
- Small-M QMM scaling bench (`VMLX_QMM_BENCH=1`): M=4 costs 1.29x M=1 on FFN-shaped 4-bit gs64, 1.54x on the 248320-row LM head, non-monotonic kernel switch at M=16 — the data for the next round (compiled/fused verify)
- Strategy matrix test: arm/think/prompt filters for subset runs

### DIFF
--- a/Libraries/MLXLMCommon/Cache/TQDiskSerializer.swift
+++ b/Libraries/MLXLMCommon/Cache/TQDiskSerializer.swift
@@ -158,11 +158,20 @@ public enum TQDiskSerializer {
     /// Read a 1-element Int32 metadata array. Tolerates both 0-dim
     /// (legacy entries written before the round-trip fix) and 1D
     /// shape-[1] (current).
-    private static func readMetaInt32(_ arr: MLXArray) -> Int32 {
+    ///
+    /// Returns nil for anything else: a disk entry is untrusted input,
+    /// and a truncated or mixed-up key set must read as a cache MISS,
+    /// never crash the host. Observed live 2026-08-20: a DFlash 2 restore
+    /// of a stored hybrid entry found a non-scalar under a metadata key
+    /// and the unconditional `item()` precondition took the whole app
+    /// down inside `DFlash2TokenIterator.init`.
+    private static func readMetaInt32(_ arr: MLXArray) -> Int32? {
         if arr.shape.isEmpty {
             // Legacy 0-dim — Int32-typed scalar.
+            guard arr.dtype == .int32 else { return nil }
             return arr.item(Int32.self)
         }
+        guard arr.ndim == 1, arr.dim(0) == 1, arr.dtype == .int32 else { return nil }
         return arr[0].item(Int32.self)
     }
 
@@ -208,8 +217,8 @@ public enum TQDiskSerializer {
     /// (pre-v2 entries). Returns `0` for dicts that don't come from this
     /// module at all.
     public static func formatVersion(of arrays: [String: MLXArray]) -> Int32 {
-        if let v = arrays[formatVersionKey] {
-            return readMetaInt32(v)
+        if let v = arrays[formatVersionKey], let version = readMetaInt32(v) {
+            return version
         }
         if arrays.keys.contains(legacyMarkerKey) {
             return 1
@@ -995,8 +1004,10 @@ public enum TQDiskSerializer {
     /// - Returns: The ordered `[MLXArray]` captured at serialize time, or
     ///   `nil` if the dict has no SSM entries.
     public static func ssmStates(from arrays: [String: MLXArray]) -> [MLXArray]? {
-        guard let countArr = arrays["__ssm_count__"] else { return nil }
-        let count = Int(readMetaInt32(countArr))
+        guard let countArr = arrays["__ssm_count__"],
+            let countValue = readMetaInt32(countArr)
+        else { return nil }
+        let count = Int(countValue)
         guard count > 0 else { return nil }
         var out: [MLXArray] = []
         out.reserveCapacity(count)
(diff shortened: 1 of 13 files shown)


jjang-ai:
Status vs the merge bar (match the oMLX MTP ladder under its own conditions — 320 generated tokens, temp 0, thinking off, k=3, single stream; their M4 Max numbers: baseline 24.9, prose 48.0, code 65.5):

| leg (settled box, PhysMem logged) | this branch | oMLX bar |
|---|---|---|
| plain eager | 15.6–17.1 | 24.9 |
| mtp-d3 prose | 19.3–23.3 | 48.0 |
| mtp-d3 code | 24.9 (1.59×) | 65.5 |

All cells byte-identical to plain decode (prose 1398/1398 chars, code 1209/1209). Compiled staged verify added in 615aba82 measured neutral-to-slightly-positive under honest ABA (the cycle is weight-streaming-bound; an initial +20% reading was an A-then-B ordering artifact) — kept as opt-in scaffolding for the fused cycle.

NOT ready to merge. Remaining ladder, in order of measured leverage:
1. **Small-M quantized-matmul kernel** (vendored fork): M=4 verify rows currently pay 1.29× (FFN shapes) to 1.54× (LM head) the M=1 weight stream — bench committed as `VMLX_QMM_BENCH=1`. A multi-row qmv that streams weights once for M≤16 takes the verify cycle from ~2.1 to ~1.1 plain steps.
2. Prose acceptance at depth 3 (currently ~0.72 per-draft vs their 0.94).
3. Draft-chain fusion (3 sequential head forwards per cycle).

jjang-ai:
Correction + progress (5b2a5266): the earlier "compiled verify is neutral" reading was wrong for a mundane reason — **it never ran**. `HardwareInfo.isCompiledDecodeSupported` is false without `VMLX_ENABLE_UNSAFE_COMPILE=1` (the #1173 model-switch gate), so every compiled leg silently measured the eager path. A trace-build probe proved it (zero trace builds without the env; exactly one per shape with it), and an immediate `asyncEval` dispatch of the replayed graph turned out to be worth another ~15% (draining the replay in one lump at the readback was the hidden cost).

Updated settled table (320 tokens, temp 0, thinking off, byte-identical everywhere):

| leg | before | now | oMLX bar |
|---|---|---|---|
| plain eager | 17.5 | 17.5 | 24.9 |
| mtp-d3 prose | 19.3–23.3 | **24.4–25.1** (1.43×) | 48.0 |
| mtp-d3 code | 24.9 | **30.5–31.5** (1.80×) | 65.5 |
| mtp-d4 code (hot) | — | **34.8–35.9** | 65.5 |

Remaining gap ~1.4–1.9×, attribution: (1) prose acceptance 2.52 committed/verify vs their calibrated ~3.4+ — likely needs an MTPLX-style calibrated head from the quant pipeline, not runtime work; (2) draft chain still eager (3 sequential head forwards ≈ 15ms/cycle); (3) M=4/5 verify QMM tax (multi-row kernel attempt measured a loss — L2 already amortizes; committed opt-in for future tuning). Note compiled verify currently requires `VMLX_ENABLE_UNSAFE_COMPILE=1` + `VMLX_MTP_COMPILED_VERIFY=1`, same gating story as plain compiled decode.

PR #284, opened 2026-08-20
Stop a prose first turn from locking MTP off for the whole model residency

jjang-ai:
## Problem

`NativeMTPHybridWarmupMemo`'s contract (its own doc comment) is that it stores *a property of the model, not the prompt*, and that acceptance-ratio fallbacks stay per-request because they are content-dependent. But the warmup verdict it memoizes IS an acceptance criterion — 0.55 × depth over the first 16 verify cycles.

Acceptance is content. Measured on the same loaded Qwen3.8-27B-JANG_4D at temp 0:

| prompt | warmup averageAccepted | floor (d3) |
|---|---|---|
| prose | **1.00** | 1.65 |
| code | **3.04** | 1.65 |

So a prose-grade *first* turn memoized a permanent negative verdict, and every later **code** turn — where MTP speculates at 1.7× — silently ran plain AR (`verifyCalls=0`, `adaptiveFallback=hybrid_warmup_memo`) until the model was reloaded. In a chat session that starts with a greeting and moves to code, MTP is off for the whole residency. Observed live in the strategy matrix: round 1 prose recorded the memo, round 2 ran zero verify calls.

## Fix

Memoize a failure only when it is **catastrophic**: below even the depth-1 floor (0.55), which no depth could ever clear — that is the broken/mismatched-head case the memo was built for (its census: 84% of requests burning the 16-cycle probe to land in AR again). A miss above that line falls back for THIS request and re-probes on the next one.

Passing verdicts still memoize; the broken-head case still memoizes; only the content-dependent band re-probes.

## Evidence (DFlash2StrategyMatrixTests, prose, temp 0, 320 tok)

Before:
```
r1 mtp-d3: verifyCalls=16 ... adaptiveFallback=hybrid_warmup_avg_accept=1.00   (records memo=false)
r2 mtp-d3: verifyCalls=0  ... adaptiveFallback=hybrid_warmup_memo              (locked out)
```
After:
```
r1 mtp-d3: verifyCalls=16 ... adaptiveFallback=hybrid_warmup_avg_accept=1.00
r2 mtp-d3: verifyCalls=16 ... adaptiveFallback=hybrid_warmup_avg_accept=1.00   (re-probed)
```
Output byte-identical to plain decode both rounds (sharedPrefix 1398/1398). Re-probe overhead on a turn that stays AR: 0.94× vs 0.91× — ~3% of a prose turn, paid only on turns that fail warmup.

## Tests

`NativeMTPWarmupMemoScopeTests` (focused suite) pins the boundary: prose-grade misses (1.00, 0.80, 0.55) are NOT memoizable; catastrophic misses (0.0, 0.30, 0.54) are; the memo still records per model instance for its real case.

Live in-app prose→code proof to follow in a comment before merge.

### DIFF
--- a/Libraries/MLXLMCommon/SpecDec/NativeMTPTokenIterator.swift
+++ b/Libraries/MLXLMCommon/SpecDec/NativeMTPTokenIterator.swift
@@ -1650,7 +1650,19 @@ struct NativeMTPTokenIterator: TokenIteratorProtocol {
                 hybridSafetyWarmupComplete = true
                 NativeMTPHybridWarmupMemo.record(true, for: model)
             } else {
-                NativeMTPHybridWarmupMemo.record(false, for: model)
+                // The memo's contract is "a property of the model, not the
+                // prompt" — but this floor is an acceptance criterion, and
+                // acceptance is content. Memoize the failure only when it is
+                // catastrophic (below even the depth-1 floor), which is the
+                // broken/mismatched-head case the memo exists for. A miss
+                // above that line means THIS prompt drafted badly (prose
+                // warms up around 1.0, code above 3.0 on the same model);
+                // fall back for this request and let the next one re-probe,
+                // or a prose first turn locks every later code turn out of
+                // MTP for the whole residency.
+                if Self.warmupFailureIsModelProperty(averageAccepted: averageAccepted) {
+                    NativeMTPHybridWarmupMemo.record(false, for: model)
+                }
                 enableAutoregressiveFallback(
                     reason: String(
                         format: "hybrid_warmup_avg_accept=%.2f",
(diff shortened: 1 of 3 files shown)


jjang-ai:
## Merge evidence

**Focused tests** (`NativeMTPWarmupMemoScopeTests`, swift-testing): 3/3 pass — prose-grade misses (1.00/0.80/0.55) not memoizable, catastrophic misses (0.0/0.30/0.54) memoizable, per-instance memo behavior intact.

**Live matrix, real bundle (Qwen3.8-27B-JANG_4D, temp 0, prose, 320 tok)** — the exact lockout and its removal:

Before (pinned 16486d6e):
```
r1 mtp-d3: verifyCalls=16 arFallbackTokens=286 adaptiveFallback=hybrid_warmup_avg_accept=1.00
r2 mtp-d3: verifyCalls=0  arFallbackTokens=318 adaptiveFallback=hybrid_warmup_memo   ← locked out
```
After (this branch):
```
r1 mtp-d3: verifyCalls=16 arFallbackTokens=286 adaptiveFallback=hybrid_warmup_avg_accept=1.00
r2 mtp-d3: verifyCalls=16 arFallbackTokens=286 adaptiveFallback=hybrid_warmup_avg_accept=1.00  ← re-probes
```
Byte-identical to plain both rounds (sharedPrefix 1398/1398); re-probe overhead ~3% on a turn that stays AR (0.94× vs 0.91×).

**Host-app surface check** (osaurus proof app, isolated root, background AX drive): the chat UI *by pinned policy* marks every turn `samplingParametersAreImplicit` and therefore never engages native MTP regardless of Mode/temperature — verified live (warm greedy turn, agent temperature 0, JANG_4D loaded: no `[NativeMTP]` engagement, 29.3 tok/s plain AR; the policy is test-enshrined in RuntimePolicySourceTests "marks implicit sampling without forcing native MTP"). Native MTP's in-app surface is API/agent requests with explicit greedy sampling — the same engine entry the matrix drives, so the matrix legs above are the end-to-end proof for the surface this code actually serves. Chat-side policy is out of scope here and recorded upstream.

Also measured while gated on this path: compiled verify (`VMLX_MTP_COMPILED_VERIFY=1` + `VMLX_ENABLE_UNSAFE_COMPILE=1`) lifts mtp-d3 from 28–29 to **31.5/30.4 tok/s (1.90× over plain), byte-identical** — the next lever once osaurus #1173's compile gate can be revisited.

PR #294, opened 2026-08-22
Presence and frequency penalties apply to generated tokens only

rcfa:
**Correctness fix.** `presence_penalty` and `frequency_penalty` currently apply to the prompt as well as the
output, which is not what either parameter means. The change is two function bodies becoming empty.

`PresencePenaltyContext` and `FrequencyPenaltyContext` seed their ring with the prompt:

```swift
mutating public func prompt(_ prompt: MLXArray) {
    ring.loadPrompt(prompt)
}
```

so both penalise tokens **the model never chose**. `presence_penalty` and `frequency_penalty` are OpenAI's parameters, and vLLM — the implementation most published values are actually measured against — computes both from the **output** tokens alone (`vllm/model_executor/layers/sampler.py`, `apply_penalties`):

```python
output_bin_counts, output_mask = get_token_bin_counts_and_mask(
    output_tokens_tensor, vocab_size, num_seqs)
logits -= frequency_penalties.unsqueeze(dim=1) * output_bin_counts
logits -= presence_penalties.unsqueeze(dim=1) * output_mask
```

(vLLM's `prompt_tokens_tensor` feeds `repetition_penalties` only — the same split this PR makes.) HuggingFace has no `presence_penalty`/`frequency_penalty` at all, so vLLM/OpenAI is the whole of the published definition.

The divergence is not subtle at the boundary. `presenceContextSize` defaults to 20, and prompts are longer than 20 tokens essentially always, so at the **first decode step the window is 100% prompt**: every opening token is penalised for the *user's* wording, and only after 20 generated tokens does the window contain what the model actually said. `frequency_penalty` is worse in kind rather than degree — the penalty is scaled by a **count**, so a word the user repeated three times in the prompt is suppressed three times as hard.

This matters because real model cards ship these values. Qwen3 publishes `presence_penalty: 1.5` for non-thinking operation, and the Qwen3.6/3.8 bundles carry it in their sampling modes; under the current code that setting measures a quantity its authors did not specify. A user who copies a published sampling config gets neither the published behaviour nor a documented alternative.

**`RepetitionContext` is deliberately unchanged**, and that is the point of the split rather than an omission. HuggingFace defines `repetition_penalty` over `input_ids` — prompt included — in `RepetitionPenaltyLogitsProcessor`, so seeding it there is correct and matches *its* reference. The three processors share a shape but not a specification.

**Scope:** two functions in one file, each becoming a no-op body. No API change, no new parameter, nothing to migrate. Behaviour is identical wherever the prompt is shorter than the context window and no prompt token recurs across the boundary; it differs exactly where the current behaviour is wrong.

**A separate note, not part of this PR.** vLLM applies these penalties over *every* token generated so far — there is no window, because presence is a property of the whole output. The ring here is a 20-token window by default, so a token penalised at step 100 is forgiven by step 120 and penalised again at 140, which is a different function from the published one on any generation longer than the window. That one needs no patch: `presenceContextSize` is already a caller parameter, so a caller wanting conformance can size the ring to its own generation budget (we do). It may still be worth revisiting the **default** — 20 is a sensible default for `repetitionContextSize`, whose HF reference is also windowed in practice, but it is not a neutral default for a parameter whose reference has no window at all.

### DIFF
--- a/Libraries/MLXLMCommon/Evaluate.swift
+++ b/Libraries/MLXLMCommon/Evaluate.swift
@@ -928,9 +928,23 @@ public struct PresencePenaltyContext: LogitProcessor {
         self.ring = TokenRing(capacity: presenceContextSize)
     }
 
-    mutating public func prompt(_ prompt: MLXArray) {
-        ring.loadPrompt(prompt)
-    }
+    /// DELIBERATELY EMPTY: the presence penalty applies to GENERATED tokens only.
+    ///
+    /// `presence_penalty` is OpenAI's parameter, and vLLM — the reference implementation most
+    /// published values are measured against — computes it from `output_tokens` alone:
+    ///
+    ///     output_bin_counts, output_mask = get_token_bin_counts_and_mask(output_tokens_tensor, …)
+    ///     logits -= presence_penalties.unsqueeze(dim=1) * output_mask
+    ///
+    /// Seeding the ring with the prompt penalises tokens the model never chose. With a long prompt the
+    /// window can be ENTIRELY prompt at the first decode step, so the opening tokens are penalised for
+    /// the user's wording. Qwen publishes `presence_penalty: 1.5` for non-thinking operation against
+    /// that definition; applying it to prompt tokens measures something its authors did not specify.
+    ///
+    /// NOTE the contrast with `RepetitionContext`, which SHOULD keep seeding the prompt: HuggingFace's
+    /// `repetition_penalty` is defined over `input_ids`, prompt included. The parameters differ in
+    /// scope, so they differ here.
+    mutating public func prompt(_ prompt: MLXArray) {}
 
     public func process(logits: MLXArray) -> MLXArray {
         guard let indices = ring.validTokens?.asType(.uint32) else { return logits }
@@ -956,9 +970,10 @@ public struct FrequencyPenaltyContext: LogitProcessor {
         self.ring = TokenRing(capacity: frequencyContextSize)
     }
 
-    mutating public func prompt(_ prompt: MLXArray) {
-        ring.loadPrompt(prompt)
-    }
+    /// DELIBERATELY EMPTY, for the same reason as `PresencePenaltyContext.prompt`: `frequency_penalty`
+    /// is OpenAI's parameter and vLLM counts occurrences in `output_tokens` only. Counting prompt
+    /// occurrences would scale the penalty by how often the USER said a word.
+    mutating public func prompt(_ prompt: MLXArray) {}
 
     public func process(logits: MLXArray) -> MLXArray {
         guard let validTokens = ring.validTokens else { return logits }


jjang-ai (approved):
Checked this against the code rather than the description, and it holds up.

**The reference semantics are stated correctly.** vLLM's `apply_penalties` builds `prompt_mask` and feeds it to `repetition_penalties` only; frequency uses `output_bin_counts` and presence uses `output_mask`. HuggingFace has no equivalent of either parameter, so vLLM/OpenAI is the whole published definition. Leaving `RepetitionContext` seeded from the prompt is right against *its* reference (`RepetitionPenaltyLogitsProcessor` over `input_ids`) — the split is the point, not an omission.

**Emptying the two `prompt()` bodies is structurally safe**, which is the thing worth verifying here. `TokenRing.init` already allocates `buffer = zeros([capacity])` and sets `count = 0`, and `validTokens` returns `nil` while `count == 0`. So an unseeded ring is a well-defined empty ring, not uninitialized state: `process` returns logits untouched until the first `didSample`. There is no path where this leaves a `count`/`buffer` desync — which is the failure this ring has actually had before, per the `Index out of range` note in `validTokens`.

The severity argument is right too: `presenceContextSize` defaults to 20 and prompts are essentially always longer, so at the first decode step the window is 100% prompt. Frequency is worse in kind rather than degree, since the penalty scales with a count.

Approving.

jjang-ai (changes requested):
Correcting my approval above — I built and ran this before merging, and it takes the suite red. The reasoning in the PR is still right; the problem is that the tests asserting the old semantics were not updated.

**Measured, with a control.** Same worktree, same metallib, only the merge differs:

| tree | `swift test --filter SampleTests` |
|---|---|
| `main` @ 3aa7cb36 | Executed 18 tests, **0 failures** |
| main + #294 alone | Executed 18 tests, **10 failures** |
| main + #294 + #296 + #297 | Executed 18 tests, **10 failures** |

#294 alone accounts for all of it. The five tests are:

```
testPresencePenaltyContextPenalizesSeenTokens
testPresencePenaltyContextPenalizesUniqueSeenTokens
testFrequencyPenaltyContextPenalizesByCount
testFrequencyPenaltyContextPenalizesByTokenCount
testGenerateParametersPenaltyProcessorComposesPenaltiesInOrder
```

**They are not finding a bug — they are pinning the behaviour this PR deliberately removes.** Each seeds the ring through the prompt and then asserts the penalty landed on prompt tokens:

```swift
var processor = PresencePenaltyContext(presencePenalty: 0.5, presenceContextSize: 20)
processor.prompt(MLXArray([1, 1, 3]))
…
XCTAssertEqual(values[1], 1.5, accuracy: 1e-6)   // token 1 penalised — from the PROMPT
```

With `prompt()` a no-op that is now `2.0`, correctly. So the tests are the old contract written down, and this PR owns them: a change that redefines what these processors measure has to carry the tests that assert the definition.

What they should assert instead is the new contract, which is strictly more valuable than what they cover today:
- a token seen ONLY in the prompt is **not** penalised (this is the whole PR, and nothing currently covers it),
- a token produced by `didSample` **is** penalised,
- for frequency, a token generated twice is penalised twice while a token repeated twice in the prompt is not penalised at all — that pair is the sharpest statement of the split,
- `RepetitionContext` still penalises from the prompt, so the divergence between the three processors is pinned rather than assumed.

Happy to push that test rewrite if you'd rather not — say the word. Otherwise this is the only thing standing between the PR and a merge; the implementation itself I already checked and it is sound (`TokenRing.init` allocates and `validTokens` returns nil while `count == 0`, so an unseeded ring is well-defined rather than uninitialized).

Same note applies to #296, which inherits the failures through #294.

jjang-ai:
Merged, with the test rewrite following immediately in #301 so main was only red for the gap between the two — verified green on the merge commit itself (`1c17ce83`): `SampleTests` 20 tests, 0 failures.

I wrote the rewrite rather than sending it back, since the reasoning here was already settled and the tests were the only thing outstanding. #301 also carries the negative-context-size guard from the #296 thread.

One thing worth passing on, because it bit me writing the test that pins the divergence: sharing a single logits array across two processors silently compares the wrong thing. `RepetitionContext.process` writes through `logits[0..., indices] = …`, and `MLXArray` is a handle to shared storage, so the second processor sees the first one's output. My presence assertions came back as `1.333`/`2.667` — those were the repetition penalty's numbers, not presence's. Fresh array per processor.

PR #296, opened 2026-08-22
Presence and frequency penalties are unbounded by default

rcfa:
**Follow-up to #294**, and it needs that one first: #294 stops these penalties from being applied to the PROMPT; this one stops them from being applied through a 20-token window.

`presenceContextSize` and `frequencyContextSize` default to 20. vLLM — which defines what these parameters do in practice, HuggingFace having no equivalent — applies both across **every token generated so far**, with no window:

```python
output_bin_counts, output_mask = get_token_bin_counts_and_mask(
    output_tokens_tensor, vocab_size, num_seqs)
logits -= frequency_penalties.unsqueeze(dim=1) * output_bin_counts
logits -= presence_penalties.unsqueeze(dim=1) * output_mask
```

Presence is a property of the whole output: penalise a token once, and the model routes around it. Through a 20-token ring it is instead forgiven at step 120 and penalised again at 140 — a *different function*, on any generation longer than the window, which is all of them.

### What changes

Both become `Int?`, defaulting to `nil` = unbounded. **The ring stays**: a positive value still selects it, for callers who want the cheaper approximation on very long generations and accept that it deviates; `0` still disables. The window does not go away, it stops being the default — a default should express what the parameter means, and a caller who wants something else should say so.

### The unbounded path is not the slow path

The unbounded mode keeps a per-vocabulary counter and updates **one element** per sampled token. Cost is fixed by the vocabulary, not by how long generation runs, so a 32k-token answer costs the same per step as a 32-token one.

For **frequency** it is *strictly cheaper than the code it replaces*. The current windowed implementation allocates a vocab-sized histogram and rebuilds it from the ring on every decode step:

```swift
let ones = MLXArray.ones([validTokens.dim(0)], type: Float32.self)
let histogram = MLXArray.zeros([vocabSize], type: Float32.self)
    .at[validTokens.asType(.int32)].add(ones)
```

The unbounded version keeps that histogram and updates it in place — the O(vocab) subtract is unchanged, the per-step allocation and scatter-add are gone. So conformance and speed point the same way here.

For **presence** it is a wash — a dense fused multiply-subtract over the vocabulary instead of an indexed gather/scatter over ≤20 entries. Not a regression worth worrying about beside a transformer forward pass, but I would rather state it than imply a win that isn't there.

### Why the counter is a class

The vocabulary size is only knowable once logits arrive, and `LogitProcessor.process` is non-mutating, so the accumulator cannot be a stored property of the struct. It is a small `final class` held by both contexts. This is not a new liberty in this file: `TokenRing.buffer` is already an `MLXArray` — itself a handle to shared storage that the non-mutating `process` writes through (`logits[0..., indices] = …`).

`RepetitionContext` is untouched, for the same reason as in #294: HuggingFace's `repetition_penalty` is defined over `input_ids` and is windowed in practice, so 20 remains a sensible default there. The three processors share a shape but not a specification.

### Verified behaviour-preserving

Compared against a build whose ring is sized to the generation budget — a ring that can never evict, so it is already the unbounded semantics, just expressed the expensive way. Same model, same seed, `presence_penalty: 1.5`, `temperature: 0.7`, gsm8k 5-shot:

```
unbounded counts  vs  ring(capacity = maxTokens)   →  generations byte-identical
```

Temperature sampling is the point of that setup: any numerical divergence, including a silent dtype promotion in the new dense subtract, would have moved the sampled tokens and shown up as different text.

With a control, because identical output between two paths that both do nothing proves nothing — the same run at `presence_penalty 0` produces *different* text (114 vs 120 tokens on the same item), confirming the penalty was actually being applied in both arms.

### DIFF
--- a/Libraries/MLXLMCommon/Evaluate.swift
+++ b/Libraries/MLXLMCommon/Evaluate.swift
@@ -928,9 +928,23 @@ public struct PresencePenaltyContext: LogitProcessor {
         self.ring = TokenRing(capacity: presenceContextSize)
     }
 
-    mutating public func prompt(_ prompt: MLXArray) {
-        ring.loadPrompt(prompt)
-    }
+    /// DELIBERATELY EMPTY: the presence penalty applies to GENERATED tokens only.
+    ///
+    /// `presence_penalty` is OpenAI's parameter, and vLLM — the reference implementation most
+    /// published values are measured against — computes it from `output_tokens` alone:
+    ///
+    ///     output_bin_counts, output_mask = get_token_bin_counts_and_mask(output_tokens_tensor, …)
+    ///     logits -= presence_penalties.unsqueeze(dim=1) * output_mask
+    ///
+    /// Seeding the ring with the prompt penalises tokens the model never chose. With a long prompt the
+    /// window can be ENTIRELY prompt at the first decode step, so the opening tokens are penalised for
+    /// the user's wording. Qwen publishes `presence_penalty: 1.5` for non-thinking operation against
+    /// that definition; applying it to prompt tokens measures something its authors did not specify.
+    ///
+    /// NOTE the contrast with `RepetitionContext`, which SHOULD keep seeding the prompt: HuggingFace's
+    /// `repetition_penalty` is defined over `input_ids`, prompt included. The parameters differ in
+    /// scope, so they differ here.
+    mutating public func prompt(_ prompt: MLXArray) {}
 
     public func process(logits: MLXArray) -> MLXArray {
         guard let indices = ring.validTokens?.asType(.uint32) else { return logits }
@@ -956,9 +970,10 @@ public struct FrequencyPenaltyContext: LogitProcessor {
         self.ring = TokenRing(capacity: frequencyContextSize)
     }
 
-    mutating public func prompt(_ prompt: MLXArray) {
-        ring.loadPrompt(prompt)
-    }
+    /// DELIBERATELY EMPTY, for the same reason as `PresencePenaltyContext.prompt`: `frequency_penalty`
+    /// is OpenAI's parameter and vLLM counts occurrences in `output_tokens` only. Counting prompt
+    /// occurrences would scale the penalty by how often the USER said a word.
+    mutating public func prompt(_ prompt: MLXArray) {}
 
     public func process(logits: MLXArray) -> MLXArray {
         guard let validTokens = ring.validTokens else { return logits }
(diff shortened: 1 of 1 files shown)


jjang-ai (commented):
I went looking for the hazard this design implies, and it is not there — but I found a small edge case that is.

**The shared-accumulator hazard is defused, and it is worth recording why so the next reader doesn't re-derive it.** A `final class` counter held by a value-type context means every copy shares one accumulator, which would be a real bug two ways: a reused `GenerateParameters` carrying counts into the next generation (worse after #294, since the now-empty `prompt()` was the natural reset point), and a batched engine sharing one processor across concurrent sequences.

Neither can happen. `processor()` is a function that constructs all contexts fresh on every call, and every call site builds one per iterator:

```
Evaluate.swift:1482, :1566, :3130   -> per TokenIterator
BatchScheduler.swift:190            -> per REQUEST (request.parameters.processor())
NativeMTPTokenIterator.swift:407    -> per MTP iterator
```

`BatchScheduler` calling it per request is the decisive one: each sequence in a batch gets its own accumulator, so there is nothing to race on and no state to reset.

**The edge case: a negative context size now silently inverts meaning.**

The guard moved from `presenceContextSize > 0` to `presenceContextSize != 0`, and negatives fall through the gap:

| value | before | after |
|---|---|---|
| `20` | 20-token ring | 20-token ring |
| `0` | no penalty | no penalty |
| `-1` | **no penalty** (`-1 > 0` false, context is `nil`) | **UNBOUNDED penalty** (`-1 != 0` passes the guard, then `if let x, x > 0` fails in `init` and lands in the `else` branch allocating `GeneratedTokenCounts()`) |

So a caller passing a negative to mean "off" — which worked, because the old guard treated every non-positive value as off — now gets the strongest possible setting, silently and in the opposite direction to the intent.

Nothing in the tree passes a negative today, so this is a latent API edge rather than a live bug, and I don't think it should block the merge. One line at the guard site preserves the old semantics:

```swift
if let presencePenalty, presencePenalty != 0,
   presenceContextSize.map({ $0 > 0 }) ?? true { … }
```

Same for `frequencyContextSize`. The guard site is where the old "non-positive means off" behaviour lived, so that's where it should be kept rather than normalising inside `init`.

Blast radius otherwise checks out: `processor()` only builds these contexts when the penalty is non-zero, so every caller not setting `presence_penalty`/`frequency_penalty` is byte-identical before and after — only callers already using the parameters see a change, and for them it moves toward the published definition. The `Int?` change also has no production call sites outside this repo's own tests.

PR #297, opened 2026-08-22
Adopt model-declared presence_penalty and frequency_penalty

rcfa:
`GenerationConfigFile` decodes `temperature`, `top_p`, `top_k`, `min_p`, `repetition_penalty`, `do_sample`, `suppress_tokens` and `max_new_tokens`, and `GenerateParameters(generationConfig:fallback:)` adopts every one of them:

```swift
if let repetitionPenalty = generationConfig.repetitionPenalty {
    self.repetitionPenalty = repetitionPenalty
}
if generationConfig.doSample == false {
    self.temperature = 0
}
```

`presence_penalty` and `frequency_penalty` are the only sampling parameters missing from it. `GenerateParameters` already has both fields and both processors — only the adoption is absent.

HuggingFace's own `GenerationConfig` has neither key, which is presumably why they were skipped; they are vLLM/OpenAI parameters. But model authors write them into this file anyway, because it is where a bundle's sampling defaults live in practice.

**What this is and isn't.** In the bundles I can inspect, the key is present but set to `0.0` — the base default — with any non-zero value living in a mode-specific block instead. So this is a **consistency** fix rather than an unlock: a caller using the opt-in defaults currently cannot distinguish "the author declared 0" from "nobody parsed it", and a bundle that *does* declare a non-zero value has it silently dropped. I'd rather present it that way than imply it changes numbers today.

**Opt-in only, so nothing existing changes.** Per `ModelContainer.defaultGenerateParameters`, `generate()` and `streamGenerate()` do not consume these defaults — the runtime keeps the explicit caller-controlled `GenerateParameters` contract. The doc comment listing the covered fields is updated to include the two new ones.

Three files, ~26 lines: two decoded fields, two adopted, one doc list.

### DIFF
--- a/Libraries/MLXLMCommon/Evaluate.swift
+++ b/Libraries/MLXLMCommon/Evaluate.swift
@@ -928,9 +928,23 @@ public struct PresencePenaltyContext: LogitProcessor {
         self.ring = TokenRing(capacity: presenceContextSize)
     }
 
-    mutating public func prompt(_ prompt: MLXArray) {
-        ring.loadPrompt(prompt)
-    }
+    /// DELIBERATELY EMPTY: the presence penalty applies to GENERATED tokens only.
+    ///
+    /// `presence_penalty` is OpenAI's parameter, and vLLM — the reference implementation most
+    /// published values are measured against — computes it from `output_tokens` alone:
+    ///
+    ///     output_bin_counts, output_mask = get_token_bin_counts_and_mask(output_tokens_tensor, …)
+    ///     logits -= presence_penalties.unsqueeze(dim=1) * output_mask
+    ///
+    /// Seeding the ring with the prompt penalises tokens the model never chose. With a long prompt the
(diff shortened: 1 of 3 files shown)


jjang-ai (approved):
Verified the claim that makes this safe: it really is opt-in.

`ModelContainer.defaultGenerateParameters(fallback:)` is an explicit call — `generate()` and `streamGenerate()` take caller-supplied `GenerateParameters` and never route through it. So adopting two more keys into `GenerateParameters(generationConfig:fallback:)` can only affect a caller that has already asked for config-derived defaults. Nothing existing changes.

Appreciate the honesty in the framing, too — saying "the key is present but `0.0` in the bundles I can inspect, so this is a consistency fix rather than an unlock" is more useful than implying it moves numbers today. The real value is that a caller currently cannot distinguish "the author declared 0" from "nobody parsed it", and a bundle that does declare a non-zero value has it silently dropped.

Approving.

Issue #299, opened 2026-08-23
A model should be able to say what reasoning levels it has (protocol over the existing per-family policies)


rcfa:
# A model should be able to say what reasoning levels it has

vmlx knows how to *set* reasoning effort for several model families. There is no way to *ask* a model
what levels it has, what its default is, or whether thinking can be turned off at all. Every consumer
therefore re-derives that from the bundle, and gets it wrong in the same way vmlx already documented
once.

## The same mechanism, twice, with different keys

```jinja
gpt-oss/chat_template.jinja
  {%- if reasoning_effort is not defined %}{%- set reasoning_effort = "medium" %}{%- endif %}
  {{- "Reasoning: " + reasoning_effort + "\n\n" }}

muse_glimmer/chat_template.jinja
  {%- set rs = reasoning_strength if reasoning_strength is defined and reasoning_strength else 'high' -%}
  {{- 'Reasoning strength: ' + rs + '.' -}}
```

Structurally identical: a template variable, a vocabulary, an in-template default. They differ only in
the key (`reasoning_effort` / `reasoning_strength`) and the default (`medium` / `high`).

`LLMModelFactory` populates the first (`context["reasoning_effort"] = effort`). Nothing populates the
second — `reasoning_strength` does not appear anywhere in the library — so every Muse-Glimmer request
runs at `high` no matter what the caller asked for, and the caller has no way to find out.

That failure mode is already on record in this repo, for the other model:

> `// … defaulting to "medium" when the key is ABSENT — which is why an undeclared gpt-oss silently ran`
> `// at medium while claiming level 1.`
> — `ReasoningContract` fallback table comment

It recurred because the knowledge has nowhere to live except a comment.

## What exists today

| | mechanism | can a caller set it? | can a caller ENUMERATE it? |
|---|---|---|---|
| DeepSeek V4 | `DeepseekV4ReasoningPolicy` | yes | no — only `normalizedReasoningEffort(_:)` |
| Hunyuan 3 | `Hy3ReasoningTemplateContext` | yes | no — only `applies(to:)` |
| gpt-oss | generic template passthrough | yes | no |
| Muse Glimmer | — | **no** | no |

Two hand-written policies, two `modelType` string checks, two private vocabularies, two aliasing
tables (`minimal|medium → low` in one, `instruct|none|no_think|off|false` in the other), and
`grep "protocol.*Reasoning"` returns nothing. Each answers "normalise this string for me". Neither
answers "what are your levels".

[...]

PR #300, opened 2026-08-23
A model can be asked what reasoning levels it has

rcfa:
Implements [#299](https://github.com/osaurus-ai/vmlx-swift/issues/299).

**The Muse Glimmer fix that motivated this has moved to #303** and can merge without this PR. What
remains here is the capability surface on its own — see "Update: split" below for what changed after
review. The section immediately following is kept because it is the evidence for WHY the surface is
worth having, not a claim about what this diff now fixes.

vmlx can be TOLD a model's reasoning effort. It cannot be ASKED what efforts a model has. So every
consumer re-derives that from the bundle and gets it wrong — including us, which is how this started.

## The bug that motivated it (now fixed separately, in #303)

Muse Glimmer's chat template reads `reasoning_strength`:

```jinja
{%- set rs = reasoning_strength if reasoning_strength is defined and reasoning_strength else 'high' -%}
{{- 'Reasoning strength: ' + rs + '.' -}}
```

`reasoning_strength` appears nowhere in the library. `llmMergedAdditionalContext` hardcodes
`context["reasoning_effort"] = effort`, so **every Muse Glimmer request runs at `high`** regardless of
what the caller asked for, and the caller has no way to notice.

That is the same failure this repo already documented for the other family:

> `// … defaulting to "medium" when the key is ABSENT — which is why an undeclared gpt-oss silently ran`
> `// at medium while claiming level 1.`

Two models, structurally identical templates, different keys — and the knowledge of which key belongs
to which family had nowhere to live except a comment.

## What this adds

**`ReasoningEffortPolicy`** — what one family knows about its own controls: `templateKey`, `efforts`
(ordered), `defaultEffort`, `supportsDisabling`, `offEffort`, `normalize(_:)`.

`supportsDisabling` and `offEffort` are separate on purpose, because the families genuinely differ:
DeepSeek V4 disables via `enable_thinking = false` (per its own deprecation note), while Hunyuan 3
expresses off as the effort *value* `no_think`.

**`ReasoningCapability`** — the querying surface, as one integer scale. `0` = not reasoning, `1…N` =
increasing effort. The whole space falls out of which levels are present, with no separate flags:

| levels | meaning | example |
|---|---|---|
| `[0]` | non-thinking model | |
| `[1]` | think-only, one effort | |
| `[0, 1]` | thinking on/off | most models |
| `[1, 2, 3, 4]` | think-only with four efforts | Muse Glimmer |
| `[0, 1, 2, 3]` | on/off plus three efforts | gpt-oss, DeepSeek V4 |

`levels.contains(0)` replaces a string comparison against `["no_think", "none", "off", "false",
"instruct", …]` that each family currently maintains separately.

**Four conformances.** `GptOssReasoningPolicy` and `MuseGlimmerReasoningPolicy` are new — the first is
declaration-only (its setting path already worked), the second is the bug fix. `DeepseekV4EffortPolicy`
and `Hy3EffortPolicy` delegate `normalize` to the existing types, so the accepted sets cannot drift
from the code that already enforces them.

**Fallbacks, because most models declare nothing.** They are still one of three definite shapes, and a
caller has to know which — asking a think-only model to stop, or a non-thinking model to think, both
return something other than what was requested. Resolution order: family policy → the model's own
declaration → `supports_thinking` → does the template read `enable_thinking` → `[0, 1]`. When none of
those answer, the question is no longer "which control" but "is there one": a template that does not
read `enable_thinking` cannot switch reasoning from the prompt whatever we claim, so the terminal step
asks only whether the model reasons at all, and reports no control rather than a toggle it could not
honour. Claiming the toggle there would be the worst of the three — it advertises a switch, a caller
sets it, the template ignores it, and the run is recorded as having reasoned.

## Scope

Additive. No behaviour change for DeepSeek V4, Hunyuan 3 or gpt-oss: their entry points are untouched
and gain a declaration. One existing line moves — `Hy3ReasoningTemplateContext.applies(to:)` now
delegates to `Hy3EffortPolicy.applies`, so "is this Hunyuan 3" has one definition rather than two that
can drift, and it sits beside `DeepseekV4ReasoningPolicy.isDeepseekV4` in the same module.

Muse Glimmer changes behaviour, because it currently ignores a control its own template reads.

`llmMergedAdditionalContext` is deliberately **not** touched in this draft: it carries
`_ = modelType // intentionally unused: no family-name coercion`, and wiring the registry in there is
exactly the decision I would rather you made than assumed. The policy exposes `templateKey`, so the
call site becomes a lookup rather than a hardcoded key — but where that lookup belongs is your call.

## Tests

`ReasoningEffortPolicyTests` — pure logic, no MLXArray, so it needs no GPU. Covers each family's shape
and default, all three fallbacks, the resolution order, and two invariants across every registered
family: levels are contiguous and contain their own default, and the predicates are mutually disjoint.

`ReasoningCapabilitySourceTests` — pins which rung produces which `Source`, and that relabelling a shape
as declared does not quietly change the shape. These replace the fill tests an earlier revision carried;
the fill itself is gone.

**The diff touches `Package.swift`, and that is part of the fix.** `Tests/MLXLMCommonFocusedTests`
enumerates its `sources:` explicitly, so a test file that is not listed there compiles nowhere — the
build stays green and the suite reports nothing missing, because from its point of view nothing is.
An earlier revision of this PR shipped `ReasoningEffortPolicyTests.swift` without the corresponding
entry, which meant 20 tests that had never executed once. Registering them surfaced a raw-string escape
(`\#(...)` inside a normal literal) and two genuine failures about what an unrecognised model should
report — the second of which is the reason the terminal fallback above returns no control rather than a
toggle. Both test files are now listed; 604 tests in that target pass.

Worth flagging beyond this PR: any test file added to that target without a `sources:` entry is silently
inert, and nothing in the build output says so. — Since writing that, you have generalised it in #302,
including a second inert file I had not found. Verified against `a3f857e`: 61 test files on disk, 61
listed, so that guard passes with this branch applied.

## Where it came from

A benchmarking harness needs to ask a model what thinking levels it has, select one, and record *which
level a published number was measured at* — otherwise two rows labelled "thinking" are not comparable.
We built that guess-work locally and it read Muse Glimmer's four levels as zero, because the model
declares them under `modes` while we looked for `reasoning_effort_levels`. The reconciliation belongs
here: vmlx is the only layer that sees both the bundle and the family-specific code.

A model host's UI wants the same thing for a simpler reason — to render the control at all, you first
have to know whether there is one, and what its stops are.

---

## Update: split, and the review notes addressed

**The Muse Glimmer fix has moved out of this PR**, to #303, at your suggestion — a user-visible bug
should not move at the speed of a design discussion. That PR is a `MuseGlimmerReasoningTemplateContext`
beside `Hy3ReasoningTemplateContext`, exactly the shape you pointed at, with no dependency on anything
here. It stands alone and can merge without this.

**The `llmDefaultAdditionalContext` fill is gone from this PR too.** It was what made the registry
change behaviour, and with the translation living in #303 it has no bug left to fix. What remained was
inert by construction: for every family with an effort vocabulary the filled value IS the consumer's own
fallback — Hunyuan 3's template assigns `no_think` when the key is missing, Muse Glimmer's falls through
to `'high'`, `DeepseekV4ChatEncoder` switches on `reasoningEffort ?? .low`, harmony runs
`if reasoning_effort is not defined` — so it restated defaults already in force. Inert surface area in a
change being judged on its design is only risk. `LLMModelFactory` is now untouched by this PR entirely,
which also retires the question of where a registry lookup belongs on that call site: it does not, yet.

So what is left here is only the thing you said was genuinely missing: being able to ASK a model what
levels it has.

### Delegation is now a rule, not a habit

You asked for this to be stated rather than just done for the two families that have normalisers. It is
now on the protocol requirement itself, with the reason: a second copy of a vocabulary drifts from the
first the moment either changes, and the copy that drifts is the one no caller is testing. Families
without an existing normaliser get the default implementation, which is derived from `efforts` and
`offEffort` and so cannot drift from them either.

### A fallback is now distinguishable from a declaration

This was the sharpest note, and it was right: if the guess is wrong the caller has no way to notice,
which is the same failure mode as the bug this exists to prevent. `ReasoningCapability` now carries where
it came from:

```swift
public enum Source: String, Sendable, Equatable, CaseIterable {
    case policy     // a registered family — strongest, it also knows the template key
    case declared   // the model's own words
    case inferred   // our deduction from artefacts that merely suggest a shape
}
```

The ladder records it per rung. `supports_thinking` is the model's own word, so a shape derived from it
reports `declared` even though it carries no vocabulary; the template scan and the marker sweep are ours,
and report `inferred`. A UI can render a control confidently for `policy` and `declared`, and treat
`inferred` as a best effort that may be silently wrong — which is exactly the distinction that was
missing.

One consequence worth flagging, since it changed while addressing this: an unrecognised model with no
declaration, no `supports_thinking` and no toggle in its template now reports `[0]` — no control — where
an earlier revision of this PR said `[0, 1]` on the grounds that a toggle "preserves today's behaviour".
That was wrong. `forModel(at:)` funnels a signal-less bundle into the same call, and under `[0, 1]` such
a bundle advertises a switch, a caller sets it, the template ignores it, and the run is recorded as
having reasoned. A toggle is a claim that reasoning can be changed from the prompt; where nothing reads
the key, that claim is false however permissively it was meant.

## Coverage: one bundle per dispatched model type

The concern with a change like this is not the families it names, it is the 80-odd it does not. So here
is what the resolver makes of ONE REPRESENTATIVE BUNDLE for every `model_type` the factories dispatch,
resolved from real metadata — `config.json`, the chat template, `tokenizer_config.json`. No weights were
downloaded; the question is answerable from kilobytes.

**Read the rows as "this bundle", not "this architecture".** Capability is a property of the BUNDLE, not
of the model type, and the two are routinely different:

| bundle | `model_type` | levels |
|---|---|---|
| `DeepSeek-R1-Distill-Llama-8B` | `llama` | **`[1]`** |
| `Llama-3.2-1B-Instruct`, `Meta-Llama-3.1-8B-Instruct` | `llama` | `[0]` |
| `Phi-4-reasoning-plus` | `phi3` | **`[1]`** |
| `AI21-Jamba-Reasoning-3B` | `jamba` | **`[1]`** |
| a qwen3-235b thinking merge | `qwen3_moe` | **`[1]`** |

A reasoning distill and its base share an architecture and differ entirely in whether they expose a
thinking channel. That is exactly why `forModel(at:)` takes a directory rather than a model type, and
why the family policies cover the CONTROL VOCABULARY (which key, which efforts) while the presence of
reasoning at all is read from the artefacts. Anyone reading the table below as "llama is non-thinking"
would be right to object — the row means the Llama Instruct bundle that was sampled, and a Llama-based
reasoner in the same table resolves to `[1]`.

**91 types resolved, 1 without a template** (that one is marked, because a shape derived from
missing evidence is not a result). Distribution:

| shape | count | reading |
|---|---|---|
| `[0]` | 57 | the sampled bundle exposes no thinking channel — base Llama/Mistral/Gemma-3/Phi-2 instructs, starcoder, pixtral … |
| `[0,1]` | 22 | a real toggle — qwen3 family, gemma4, bailing, diffusion_gemma |
| `[1]` | 6 | think-only, uncontrollable |
| `[1,2,3,4]` | 2 | muse_glimmer (+ its text tower), default `high` |
| `[0,1,2]` | 2 | |
| `[0,1,2,3]` | 2 | gpt-oss (default `medium`), DeepSeek V4 (default `low`) |

Spot-checks that can be verified without trusting the tool: `gemma`, `gemma2` and `gemma3` are
non-thinking while `gemma4` is a toggle; `llama` and `mistral` are non-thinking; gpt-oss defaults to
level 2, not level 1.

Building this table found a defect rather than confirming correctness, which is the argument for
having it. The reader looked for the chat template only in `chat_template.jinja` — and across these
bundles the template is split almost evenly between that file and a `chat_template` field inside
`tokenizer_config.json`. Half the collection was reading as "no template", which reads as "no toggle
expressible", which reads as "non-thinking": a confident answer produced by looking in one place.

<details><summary>Full table (91 types)</summary>

| model_type | levels | default | key | evidence |
|---|---|---|---|---|
| `NemotronH_Nano_Omni_Reasoning_V3` | `[0, 1]` | 1 | `—` | template |
| `afmoe` | `[0]` | 0 | `—` | template |
| `apertus` | `[1]` | 1 | `—` | template |
| `apertus1p5` | `[0, 1]` | 1 | `—` | template |
| `baichuan_m1` | `[0]` | 0 | `—` | template |
| `bailing_moe` | `[0, 1]` | 1 | `enable_thinking` | template |
| `bitnet` | `[0]` | 0 | `—` | template |
| `chatglm` | `[0]` | 0 | `—` | template |
| `cohere2_moe` | `[0]` | 0 | `—` | template |
| `cohere_asr` | `[0]` | 0 | `—` | template |
| `cohere_compass` | `[0]` | 0 | `—` | template |
| `deepseek_v3` | `[1]` | 1 | `—` | template |
| `deepseek_v4` | `[0, 1, 2, 3]` | 1 | `reasoning_effort` | template |
| `deepseekocr` | `[0]` | 0 | `—` | template |
| `diffusion_gemma` | `[0, 1]` | 1 | `—` | template |
| `ernie4_5` | `[0]` | 0 | `—` | template |
| `exaone4` | `[0, 1]` | 1 | `—` | template |
| `falcon_h1` | `[0]` | 0 | `—` | template |
| `gemma` | `[0]` | 0 | `—` | template |
| `gemma2` | `[0]` | 0 | `—` | template |
| `gemma3` | `[0]` | 0 | `—` | template |
| `gemma3_text` | `[0]` | 0 | `—` | template |
| `gemma3n` | `[0]` | 0 | `—` | template |
| `gemma4` | `[0, 1]` | 1 | `—` | template |
| `gemma4_text` | `[0]` | 0 | `—` | template |
| `gemma4_unified` | `[0, 1]` | 1 | `—` | template |
| `gemma4_unified_text` | `[0, 1]` | 1 | `—` | template |
| `glm4` | `[0]` | 0 | `—` | template |
| `glm4_moe` | `[0, 1]` | 1 | `—` | template |
| `glm4_moe_lite` | `[0, 1]` | 1 | `—` | template |
| `glm4v` | `[0, 1]` | 1 | `—` | template |
| `glm4v_moe` | `[0, 1]` | 1 | `—` | template |
| `glm_ocr` | `[0, 1]` | 1 | `—` | template |
| `gpt_oss` | `[0, 1, 2, 3]` | 2 | `reasoning_effort` | template |
| `granite` | `[1]` | 1 | `—` | template |
| `granitemoehybrid` | `[0]` | 0 | `—` | template |
| `hunyuan_v1_dense` | `[0, 1, 2]` | 0 | `reasoning_effort` | template |
| `hy_v3` | `[0, 1, 2]` | 0 | `reasoning_effort` | template |
| `idefics3` | `[0]` | 0 | `—` | template |
| `internlm2` | `[0]` | 0 | `—` | template |
| `jamba` | `[0]` | 0 | `—` | template |
| `kimi_vl` | `[0]` | 0 | `—` | template |
| `laguna` | `[0, 1]` | 1 | `—` | template |
| `lfm2` | `[0]` | 0 | `—` | template |
| `lfm2-vl` | `[0]` | 0 | `—` | template |
| `lfm2_moe` | `[0]` | 0 | `—` | template |
| `lfm2_vl` | `[0]` | 0 | `—` | template |
| `lille-130m` | `[0]` | 0 | `—` | template |
| `llama` | `[0]` | 0 | `—` | template |
| `llama4` | `[0]` | 0 | `—` | template |
| `llava_qwen2` | `[0]` | 0 | `—` | template |
| `mimo` | `[0]` | 0 | `—` | template |
| `mimo_v2` | `[0, 1]` | 1 | `—` | template |
| `mimo_v2_flash` | `[0, 1]` | 1 | `—` | template |
| `minicpm` | `[0]` | 0 | `—` | template |
| `minimax_music3` | `[0]` | 0 | `—` | config only |
| `mistral` | `[0]` | 0 | `—` | template |
| `mistral3` | `[0]` | 0 | `—` | template |
| `mixtral` | `[0]` | 0 | `—` | template |
| `muse_glimmer` | `[1, 2, 3, 4]` | 3 | `reasoning_strength` | template |
| `muse_glimmer_text` | `[1, 2, 3, 4]` | 3 | `reasoning_strength` | template |
| `nanbeige` | `[0, 1]` | 1 | `—` | template |
| `nanochat` | `[0]` | 0 | `—` | template |
| `needle` | `[0]` | 0 | `—` | template |
| `nemotron_h` | `[0, 1]` | 1 | `—` | template |
| `olmo2` | `[0]` | 0 | `—` | template |
| `olmo3` | `[0]` | 0 | `—` | template |
| `olmoe` | `[0]` | 0 | `—` | template |
| `openelm` | `[0]` | 0 | `—` | template |
| `paddleocr_vl` | `[0]` | 0 | `—` | template |
| `paligemma` | `[0]` | 0 | `—` | template |
| `phi` | `[0]` | 0 | `—` | template |
| `phi3` | `[0]` | 0 | `—` | template |
| `phimoe` | `[0]` | 0 | `—` | template |
| `pixtral` | `[0]` | 0 | `—` | template |
| `qwen2` | `[0]` | 0 | `—` | template |
| `qwen2_5_vl` | `[0]` | 0 | `—` | template |
| `qwen2_vl` | `[0]` | 0 | `—` | template |
| `qwen3` | `[0, 1]` | 1 | `—` | template |
| `qwen3_5` | `[0, 1]` | 1 | `—` | template |
| `qwen3_5_moe` | `[0, 1]` | 1 | `—` | template |
| `qwen3_moe` | `[1]` | 1 | `—` | template |
| `qwen3_next` | `[0]` | 0 | `—` | template |
| `qwen3_vl` | `[0]` | 0 | `—` | template |
| `seed_oss` | `[1]` | 1 | `—` | template |
| `smollm3` | `[0, 1]` | 1 | `—` | template |
| `smolvlm` | `[0]` | 0 | `—` | template |
| `starcoder2` | `[0]` | 0 | `—` | template |
| `step3p5` | `[1]` | 1 | `—` | template |
| `vision-encoder-decoder` | `[0]` | 0 | `—` | template |
| `wav2vec2` | `[0]` | 0 | `—` | template |

</details>

### DIFF
--- a/Libraries/MLXLLM/Hy3ReasoningTemplateContext.swift
+++ b/Libraries/MLXLLM/Hy3ReasoningTemplateContext.swift
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: MIT
 
 import Foundation
+import MLXLMCommon
 
 /// Template-context adapter for official Hunyuan v3 (`model_type = hy_v3`).
 ///
@@ -19,10 +20,10 @@ import Foundation
 /// - with neither present, nothing is injected and the template's own
 ///   default (`no_think`) applies.
 public enum Hy3ReasoningTemplateContext {
+    /// Delegates to `Hy3EffortPolicy`, which declares this family's efforts and default alongside
+    /// the other families. One definition of "is this Hunyuan 3", not two that can drift apart.
     public static func applies(to modelType: String?) -> Bool {
-        guard let t = modelType?.lowercased() else { return false }
-        return t == "hy_v3" || t.hasPrefix("hy_v3_") || t.hasPrefix("hy3")
-            || t.hasPrefix("hunyuan")
+        Hy3EffortPolicy.applies(to: modelType)
     }
 
     public static func apply(
--- a/Libraries/MLXLMCommon/ReasoningEffortPolicy.swift
+++ b/Libraries/MLXLMCommon/ReasoningEffortPolicy.swift
@@ -1,0 +1,508 @@
+import Foundation
+
+// A model family can already be TOLD its reasoning effort. It cannot be ASKED what efforts it has.
+//
+// `DeepseekV4ReasoningPolicy` and `Hy3ReasoningTemplateContext` both normalise an effort string for
+// their family, each with a private vocabulary and its own aliases, and neither can answer "what are
+// your levels", "what is your default", or "can you stop reasoning at all". So every consumer
+// re-derives that from the bundle — and gets it wrong, because the bundle is only half the evidence:
+// the other half is here, in the family-specific code that knows which template key to populate.
+//
+// This file adds the missing question. It is deliberately additive: the existing policies keep their
+// entry points and gain a declaration, and the two families that had no policy at all get one.
+
+/// What one model family knows about its own reasoning controls.
+public protocol ReasoningEffortPolicy: Sendable {
+    /// Whether this policy governs the given `model_type`.
+    static func applies(to modelType: String?) -> Bool
+
+    /// The chat-template variable this family reads — `reasoning_effort` for most, but not all:
+    /// Muse Glimmer's template reads `reasoning_strength`, which is why nothing was populating it.
+    static var templateKey: String { get }
+
+    /// Thinking efforts, ordered weakest to strongest, in the family's own vocabulary.
+    /// Excludes any value that MEANS "off"; that is ``offEffort``.
+    static var efforts: [String] { get }
+
+    /// What the template falls back to when the key is absent. Knowing this is the difference
+    /// between "we ran at level 1" and "we ran at whatever the template chose".
+    static var defaultEffort: String { get }
+
+    /// Whether reasoning can be turned off at all. Separate from ``offEffort`` because families
+    /// disagree about HOW: DeepSeek V4 uses `enable_thinking = false` (its own deprecation note says
+    /// so), while Hunyuan 3 expresses off as the effort value `no_think`.
+    static var supportsDisabling: Bool { get }
+
+    /// The effort value meaning "do not reason", when the family expresses off that way. `nil` when
+    /// off is reached by another mechanism, or not at all.
+    static var offEffort: String? { get }
+
+    /// Map a caller's string onto this family's vocabulary, or `nil` if it has no meaning here.
+    /// Map a raw effort string onto this family's vocabulary, or `nil` when it names nothing.
+    ///
+    /// RULE FOR CONFORMANCES: where the family already has a normaliser, this must DELEGATE to it
+    /// rather than restate the accepted set. `DeepseekV4EffortPolicy` and `Hy3EffortPolicy` do, and
+    /// the reason generalises — a second copy of a vocabulary drifts from the first the moment either
+    /// changes, and the copy that drifts is the one no caller is testing. A family with no existing
+    /// normaliser gets the default implementation below, which is derived from `efforts` and
+    /// `offEffort` and so cannot drift from them either.
+    static func normalize(_ raw: String) -> String?
+}
+
+extension ReasoningEffortPolicy {
+    /// Case-insensitive exact match against the declared vocabulary. Families with aliases override.
+    public static func normalize(_ raw: String) -> String? {
+        let v = raw.trimmingCharacters(in: .whitespacesAndNewlines).lowercased()
+        if let offEffort, v == offEffort { return offEffort }
+        return efforts.first { $0 == v }
+    }
+
+    public static var capability: ReasoningCapability {
+        ReasoningCapability(efforts: efforts, defaultEffort: defaultEffort,
+                            supportsDisabling: supportsDisabling, offEffort: offEffort,
+                            templateKey: templateKey)
+    }
+}
+
+// MARK: - The querying surface
+
+/// A model's reasoning controls as one integer scale: `0` = not reasoning, `1…N` = increasing effort.
+///
+/// The scale is the useful shape because it is what a caller has to render or record. The whole space
+/// falls out of which levels are present, with no separate flags:
+///
+/// | levels | meaning |
+/// |---|---|
+/// | `[0]` | a non-thinking model |
+/// | `[1]` | think-only, one effort |
+/// | `[0, 1]` | thinking on/off |
+/// | `[1, 2, 3, 4]` | think-only with four efforts |
+/// | `[0, 1, 2, 3]` | on/off plus three efforts |
+///
+/// "Can this model stop thinking?" is `levels.contains(0)`, rather than a string comparison against
+/// `["no_think", "none", "off", "false", "instruct", …]` maintained separately in each family.
+///
+/// WHAT COUNTS AS THINKING. A thinking model is one that exposes a thinking CHANNEL — a chain-of-
+/// thought trace something can separate from the answer. What it does internally is not the question
+/// and is not observable anyway. So:
+///
+/// - no level selection, no on/off, **no trace** → `[0]`, non-thinking, whatever it does internally;
+/// - no level selection, no on/off, **but a trace** → `[1]`, thinking, just not controllable;
+/// - an on/off control → `[0, 1]`; an effort vocabulary → `[0…N]` or `[1…N]`.
+///
+/// `VibeThinker-3B` is the first case: its template is plain ChatML with no delimiter and no toggle,
+/// so the name promises reasoning the interface does not expose. Reporting `[1]` would offer a caller
+/// a channel it cannot read.
+///
+/// Trace detection is a heuristic over the markers models actually ship — a `<think>` delimiter in the
+/// template, a `reasoning_content` field in the tokenizer config, a declared reasoning parser. A model
+/// with a novel delimiter reads as non-thinking until its marker is added here, which is a gap worth
+/// knowing about rather than a claim to trust blindly.
+public struct ReasoningCapability: Sendable, Equatable {
+    /// Contiguous and ascending. Contains `0` exactly when reasoning can be disabled.
+    public let levels: [Int]
+    /// The level the model uses when asked for nothing in particular.
+    public let defaultLevel: Int
+    /// Effort names for levels `1…N`, in order. Empty when the family has no effort vocabulary.
+    public let efforts: [String]
+    /// The effort value meaning off, when the family expresses off that way.
+    public let offEffort: String?
+    /// The chat-template variable to populate, when there is one.
+    public let templateKey: String?
+    /// Where this capability came from. See ``Source``.
+    public let source: Source
+
+    /// How much the rest of this value is worth trusting.
+    ///
+    /// The fallbacks are the part most likely to rot: a model that declares nothing still resolves to
+    /// one of three definite shapes, and if that guess is wrong the caller has no way to notice —
+    /// which is the same failure mode as the bug this type exists to prevent. A caller must be able to
+    /// tell a declaration from a guess, so the distinction is carried here rather than being flattened
+    /// away. Render a control confidently for ``policy`` and ``declared``; treat ``inferred`` as a
+    /// best effort that may be silently wrong.
+    public enum Source: String, Sendable, Equatable, CaseIterable {
+        /// A registered family policy — the strongest, because it also knows the template key.
+        case policy
+        /// The model's own declaration, from its bundle.
+        case declared
+        /// Our guess, from artefacts that merely suggest a shape.
+        case inferred
+    }
+
+    public init(efforts: [String], defaultEffort: String?, supportsDisabling: Bool,
+                offEffort: String? = nil, templateKey: String? = nil,
+                source: Source = .policy) {
+        self.source = source
+        self.efforts = efforts
+        self.offEffort = offEffort
+        self.templateKey = templateKey
+        let thinkingLevels = efforts.isEmpty ? [1] : Array(1...efforts.count)
+        self.levels = supportsDisabling ? [0] + thinkingLevels : thinkingLevels
+        // A family may DEFAULT to not reasoning — Hunyuan 3's template falls back to `no_think` when
+        // nothing is injected — so the default can legitimately be level 0.
+        if let defaultEffort, let offEffort, defaultEffort == offEffort, supportsDisabling {
+            self.defaultLevel = 0
+        } else if let defaultEffort, let i = efforts.firstIndex(of: defaultEffort) {
+            self.defaultLevel = i + 1
+        } else {
+            // A default the family does not list is a declaration bug, not a level: fall back to the
+            // weakest thinking level rather than inventing an index for it.
+            self.defaultLevel = thinkingLevels.first ?? 0
+        }
+    }
+
+    private init(levels: [Int], defaultLevel: Int, source: Source = .inferred) {
+        self.levels = levels
+        self.defaultLevel = defaultLevel
+        self.efforts = []
+        self.offEffort = nil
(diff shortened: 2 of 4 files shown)


jjang-ai (commented):
The bug is real — I verified both halves independently, and it is worse than "a consumer gets it wrong".

- `reasoning_strength` appears **nowhere** in vmlx production code. The only hits are `MuseGlimmer_ARCHITECTURE.md`, `MuseGlimmer_PROOF_MATRIX.md`, and `MediaSalt.swift:151` — and MediaSalt only *reads* the key out of `additionalContext` for the cache scope salt. Nothing ever sets it.
- In osaurus, `reasoning_strength` appears **only in test files** (`MuseGlimmerTokenizerProbeTests`, `MuseGlimmerVisionLiveTests`). No production path sends it.

So the tests pass the key explicitly and prove the *template* honours it, while the shipping app never sends it and every request falls back to the template's `'high'`. That is exactly the shape that survives a green suite indefinitely. Our own `MuseGlimmer_PROOF_MATRIX.md:250` already says "If osaurus sends `reasoning_effort`, the template never…" — the warning was written down and not acted on, which is a fair argument that a comment was never going to be enough.

**On the shape, since that's what you're asking.** The repo already has this exact pattern: `Hy3ReasoningTemplateContext` is a per-family adapter that translates the request surface into the template's contract (Hunyuan 3 keys thinking on `reasoning_effort` with a closed set and `raise_exception`s on anything else), applied in `LLMUserInputProcessor.prepare`. A `MuseGlimmerReasoningTemplateContext` beside it fixes the shipped bug in a handful of lines and needs no new protocol.

That is not an argument against `ReasoningEffortPolicy`. Being able to *ask* a model its levels is genuinely missing, and the integer-scale design (`0` = not reasoning, `1…N` increasing) is a good one — it collapses several per-family string sets into one comparison, and `supportsDisabling`/`offEffort` being separate is correct given DeepSeek V4 disables via `enable_thinking = false` while Hunyuan 3 expresses off as the effort value `no_think`.

It is an argument for **splitting**: ship the Muse Glimmer translation as a small standalone fix that can be proven live now, and let the capability surface land on its own merits. Right now a user-visible bug is queued behind a design discussion, and those move at different speeds.

Two smaller notes on the capability surface if it lands as designed:
- Delegating `normalize` on `DeepseekV4EffortPolicy`/`Hy3EffortPolicy` to the existing types is the right call — it keeps the accepted sets from drifting from the code that already enforces them. Worth stating as a rule for future conformances, not just doing it for these two.
- The fallbacks are the part most likely to rot. A model that declares nothing still gets one of three definite shapes, and if that guess is wrong the caller has no way to notice — same failure mode as the bug this fixes. Whatever the fallback is, it should be distinguishable from a declaration at the API surface.

rcfa:
Thanks for verifying both halves independently — and the split is the right call.

Your evidence is stronger than mine was: `reasoning_strength` appearing only in docs and tests on both sides means the shipping path never sends it, so the template's support is dead code in practice rather than merely inconsistently consumed. And `Hy3ReasoningTemplateContext` being the established pattern for exactly this settles the shape question — I should have looked for the existing per-model context type before proposing a new one.

Splitting as you describe:

1. **Muse Glimmer translation** as a small standalone fix following the `Hy3ReasoningTemplateContext` pattern, provable live now.
2. **`ReasoningEffortPolicy`** — being able to ask a model its levels — kept separate, since its value does not depend on the Muse Glimmer bug and it should not hold that fix up.

I will close this in favour of the two, rather than reshape it in place, so the discussion stays attached to the right change.

PR #303, opened 2026-08-25
Muse Glimmer: translate the reasoning request into the key its template reads

rcfa:
## The bug

Muse Glimmer's chat template keys reasoning on a `reasoning_strength` variable:

```jinja
{%- set rs = reasoning_strength if reasoning_strength is defined and reasoning_strength else 'high' -%}
{{- 'Reasoning strength: ' + rs + '.' -}}
```

`reasoning_strength` appears nowhere in vmlx production code, and no caller sends it. So every request
renders at the template's own fallback of `high`, whatever effort was asked for, and nothing reports
that the request was dropped.

The bundle is not ambiguous about the contract — `jang_config.json` carries:

```json
"reasoning_control": "reasoning_strength",
"reasoning_default": "high",
"reasoning_values": ["low", "medium", "high", "xhigh"]
```

so this is a translation gap, not an unknown. `JangChatReasoning` already parses `modes` and
`default_mode`; the control KEY is the piece nothing carries.

It survives a green suite because the tests pass `reasoning_strength` explicitly and therefore prove the
template honours it, while no production path ever sets it.

## The fix

The shape already exists here: `Hy3ReasoningTemplateContext` is a per-family adapter applied in
`LLMUserInputProcessor.prepare` that translates the request surface into the template's contract.
`MuseGlimmerReasoningTemplateContext` sits beside it and needs no new protocol.

| request | result |
|---|---|
| explicit `reasoning_strength` naming a declared value | passes through |
| `reasoning_effort` — the key callers send | mapped onto `low` / `medium` / `high` / `xhigh` |
| `enable_thinking: false` | weakest strength (see below) |
| `enable_thinking: true` | nothing injected |
| nothing | nothing injected |

Two decisions worth surfacing, because both could reasonably have gone the other way:

**`enable_thinking: false` becomes the weakest strength, not nothing.** This family has no off value and
declares no off mode, so a request to stop reasoning cannot be satisfied. Honouring it as far as the
model allows seemed better than ignoring it — leaving it at `high` would answer "reason as little as
possible" with the strongest setting available.

**An unrecognised value lands on the default rather than being forwarded.** The template does not
validate, so an unknown string renders as itself — `Reasoning strength: banana.` — rather than failing
the way Hunyuan 3's `raise_exception` does. Landing on the default is what a caller gets today.

**Nothing is injected when nothing is requested.** The template's `high` already applies; restating it
would make an inert change look like a behavioural one.

## Tests

11 tests, no GPU. They cover the translation, the precedence between the two keys, the alias mapping,
and — deliberately — the cases where the adapter must do *nothing*: other families untouched, an empty
request left alone, `enable_thinking: true` injecting nothing. The other-families case is the control;
without it an adapter that fired on everything would pass the rest.

The last test is an end-to-end proof rather than another isolated assertion. It renders the template's
own reasoning line with and without the adapter, so the "before" case is the defect itself:

```
before:  Reasoning strength: high.     ← caller asked for "low"
after:   Reasoning strength: low.
```

Mutation-checked: neutering the translation fails 12 assertions across the suite.

## Scope

Additive for every other family — the adapter returns its input untouched unless `model_type` is
`muse_glimmer` or `muse_glimmer_text`. One line changes in `prepare`, wrapping the existing Hy3 call.

Split out of #300 at your suggestion, so the user-visible fix is not queued behind the capability-surface
discussion.

### DIFF
--- a/Libraries/MLXLLM/LLMModelFactory.swift
+++ b/Libraries/MLXLLM/LLMModelFactory.swift
@@ -1262,8 +1262,11 @@ private struct LLMUserInputProcessor: UserInputProcessor {
     }
 
     func prepare(input: UserInput) throws -> LMInput {
-        let additionalContext = Hy3ReasoningTemplateContext.apply(
-            additionalContext: try mergedAdditionalContext(input.additionalContext),
+        let additionalContext = MuseGlimmerReasoningTemplateContext.apply(
+            additionalContext: Hy3ReasoningTemplateContext.apply(
+                additionalContext: try mergedAdditionalContext(input.additionalContext),
+                modelType: modelType
+            ),
             modelType: modelType
         )
         let bailingMessages = BailingThinkingTemplateContext.apply(
--- a/Libraries/MLXLLM/MuseGlimmerReasoningTemplateContext.swift
+++ b/Libraries/MLXLLM/MuseGlimmerReasoningTemplateContext.swift
@@ -1,0 +1,98 @@
+// Copyright 2026 Osaurus AI. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+import Foundation
+import MLXLMCommon
+
+/// Template-context adapter for Muse Glimmer (`model_type = muse_glimmer`, `muse_glimmer_text`).
+///
+/// The chat template keys reasoning on a `reasoning_strength` variable:
+///
+/// ```jinja
+/// {%- set rs = reasoning_strength if reasoning_strength is defined and reasoning_strength else 'high' -%}
+/// ```
+///
+/// That key appears nowhere else in this library, and no caller sends it — every request therefore
+/// renders at the template's own fallback of `high`, whatever effort was asked for, and nothing
+/// reports that the request was dropped. The bundle states the contract plainly in `jang_config.json`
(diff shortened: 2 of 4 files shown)


rcfa:
Superseded by #305, which landed the same fix — and lands it better.

Muse Glimmer is registered in **both** factories. This PR wired the adapter into `LLMModelFactory` only,
so it would have missed the model whenever it loaded as a VLM — which for a multimodal bundle is the
normal case, i.e. the case that matters. #305 wires both and puts the adapter in `MLXLMCommon` where each
can reach it. It also reached the same two judgement calls independently: `max` → `xhigh` rather than
flattened into `high`, and no "off" value, so disabling maps to the `low` floor.

Closing rather than rebasing: there is nothing left in here that #305 does not already do.

On the note you left about #300 — done. `MuseGlimmerReasoningPolicy` now takes its `efforts` from
`MuseGlimmerReasoningTemplateContext.levels` and delegates `normalize` to `.normalized(_:)`, so the
vocabulary has one definition and it is the one that builds the prompt. That makes Muse Glimmer the third
family under the delegation rule rather than an exception to it. #300 is updated.

PR #304, opened 2026-08-25
Catch degenerate repetition whose unit is longer than the character scan sees

jjang-ai:
`RepetitionCycleDetector` halts a collapsed generation — but only for a verbatim unit of 32–512 characters. Above that it deliberately gives up:

> `/// Longest unit considered. A cycle longer than this is cheaper to let the token cap handle than to scan for on every chunk.`

That trade was costed against **scan time** and never against the **user's wall clock**, and the two diverge badly on slow hardware. "Let the token cap handle it" means running to `max_tokens` — 16,384 by default in the host:

| decode | time to 16,384 tokens |
|---|---|
| 12 tok/s | 22.8 min |
| **8 tok/s** | **34.1 min** |
| 6 tok/s | 45.5 min |

Reported live: an M4 Air, a 36 KB file task, a model that *"kept writing the same line over and over"*, user gave up after **35 minutes**. The arithmetic lands exactly on the report.

### Why not just raise `maximumUnitLength`

`cycle(in:)` is O(unit × tail) on **every chunk** — at 512 that is already ~1.3M character comparisons per chunk, and `tailCapacity` scales with the bound, so 2048 would cost ~16× that on the decode path. The 512 bound is defensible; I left it alone.

### What this adds

A scan at the granularity the symptom is actually reported in: **whole lines**. It compares completed lines rather than every possible character period, so it costs a handful of string comparisons per newline instead of per chunk — and it covers exactly the case the character scan abandons, a repeating block longer than 512 characters.

The character scan still runs first and is unchanged. This only fires where that scan returned nil, so no existing verdict moves.

### Guards, because a false positive is worse than the bug

- every line in the unit must be **≥24 characters trimmed AND contain a letter** — table rules (`|---|`), separators, blank lines and `    }` can never qualify
- the unit must be **primitive**, so `A A` is never reported as period 2
- **four** consecutive repeats required — at period 8 that is 32 identical substantial lines in a row
- **lines that vary never fire**: a real script whose lines differ by an index is doing useful work and must run to completion

`maximumLineCycle` is **12, not 4**. A unit longer than 512 characters necessarily spans several lines, and an earlier value of 4 could not see its own target case — an 8-line block went undetected until a test caught it. Recorded in the constant's doc comment so nobody re-tightens it.

### Tests

18 passing. New coverage: the >512-character block, four false-positive classes (table rules, blank lines, closers, short list items), varying lines, primitivity, and `VMLX_REPETITION_STOP=0` disabling the line path too.

### Scope

Does **not** cap context, lower `max_tokens`, or cap tool results — those would punish legitimate long work. It detects a pathology and stops it, which is what the existing detector already does; this just widens the class it can see.

Also does not address *near*-verbatim degeneration (a loop whose lines differ by a counter). That is genuinely ambiguous against legitimate output and is deliberately left out — the `varyingLinesDoNotFire` test pins that boundary.

### DIFF
--- a/Libraries/MLXLMCommon/RepetitionCycleDetector.swift
+++ b/Libraries/MLXLMCommon/RepetitionCycleDetector.swift
@@ -64,6 +64,27 @@ public struct RepetitionCycleDetector: Sendable {
     /// happens to echo itself is left alone.
     public static let minimumOutputLength = 128
 
+    /// Consecutive repeats of a LINE cycle required. Four back-to-back
+    /// repeats of a block whose every line clears `minimumLineLength` is
+    /// already unambiguous — at period 8 that is 32 identical substantial
+    /// lines in a row.
+    public static let minimumLineRepeats = 4
+
+    /// Longest line CYCLE considered, in lines.
+    ///
+    /// Must comfortably exceed the number of lines it takes to pass the
+    /// character scan's `maximumUnitLength` (512), or this scan cannot see its
+    /// own target case: a unit longer than 512 characters necessarily spans
+    /// several lines, and an earlier value of 4 here missed an 8-line block
+    /// entirely. Cost is a handful of whole-line comparisons per completed
+    /// line, not per chunk, so the headroom is cheap.
+    public static let maximumLineCycle = 12
+
+    /// A line must be at least this long, trimmed, before it can anchor a
+    /// line cycle — and must contain a letter. Keeps table rules (`|---|`),
+    /// separators, blank lines and `}` / `)` closers from ever qualifying.
+    public static let minimumLineLength = 24
+
     /// Rolling tail. Only the end of the stream can carry a cycle that is
     /// still running, and bounding this keeps `feed` O(1) in stream length.
     private static let tailCapacity = maximumUnitLength * (minimumRepeats + 1)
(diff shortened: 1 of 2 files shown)


jjang-ai:
Live-proven in the app, not just unit-tested. Nanbeige 4.2-3B, greedy (temperature 0), tools off, no system prompt, asked for a verbatim line repeated 400 times. **The only variable is `VMLX_REPETITION_STOP`:**

| leg | promptTokens | genTokens | stop reason | tok/s | wall clock |
|---|---|---|---|---|---|
| `VMLX_REPETITION_STOP=0` | 789 | **16,384** | `length` | 36.4 | **~450 s** |
| `VMLX_REPETITION_STOP=1` | 789 | **1,788** | **`stop`** | 59.4 | **~30 s** |

Two things make it proof rather than coincidence:

- **9.2× fewer tokens** on an identical prompt under greedy decoding, so the model's own output is deterministic and the difference cannot be sampling noise.
- **The stop REASON changes** — `length` means it exhausted the 16,384-token ceiling with nobody knowing why it stopped; `stop` means something deliberately halted it. That field cannot flip by chance.

The control leg is also the first on-demand reproduction of the reported pathology: 16,384 tokens at 36.4 tok/s is 7.5 minutes of degenerate output *on the fastest Mac made*. At the ~8 tok/s an M4 Air manages on a 12B, the same run is ~34 minutes — which is the user report, to the minute.

(The tok/s gap is expected, not suspicious: the control generates far deeper and decode slows with context.)

PR #311, opened 2026-08-26
Declare the model families that have two entry points, and check it

rcfa:
Nine model families are registered in BOTH factories, in two files that share no
symbol and never mention each other:

    LLMModelFactory   "gemma4": create(Gemma4TextConfiguration.self, Gemma4TextModel.init)
    VLMModelFactory   "gemma4": create(Gemma4Configuration.self,     Gemma4.init)

Nothing links those lines, so a change made to one is silently not made to the
other, and the result is a bundle that looks registered and quietly has no
vision. That is a real failure we hit: a Muse Glimmer fix wired the text path
and missed the multimodal one — the normal case for that bundle.

This declares the invariant so it can be checked. `DualPathFamilies.modelTypes`
lists the nine, each annotated with its two implementations, and a conformance
suite checks the declaration against both registries IN BOTH DIRECTIONS: a
family registered twice but not declared fails, and a family declared but
registered once fails. A fourth test asserts both registries are non-empty,
because otherwise an empty one would make the comparisons pass vacuously.

`ModelTypeRegistry` gains `registeredModelTypes`; there was no way to ask a
registry what it holds, which is why the two could not be compared before.

Behaviour-preserving: no registration is added, removed or changed.

## What the second registration actually expresses

Worth stating precisely, because the intuitive reading is wrong and the doc in
this PR previously gave it.

It is **not** a second bundle shape. `ModelFactoryRegistry` seeds itself with an
`MLXVLM` trampoline ahead of an `MLXLLM` one, and `load(loader:)` falls through
on *any* error — not only `unsupportedModelType` — so a routed load reaches the
LLM entry only when the VLM factory declines the bundle. For these families it
does not decline: across 63 locally held bundles spanning all nine, **every one
carries a `vision_config` and a processor config, and not one is a text-only
release**.

So in practice the LLM-side entry is reached by callers who ask
`LLMModelFactory` directly, meaning "text only" — an *intent*, expressed as a
registration because there was no parameter for it. That is worth knowing before
anyone tries to remove one of these entries: the routed path is not what keeps
them alive.

## What is, and is not, written twice

The type side is clean — `protocol VisionLanguageModelProtocol: LanguageModel` —
so a family's model is not reimplemented. Its **checkpoint-key policy** was
another matter, and that is where the risk actually sat:

| family | how the two paths relate |
|---|---|
| `diffusion_gemma` | VLM creator returns `DiffusionGemmaModel` — the same type the LLM factory registers |
| `qwen3_5_moe` | `override` + `super.sanitize(...)` |
| `gemma3` | delegates to `languageModel.sanitize` + `visionTower.sanitize` |
| `qwen3_5` | separate bodies, but the norm-convention decision goes through a shared `NormConventionResolver` |
| `muse_glimmer` | carried the centered-norm `+1` fold in **both** paths |
| `gemma4` / `gemma4_unified` | carried the JANG expert rename and the vocab trim in **both** |
| `mistral3` / `ministral3` | genuinely differ; the tied-embedding handling differs on purpose |

The last two rows are consolidated in follow-up work, not in this PR — this one
stays behaviour-preserving. They are named here because the table is the reason
the declaration matters: three families already avoid the problem structurally,
and the ones that did not are exactly the ones that drifted.

## What it does not do

It does not unify the registries. The reason is **not** that no module can see
both — an earlier draft of this description said so, and that was wrong. This
file lives in `MLXLMCommon`, and `ModelFactoryRegistry` already reaches into
both `MLXVLM` and `MLXLLM` without importing either, through `NSClassFromString`
trampolines. The compile-time direction (`MLXVLM` depends on `MLXLLM`, never the
reverse) holds, and does stop a single table naming `Gemma4TextModel` and
`Gemma4` as symbols — but a unified registry is reachable by the route the
trampolines already use. It is simply a much larger change than declaring the
invariant, and declaring it is what catches the bug.

## How the list was built

Not by reading the two factories. A source scan for `"type": create(` found
seven families and missed `mistral3` and `ministral3`, which register through
named functions (`dispatchMistral3LLM` / `dispatchMistral3VLM`) rather than an
inline creator. The test caught both on its first run — which is the argument for
the test existing: the declaration and the registries are arrived at by different
routes, so they can disagree. Maintain the list from a test failure, never from a
grep.

## Verification

- The conformance suite passes on current `main`: the declaration and both
  registries agree in both directions, and both registries are non-empty so the
  comparisons cannot pass vacuously.
- No registration is added, removed or changed. `LLMTypeRegistry` and
  `VLMTypeRegistry` are untouched.
- `DualPathFamiliesTests.swift` is registered in `MLXLMCommonFocusedTests`'s
  explicit `sources:` list, so it actually compiles — that list is opt-in, and a
  test file missing from it silently never runs.

---

## Revised after review

| request | change |
|---|---|
| drop the `> 50` / `> 15` thresholds | now non-emptiness + a known dual-path sentinel (`gemma4` in both) |
| exact contract with a useful failure | one set equality; failure names `missingFromDeclaration` and `missingFromRegistry` separately |
| machine inventory out of the API contract | the 63-bundle evidence is gone from the source doc — it lives here instead |
| narrow the routed-loader claim | the type now states it declares registry **overlap** and does not predict which registration a bundle selects |

### The five acceptance cases

They needed the comparison to become a pure function,
`DualPathFamilies.divergence(llm:vlm:declared:)`, because the live registries
agree with the declaration *by construction* — a suite that only compares them
proves the declaration is currently correct and proves nothing about the check.
Every case that matters is one where they disagree:

| case | result |
|---|---|
| registered in both, undeclared | reported in `missingFromDeclaration`, named |
| declared, registered in only one | reported in `missingFromRegistry`, named |
| fake key in one registry only | **not** dual-path — no false positive |
| both directions at once | one run names every edit needed |
| empty registries + empty declaration | agree — which is exactly why the non-emptiness guard is separate and not redundant |

**Mutation-checked**: changing `intersection` to `union` — a family registered
once counting as dual — fails four assertions.

Acceptance case 5 (no behaviour change) holds by construction: the diff removes
no line from any registry creator, factory dispatch, or loading path. It is
purely additive — a `registeredModelTypes` accessor, the declaration, and tests.

### Local evidence

CI cannot run this (see the guard noted above), so:

```
$ swift test --filter 'DualPath'
Suite "Dual-path declaration, against synthetic registries" passed
Suite "Dual-path model families" passed
Test run with 9 tests in 2 suites passed after 0.001 seconds
```

Labelled as local, not CI.

### DIFF
--- a/Libraries/MLXLMCommon/Registries/DualPathFamilies.swift
+++ b/Libraries/MLXLMCommon/Registries/DualPathFamilies.swift
@@ -1,0 +1,106 @@
+// Copyright © 2026 Osaurus AI. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+import Foundation
+
+/// Model families registered in BOTH `LLMTypeRegistry` and `VLMTypeRegistry`.
+///
+/// ## Why this exists
+///
+/// Two files register the same `model_type`, share no symbol, and never refer to each other:
+///
+///     LLMModelFactory   "gemma4": create(Gemma4TextConfiguration.self, Gemma4TextModel.init)
+///     VLMModelFactory   "gemma4": create(Gemma4Configuration.self,     Gemma4.init)
+///
+/// Nothing connects those two lines, so a change made to one is silently not made to the other and
+/// the failure surfaces as "this model has no vision" at run time, on a bundle that looks
+/// registered. That is exactly how a Muse Glimmer fix came to wire the text path and miss the
+/// multimodal one — the normal case for that bundle. This declares the invariant the two registries
+/// encode implicitly, so it can be CHECKED in both directions.
+///
+/// ## What the second registration actually expresses
+///
+/// Not a second BUNDLE SHAPE, which is the easy assumption. `ModelFactoryRegistry` seeds itself
+/// with an `MLXVLM` trampoline ahead of an `MLXLLM` one, and `load(loader:)` falls through on any
+/// error, so a routed load reaches the LLM entry only when the VLM factory declines the bundle.
+///
+/// This type declares registry OVERLAP and nothing more. It does not model, and must not be read
+/// as predicting, which registration a given bundle selects at run time: that is a property of the
+/// routed loader and of the bundle, and it is not checked here.
+///
+/// The LLM-side entry is reached by callers who ask `LLMModelFactory` directly, to mean "text
+/// only" — an INTENT, expressed as a registration because there was no parameter for it. Where a
+/// family's model supports capability-driven construction, that intent has a better home. A
+/// natively text-only RELEASE is separately trained, often at a different parameter count, and
+/// keeps needing this path.
+///
+/// ## What is, and is not, written twice
+///
+/// The type side is clean: `protocol VisionLanguageModelProtocol: LanguageModel`, so a family's
+/// model is not reimplemented. Its CHECKPOINT-KEY POLICY was another matter — `muse_glimmer`
+/// carried the centered-norm `+1` fold in both paths, and `gemma4` carried the JANG expert rename
+/// and the vocab trim in both. Those are consolidated now (`foldCenteredNorms`,
+/// `remappingSwitchMLP`, `trimmingVocabDimension`), each with one owner and a guard. `gemma3`,
+/// `qwen3_5_moe` and `diffusion_gemma` never had the problem — they delegate, subclass, and return
+/// the same type respectively. `mistral3` genuinely differs between paths and is left alone.
+///
+/// ## What it does not do
+///
+/// It does not unify the registries. Note the reason is NOT that no module can see both: this file
+/// lives in `MLXLMCommon`, and `ModelFactoryRegistry` already reaches into both `MLXVLM` and
+/// `MLXLLM` without importing either, through `NSClassFromString` trampolines. The compile-time
+/// direction (`MLXVLM` depends on `MLXLLM`, never the reverse) holds and does prevent a single
+/// table naming `Gemma4TextModel` and `Gemma4` as symbols — but a unified registry is reachable by
+/// the route the trampolines already use. It is simply a much larger change than declaring the
+/// invariant, and declaring it is what catches the bug.
+public enum DualPathFamilies {
+
+    /// `model_type` values registered in BOTH `LLMTypeRegistry` and `VLMTypeRegistry`.
+    ///
+    /// Adding a multimodal implementation for an existing text-only family means adding its
+    /// `model_type` here as well; the conformance test fails until the registration and this list
+    /// agree, in both directions.
+    ///
+    /// This list was NOT compiled by reading the two factories. It was compiled by asking the built
+    /// registries at run time, because reading them got it wrong: a source scan for `"type": create(`
+    /// found seven and missed `mistral3` and `ministral3`, which are registered through named
+    /// functions (`dispatchMistral3LLM` / `dispatchMistral3VLM`) instead of an inline creator. The
+    /// test caught that on its first run. Maintain this list from a test failure, never from a grep.
+    public static let modelTypes: Set<String> = [
+        "diffusion_gemma",   // DiffusionGemmaModel      / makeDiffusionGemmaVLM
+        "gemma3",            // Gemma3TextModel          / Gemma3
+        "gemma4",            // Gemma4TextModel          / Gemma4
+        "gemma4_unified",    // Gemma4TextModel          / Gemma4
+        "ministral3",        // dispatchMistral3LLM      / dispatchMistral3VLM
+        "mistral3",          // dispatchMistral3LLM      / dispatchMistral3VLM
+        "muse_glimmer",      // MuseGlimmerTextModel     / MuseGlimmer
+        "qwen3_5",           // Qwen35Model              / Qwen35
+        "qwen3_5_moe",       // Qwen35MoE                / Qwen35MoE
+    ]
+    /// How a pair of registries diverges from the declaration.
+    ///
+    /// A pure function of its three inputs so the contract can be exercised against synthetic
+    /// registries. Comparing only the live singletons would leave the comparison itself untested:
+    /// every case that MATTERS is one where the registries and the declaration disagree, and the
+    /// live ones agree by construction.
+    public struct Divergence: Equatable, Sendable {
+        /// Registered in both factories, absent from the declaration — a family that quietly became
+        /// dual-path. This is the direction that fails silently in production.
+        public let missingFromDeclaration: Set<String>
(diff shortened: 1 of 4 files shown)


jjang-ai:
Deep review: the invariant is useful and the implementation is narrow, but I think this needs a small revision plus executed CI before it is ready to leave draft.

## Required changes

1. **Make the non-vacuity checks durable.**

The current `> 50` / `> 15` registry-size thresholds are unrelated to the invariant and will become maintenance noise when registrations move or are conditionally compiled. The important properties are:

```swift
#expect(!llmTypes.isEmpty)
#expect(!vlmTypes.isEmpty)
#expect(llmTypes.contains("gemma4"))       // known sentinel
#expect(vlmTypes.contains("gemma4"))       // known dual-path sentinel
```

Then keep the real contract exact:

```swift
let actualDualPath = llmTypes.intersection(vlmTypes)
#expect(actualDualPath == DualPathFamilies.modelTypes)
```

One exact equality already checks both directions; if the two-direction diagnostics are clearer, keep them, but report `missingFromDeclaration` and `missingFromRegistry` explicitly so a failure tells the next author what to edit.

2. **Keep machine inventory out of the permanent API contract.**

The “63 locally held bundles” investigation is useful evidence in the PR description, but it is not a durable library invariant. Local inventory changes and does not prove what third-party bundles can contain. Source documentation should describe routed-factory semantics without presenting one machine’s inventory as universal evidence.

3. **Pin the routed-loader assumption with a focused test or narrow the prose.**

The PR correctly notes that routed loading tries VLM before LLM and currently falls through on any error. The new registry test does not verify that routing behavior. Either add a focused routing test, or state clearly that this PR only declares registry overlap and does not guarantee which registration a concrete bundle will select.

4. **Run real checks.**

Every GitHub check currently displays `SKIPPED`. Please provide an executed focused-suite command and result on the final head, plus the normal macOS build/test workflow. `MERGEABLE/CLEAN` only means Git can merge it.

## Suggested focused acceptance

- Adding a fake key to only one registry fails with a useful missing/extra diagnostic.
- Adding the same key to both registries without declaring it fails.
- Declaring a key present in only one registry fails.
- Both registries being empty cannot pass vacuously.
- No existing registry creator, factory dispatch or model-loading behavior changes.

Once those are addressed and the checks actually execute, I am comfortable with #311 landing independently. #312/#313 are conceptually motivated by this invariant, but they are not Git-stacked on #311 and do not need to be merged atomically with it.

rcfa:
Thanks — taking these in the order you suggested, starting here. One thing first,
because it changes the premise of a request that appears in all five reviews.

## The SKIPPED checks are a repo-side guard, not something a PR can fix

`.github/workflows/pull_request.yml` still carries the guard it inherited from
Apple's mlx-swift:

```yaml
lint:
  if: github.repository == 'ml-explore/mlx-swift'

mac_build_and_test:
  if: github.repository == 'ml-explore/mlx-swift'
```

In `osaurus-ai/vmlx-swift` that condition is always false, so both jobs skip
unconditionally. It is not a fork restriction and not specific to these
branches: your own merged #329 shows the same four jobs `skipping`. No
contributor can make those checks execute from a PR, so "please run real checks"
is currently unsatisfiable from my side.

The fix is one line per job — point the guard at this repository, or drop it. I
am happy to open that as a small separate PR if you want it, or leave CI
configuration to you; it seemed better to tell you than to keep posting local
results against a request I cannot meet.

Meanwhile I will keep doing what I have been doing — running the focused suite
locally on the final head and pasting the command and result — and I will label
it clearly as local evidence rather than CI.

## On the four requested changes here

All fair, and 1 and 2 are straightforwardly right:

1. the `> 50` / `> 15` thresholds are unrelated to the invariant — replacing them
   with non-emptiness plus a known dual-path sentinel, and keeping the exact
   set equality as the contract, with `missingFromDeclaration` /
   `missingFromRegistry` reported explicitly so a failure says what to edit;
2. the local-inventory evidence belongs in the PR description, not in source
   documentation presented as a universal property — removing it;
3. narrowing the prose: this PR declares registry overlap and does not guarantee
   which registration a concrete bundle selects. A routing test is a better fit
   with the routed-factory work in #315 than bolted on here, unless you would
   rather see it in this PR;
4. covered above.

Working through these now; I will post the revised head with the focused-suite
output and your five acceptance cases.

rcfa:
Revised head pushed. All four points addressed.

**1. Durable non-vacuity.** The `> 50` / `> 15` thresholds are gone — you were
right that they are unrelated to the invariant and would fail on changes that are
not violations. Now non-emptiness plus `gemma4` as a known dual-path sentinel in
each registry. The contract stays one exact set equality, but the failure reports
`missingFromDeclaration` and `missingFromRegistry` separately, since "not equal"
does not say which file to edit.

**2. Machine inventory out of the contract.** The 63-bundle paragraph is removed
from the source documentation and lives in the PR description instead. Agreed
that one machine's inventory is not a library invariant and says nothing about
third-party bundles.

**3. Routing claim narrowed.** The type now states it declares registry
**overlap** and explicitly does not predict which registration a concrete bundle
selects at run time. I have not added a routing test here: it needs the routed
factory to be observable, which is #315's work, and asserting it from this PR
would mean testing a path this PR does not touch. Say the word if you would
rather have it here anyway.

**4. Your five acceptance cases** — these turned out to require a change I would
not have made otherwise, and it improved the PR:

The live registries agree with the declaration *by construction*, so a suite that
only compares them proves the declaration is currently correct and proves nothing
about the check. Every case you listed is one where they disagree. So the
comparison is now a pure function:

```swift
public static func divergence(
    llm: Set<String>, vlm: Set<String>, declared: Set<String> = modelTypes
) -> Divergence
```

exercised against synthetic registries:

| case | result |
|---|---|
| registered in both, undeclared | reported in `missingFromDeclaration`, named |
| declared, registered in only one | reported in `missingFromRegistry`, named |
| fake key in one registry only | **not** dual-path — no false positive |
| both directions at once | one run names every edit |
| empty + empty declaration | agree — which is why the non-emptiness guard is separate, not redundant |

Mutation-checked: `intersection` → `union` fails four assertions.

Acceptance case 5 holds by construction — the diff removes no line from any
registry creator, factory dispatch or loading path. Purely additive.

**Evidence** (local, since the guard above means CI cannot run it):

```
$ swift test --filter 'DualPath'
Suite "Dual-path declaration, against synthetic registries" passed
Suite "Dual-path model families" passed
Test run with 9 tests in 2 suites passed after 0.001 seconds
```

I have left this as a draft rather than marking it ready, because your review
asked for executed CI and I can only offer a local run. Marking it ready is one
click whenever you are satisfied — or tell me and I will.

#312 next, with the caller-level parity proof you asked for.

PR #312, opened 2026-08-27
Muse Glimmer: give the centered-norm fold a single owner

rcfa:
`MuseGlimmerTextModel.sanitize` and `MuseGlimmer.sanitize` each wrote the
centered-norm `+1` fold out in full. The second copy's own comment said so:

```swift
// The VLM does not route through `MuseGlimmerTextModel.sanitize`, so the
// centered-norm fold has to happen here too. Without it the text tower
// loads unfolded gains into a forward that no longer adds the `+1`,
// which leaves every one of its norms at zero gain — the model still
// loads and still emits text, and only a behavioural probe catches it.
```

That last clause is the reason this is worth fixing rather than leaving alone.
If the two copies drift, the tower loads unfolded gains into a forward that no
longer adds the `+1`, every norm sits at zero gain, and **nothing about the load
fails** — the model still loads and still emits text. A duplicated transform
whose drift is invisible to loading is the kind that should not be duplicated.

The predicate was already shared — `isCenteredNormWeight`, which deliberately
matches both bare `norm.weight` and the VLM's
`language_model.model.norm.weight`. Only the loop was copied.

`MuseGlimmerTextModel.foldCenteredNorms` now owns it, and both callers call it.

## What deliberately stays per-path

The key **mapping**. Under the VLM the text tower is a submodule and keeps its
`language_model.` prefix; the text model strips it. Those genuinely differ, and
collapsing them would be wrong. The fold does not differ, and now is not
written twice.

## Verification

- **Behaviour-preserving**: same predicate, same `+1`, same weights.
- **The guard is mutation-checked, not assumed.** Deleting the VLM's call to
  `foldCenteredNorms` fails 2 of the new suite's 3 tests. A guard that has never
  been observed to fail is not yet known to be a guard.
- The **existing real-bundle test would not have caught this** — it compares two
  *VLM* loads to each other, so a fold that drifted between the VLM and LLM
  paths passes it. That is why the new suite exercises the VLM `sanitize` path
  directly and compares it against the shared helper.
- Test weights are built `as [Float]` deliberately: a bare Swift float literal
  array makes a float64 `MLXArray`, which traps on the GPU. Caught by the test
  crashing, not by review — noted in the file so it is not "simplified" back.

---

## Caller-level proof (requested on review)

The previous tests proved the helper and spot-checked each path. These run one
fixture through the **real callers** — `MuseGlimmerTextModel.foldCenteredNorms`
and `MuseGlimmer.sanitize` — and assert the exact key set, dtype, shape and
every value, including the ones that must not move.

Three things make it a contract test rather than a self-agreement test:

- **expected folding is stated in the test**, not read from the implementation;
- **the fixture carries near-misses on purpose** — `self_attn.q_norm.weight`
  ends in `norm.weight` and must *not* fold, which a slightly wrong suffix
  predicate gets backwards;
- **the two callers are compared to each other**, not each to its own
  expectation.

### The double-fold case

A single shared owner makes one failure possible that two copies did not:
calling it twice. The fold is deliberately **not idempotent**, so that is
observable —

```
one fold : 2.0 -> 3.0
two folds: 2.0 -> 4.0
```

— and both real callers land on the single-fold value. If the fold were
idempotent this could not be tested at all.

**Mutation-checked**: removing the `foldCenteredNorms` call from the VLM caller
fails **13** assertions.

### Local evidence

```
$ swift test --filter 'MuseGlimmerNormFoldOwnership'
Test run with 6 tests in 1 suite passed after 0.075 seconds
```

Labelled local, not CI — see #332 for why CI cannot run this.

### Branch relationship

This targets `main` independently. It is conceptually downstream of #311 but
**does not contain #311 in Git ancestry** — the two can merge in either order.

### DIFF
--- a/Libraries/MLXLLM/Models/MuseGlimmerText.swift
+++ b/Libraries/MLXLLM/Models/MuseGlimmerText.swift
@@ -445,10 +445,7 @@ public class MuseGlimmerTextModel: Module, LLMModel, KVCacheDimensionProvider {
         // Fold the centered-norm `+1` in once. Every norm this model uses is
         // zero-centered, and the vision tower's LayerNorms are excluded above,
         // so the transform applies to exactly the text tower's RMSNorm gains.
-        for (key, value) in out where Self.isCenteredNormWeight(key) {
-            out[key] = value + 1
-        }
-        return out
+        return Self.foldCenteredNorms(out)
     }
 
     /// Zero-centered RMSNorm gains in the text tower. Matched by suffix because
@@ -459,6 +456,22 @@ public class MuseGlimmerTextModel: Module, LLMModel, KVCacheDimensionProvider {
     /// them leaves those layers without their `+1` entirely, which shifts unit
     /// gain to zero gain — the model still runs and still emits fluent text, so
     /// `MuseGlimmerCenteredNormCoverage` pins the list against the checkpoint.
+    /// Apply the centered-norm `+1` fold to every gain this architecture centers.
+    ///
+    /// SINGLE OWNER, deliberately. This model and the `MuseGlimmer` VLM wrapper must apply the
+    /// fold identically, and while the loop lived in both files the drift failure was SILENT:
+    /// unfolded gains load into a forward that no longer adds the `+1`, leaving every norm at
+    /// zero gain. The model still loads and still emits text; only a behavioural probe catches
+    /// it. `isCenteredNormWeight` already matches both key shapes — bare `norm.weight` here and
+    /// `language_model.model.norm.weight` under the VLM — so one function serves both callers.
+    public static func foldCenteredNorms(_ weights: [String: MLXArray]) -> [String: MLXArray] {
+        var out = weights
+        for (key, value) in out where isCenteredNormWeight(key) {
+            out[key] = value + 1
+        }
+        return out
+    }
+
     public static func isCenteredNormWeight(_ key: String) -> Bool {
         guard key.hasSuffix(".weight") else { return false }
         return key.hasSuffix("input_layernorm.weight")
--- a/Libraries/MLXVLM/Models/MuseGlimmer.swift
+++ b/Libraries/MLXVLM/Models/MuseGlimmer.swift
@@ -231,14 +231,13 @@ public class MuseGlimmer: Module, VLMModel, KVCacheDimensionProvider {
             }
             out[rehomed] = value
         }
-        // The VLM does not route through `MuseGlimmerTextModel.sanitize`, so the
-        // centered-norm fold has to happen here too. Without it the text tower
-        // loads unfolded gains into a forward that no longer adds the `+1`,
-        // which leaves every one of its norms at zero gain — the model still
-        // loads and still emits text, and only a behavioural probe catches it.
-        for (key, value) in out where MuseGlimmerTextModel.isCenteredNormWeight(key) {
-            out[key] = value + 1
-        }
-        return out
+        // The VLM does not route through `MuseGlimmerTextModel.sanitize` — the key mapping
+        // genuinely differs, since here the text tower is a SUBMODULE and keeps its
+        // `language_model.` prefix. But the fold itself must be identical on both paths, so it
+        // is called, not copied: `foldCenteredNorms` is its single owner. Copying it is what
+        // this file used to do, and the drift failure is silent — unfolded gains against a
+        // forward that no longer adds the `+1` leave every norm at zero gain, and the model
+        // still loads and still emits text.
+        return MuseGlimmerTextModel.foldCenteredNorms(out)
     }
 }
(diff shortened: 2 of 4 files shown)


jjang-ai:
Review disposition: the single-owner refactor is directionally good and appropriately keeps prefix/key mapping path-specific. Before merge, please add final-head proof that this is behavior-preserving across both callers rather than only proving the helper in isolation.

## Requested proof

1. Run the focused suite on the final SHA; all GitHub jobs currently show `SKIPPED`.
2. Feed the same deterministic weights through:
   - the text-model sanitize path with bare keys;
   - the VLM sanitize path with `language_model.`-prefixed keys.
3. Assert the exact transformed key set, dtype, shape and values, including:
   - centered norm receives `+1` exactly once;
   - a non-centered norm is unchanged;
   - unrelated weights are unchanged;
   - VLM prefix mapping remains intact;
   - calling the shared helper cannot double-fold an already processed result inside either normal caller.
4. If a Muse Glimmer bundle is available, run one text-only and one VLM load with the same prompt and record that both produce visible coherent output with no protocol leakage. This is integration proof, not a unit-test replacement.

Please also clarify the branch relationship in the description: #312 targets `main` independently; it is conceptually downstream of #311 but does not contain #311 in Git ancestry.

No architectural blocker found in the refactor itself. The missing item is executed caller-level proof on the final head.

rcfa:
Caller-level proof added; head pushed.

The same fixture now goes through both **real callers** — not the helper — with
exact key set, dtype, shape and every value asserted, including the ones that
must not move. Three details make it a contract test rather than a
self-agreement test:

- expected folding is stated in the test, not read from the implementation;
- the fixture carries deliberate near-misses: `self_attn.q_norm.weight` ends in
  `norm.weight` and must **not** fold, which a slightly wrong suffix predicate
  gets backwards;
- the two callers are compared to **each other**, not each to its own expectation.

On your point 3, "cannot double-fold": worth saying explicitly, because it is
only testable by luck. The fold is **not idempotent** —

```
one fold : 2.0 -> 3.0
two folds: 2.0 -> 4.0
```

— so a double call is observable, and both real callers land on the single-fold
value. Had the operation been idempotent there would be no way to test the
property you asked for, and I would have had to say so rather than claim it.

**Mutation-checked**: removing the `foldCenteredNorms` call from the VLM caller
fails 13 assertions.

```
$ swift test --filter 'MuseGlimmerNormFoldOwnership'
Test run with 6 tests in 1 suite passed after 0.075 seconds
```

Local, not CI — #332 has the reason and the options.

Branch relationship clarified in the description: this targets `main`
independently and does **not** contain #311 in Git ancestry; either can merge
first.

On your item 4 — a real Muse Glimmer bundle load, one text-only and one VLM,
same prompt: I have the bundle and can run it. Before I do, is generated output
from my machine evidence you actually want in the PR, or does it fall under the
same "not CI" caveat and belong in the runtime matrix you described on #315?
Happy either way; I would rather ask than paste transcripts you consider
diagnostic-only.

#313 next — I will take the ownership decision first, since that determines
whether the Gemma4 policy commit lives there or in #315.

rcfa:
Item 4 (integration proof) — done, on `OsaurusAI/Muse-Glimmer-30B-JANG_6M` (`muse_glimmer`, vision), against main at `c156413dd5`.

Same prompt both times: *"Describe what you are given, briefly, in two sentences."*

**No image attached:**
> You have provided a single prompt asking me to describe what you are given in two brief sentences. There is no additional context, data, or attachments included with the request.

**Same prompt + an image** (an MMMU accounting table):
> You are given two call options A and B with identical contract terms: time to maturity T = 0.5 yr, strike X = $50 and current underlying S = $55.
> They are quoted at different market prices, $10 for option A and $12 for option B, so the two options have the same payoff profile but different prices.

Every number there is in the PNG, not the prompt, so the vision path is genuinely feeding the tower. Both outputs are coherent and neither leaks protocol tokens or channel markers.

At benchmark level, `mmmu` on the same bundle answers correctly on the item above (engine reported `vmlx-swift@c156413dd5`).

Two notes for the record:

- Worth knowing that this bundle spends a fair number of tokens before it emits anything visible. At `--max-tokens 512` the image case returned an empty string; at 4096 it produced the text above. That is budget, not leakage — but it makes a low cap look like a broken model, which cost me a couple of runs.
- On "one text-only and one VLM load": I only have the VLM bundle locally, so the first case is that same bundle with nothing attached, not a standalone `MuseGlimmerTextModel` load through `LLMModelFactory`. In practice it still exercises the fold on both paths, because the VLM holds the text tower as its `language_model` submodule and its `sanitize` ends with `return MuseGlimmerTextModel.foldCenteredNorms(out)` — the single owner this PR introduces. A VLM load therefore runs the text model's fold rather than a copy of it, which is the property the PR is about. If you want the strict text-only-bundle case as well, point me at one and I will run it.

PR #313, opened 2026-08-27
Gemma4: share the checkpoint-key policy between the text model and the VLM

rcfa:
`Gemma4TextModel.sanitize` and `Gemma4.sanitize` read the same checkpoints, and
each wrote out the same two transforms:

1. the JANG expert rename, `.switch_mlp.` → `.experts.switch_glu.`
2. the vocab-dimension trim, over the same six tensors

Both now live on `Gemma4TextModel` — `remappingSwitchMLP` and
`trimmingVocabDimension` — and both callers call them. `MLXVLM` already depends
on `MLXLLM`, so this needs no new dependency, only an import.

## What is a parameter, and what is not

Only the key **prefix**: bare for the text model, `language_model.` under the
VLM, where the text tower is a submodule. *Which* tensors carry a vocab
dimension, and the rename itself, are not path-dependent — and are no longer
written twice.

## Why it matters

The failure mode is quiet. A rename that drifts between the two lands MoE
experts under keys no module claims, **on one path only** — so it survives
whichever path the author happened to test.

## Verification

- **Behaviour-preserving**: same renames, same tensors, same trim.
- **Mutation-checked guard.** Re-inlining the rename in `Gemma4.swift` fails
  with `Gemma4.swift inlines the expert rename again`.
- The guard is deliberately a **source-level** assertion as well as a
  behavioural one. A behavioural test of the helper alone would not catch a
  caller quietly restating the rename, and constructing a Gemma4 VLM in a unit
  test needs a full bundle. The assertion allows the literal rewrite exactly
  once — inside the helper — and forbids it in the wrapper.

### DIFF
--- a/Libraries/MLXLLM/Models/Gemma4Text.swift
+++ b/Libraries/MLXLLM/Models/Gemma4Text.swift
@@ -890,30 +890,49 @@ public class Gemma4TextModel: Module, LLMModel {
                 continue
             }
 
-            // Remap JANG expert naming to match module tree:
-            // JANG uses:          switch_mlp.{gate,up,down}_proj.*
-            // mlx-community uses: experts.switch_glu.{gate,up,down}_proj.*
-            // Module tree expects: experts.switch_glu.{gate,up,down}_proj.*
-            if newKey.contains(".switch_mlp.") {
-                newKey = newKey.replacingOccurrences(of: ".switch_mlp.", with: ".experts.switch_glu.")
-            }
+            newKey = Self.remappingSwitchMLP(newKey)
 
             processedWeights[newKey] = value
         }
 
-        // Trim vocab-dimension tensors to match config
-        let expectedVocab = config.vocabSize
-        for key in [
+        Self.trimmingVocabDimension(&processedWeights, prefix: "", vocabSize: config.vocabSize)
+
+        return processedWeights
+    }
+
+    /// Remap JANG expert naming to the module tree's.
+    ///
+    ///     JANG:           switch_mlp.{gate,up,down}_proj.*
+    ///     mlx-community:  experts.switch_glu.{gate,up,down}_proj.*
+    ///     module tree:    experts.switch_glu.{gate,up,down}_proj.*
+    ///
+    /// SINGLE OWNER: the `Gemma4` VLM wrapper reads the same checkpoints and needs the same
+    /// rename. While both files spelled it out, a change to one left the other's experts under
+    /// keys no module claims — on that path only.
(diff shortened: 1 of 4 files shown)


jjang-ai:
Review disposition: centralizing the Gemma4 checkpoint-key policy is useful, but please resolve the overlap with #315 and strengthen caller-level proof before merge.

## Required before merge

1. **Choose one owner for this patch.** #315 contains its own rebased/equivalent Gemma4 shared-policy commit. Do not merge #313 and then merge #315 unchanged. Either:
   - land #313 first and rebase/drop the duplicate commit from #315; or
   - keep this closed/draft and let the corrected #315 carry it.

2. **Test both real callers, not only the helper/source text.** For identical synthetic dictionaries, assert exact output keys/shapes/values through:
   - `Gemma4TextModel.sanitize` with bare keys;
   - `Gemma4.sanitize` with the `language_model.` prefix.

Cover the expert rename and all six vocab-bearing tensors. A source scan preventing a repeated string is a useful anti-drift guard, but it is not a behavioral substitute.

3. **Pin undersized-vocab behavior.** The inherited policy trims whenever dimension 0 differs from `vocabSize`. Please add cases for:
   - equal size: unchanged;
   - larger size: trims exactly to `vocabSize`;
   - smaller size: explicit, intentional failure/diagnostic rather than an accidental out-of-range slice.

If changing undersized behavior is deliberately outside this refactor, document it and add a regression test exposing the current contract so it is not ambiguous.

4. **Execute checks on the final head.** GitHub currently reports all jobs as `SKIPPED`. For the active Gemma4 correctness lane, include a real supported QAT/JANG_4M or MXFP4 routed load with bundle generation config, visible coherent output, no marker leakage, and truthful cache telemetry. Do not substitute a BF16/source-bundle result.

With the duplicate ownership resolved and the caller-level proof present, this can be evaluated independently of the broad capability-construction redesign.

rcfa:
Agreed on both points.

**1. Owner.** Taking your first option: land #313 first, and #315 drops the duplicate. #315 is still draft with two blockers open (routing selection, hermetic tests), so making it carry the shared-policy commit would hold a useful fix behind unrelated work.

**2. Both real callers.** Fair — the test asserted the helper and the source text, which cannot distinguish "the policy is centralised" from "the policy is centralised and both callers use it". That is the same gap you flagged on #315: a test that does not go through the production path passes whether or not the production path is wired.

Adding assertions over identical synthetic dictionaries for `Gemma4TextModel.sanitize` with bare keys and `Gemma4.sanitize` with the `language_model.` prefix, comparing exact output keys, before I mark this ready.

PR #315, opened 2026-08-27
Capability-driven construction: build the towers the caller asked for

rcfa:
A VLM bundle can only be loaded one way today: with every tower its config
declares. A caller who wants text out of a multimodal bundle either allocates a
vision tower it will never use, or reaches for a second registration in
`LLMModelFactory` that exists only to say "text only".

This makes that intent a parameter.

```swift
let full = MuseGlimmer(config)                        // everything the config offers
let text = try MuseGlimmer(config, requesting: [.text])  // no vision tower allocated
text.modalities        // [.text] — what this instance CARRIES
```

A request that exceeds what the config can build throws at construction, not
later inside `prepare()` where the cause is no longer visible.

## It reuses the vocabulary that already exists

The modality names are `MLXLMCommon`'s own `ModelRuntimeRequestModality` — the
type `ModelRuntimeCapabilitySnapshot.validate(request:)` already speaks — so
nothing translates between the two. `ModelConstructionModalities.swift` adds only
what was genuinely absent: construction-time resolution and its two errors.

The snapshot and this answer **different questions**, and are deliberately not
merged:

| | question | shape |
|---|---|---|
| `ModelRuntimeCapabilitySnapshot` | what was this bundle **trained** for? | tri-state, incl. `.unknown` |
| `constructibleModalities(of:)` | which towers can this **config** instantiate? | structural, no third state |

A `supports_vision` stamp over a config with no vision section is a broken
conversion. Keeping the two apart leaves that visible instead of averaging it
away.

## Two properties that are load-bearing

- **`.text` is never implied.** A model whose only lanes are audio-in and
  audio-out is unusual but not incoherent; a default that quietly added text
  would misdescribe it.
- **An empty selection is refused on both sides.** With every component optional,
  "nothing" is reachable by a config that merely failed to parse, and a model
  that can do nothing is a bug far more often than an intent.

## Applied to EVERY dual-path family

A convention a few families follow is a proposal, not an architecture — the next
person adding a multimodal model should find a pattern rather than a special
case. So all nine are converted.

They started in different places, and two were already most of the way there:

| family | starting point |
|---|---|
| `muse_glimmer` | full conversion |
| `mistral3` / `ministral3` | full conversion, across all three classes sharing the config — `Mistral3VLM`, `Mistral3VLMJANGTQ`, `Mistral4VLM` |
| `gemma3` | full conversion |
| `qwen3_5` / `qwen3_5_moe` | full conversion; the MoE variant inherits it by subclassing |
| `gemma4` / `gemma4_unified` | towers and `audioConfig` were **already** optional; needed the config section and the API |
| `diffusion_gemma` | already returned the **same type** from both factories and already attached the tower conditionally; needed only the API |

## The declared lanes differ, and that is the substance

| family | declares | why |
|---|---|---|
| `qwen3_5` | `.text .vision .video` | `prepare` consumes `input.video`; the tower takes a `videoGridTHW` |
| `mistral3` | `.text .vision` | Pixtral is image-only |
| `gemma3` | `.text .vision` | SigLIP is image-only |
| `gemma4` | `.text .vision` (+`.audio`) | `.audio` only for the **conformer** tower; never `.video` — `prepare` refuses it outright |

One "vision" flag could not express that. Declaring `.video` on Gemma 4 would let
a video request pass construction and fail at `prepare()` — precisely the failure
this design moves earlier.

`diffusion_gemma` is the interesting case: its modality set is a `var`, not a
`let`, and honestly so. The type is built as the text engine and *gains* its
vision lane when `installVision` attaches a tower — there is no separate
multimodal class to carry a fixed set.

## Measured, not asserted

| bundle | spared by a text-only load |
|---|---|
| Muse Glimmer 30B | 1,921,845,760 params — ≥3.84 GB at fp16 |
| Mistral 3 (real bundle) | 428,472,320 params — ~0.80 GiB |

Mistral's is smaller because Pixtral's tower is a flat ~0.75 GiB whatever the
text model's size. Same win, honestly smaller.

For Muse Glimmer the language tower is verified **identical** in both loads, so
the narrowing removes the vision tower and nothing else.

## Verification

- Real-bundle tests for both families; they skip rather than fail where the
  bundle is absent.
- **Mutation-checked.** Building the tower regardless of requested modalities,
  and not dropping vision weights in `sanitize`, each fail the suite.
- **A mutation that did NOT fail, and what it exposed.** Building Gemma 3's tower
  regardless of the request passed the entire suite — not because the assertion
  was weak, but because no `gemma3` bundle exists on the machine, so every
  `gemma3` assertion silently skipped. Every test here is bundle-gated, and a
  gate that closes quietly is indistinguishable from a passing test. There is now
  a test that **prints which families were exercised** and fails if none were;
  the same mutation re-run against `gemma4` (11 bundles) and `qwen3_5` (21) fails
  as it should. **The loop is now closed on `gemma3` itself**: with a multimodal
  `gemma-3-4b` present, that original mutation fails on *"a narrowed instance
  allocates no vision parameters, in every family"* — the very test it had slipped
  past. Exercised: `diffusion_gemma`, `gemma3`, `gemma4`, `gemma4_unified`,
  `mistral3`, `muse_glimmer`, `qwen3_5`, `qwen3_5_moe`. Only `ministral3` lacks a
  bundle, and it shares `mistral3`'s code paths entirely.
- **Declaring a narrower set and building a narrower model are different claims**,
  and only the second saves anything, so the suite asserts the parameter set and
  not merely the declaration.
- `sanitize` drops vision weights a text-only instance cannot receive, rather
  than rehoming them onto modules that were never built — the same stance Gemma4
  already takes with `audio_tower.*` on audio-less bundles.
- Existing callers are untouched: the plain `init(_ config:)` still builds
  everything the config offers, and is non-throwing by construction rather than
  by argument.

---

**Now includes the Qwen4Exp adoption (previously proposed separately as #321).**

They were split so the API and each adopter could be reviewed apart. That split does not survive as a
MERGE order: with the capability API in place and `Qwen4Exp` still on the old construction path, the
tree does not compile —

```
Qwen4Exp.swift: error: value of optional type 'Qwen35Configuration.VisionConfiguration?'
```

— so landing the API first would break `main` for however long the follow-up took. Folding the
adoption in keeps every commit on `main` buildable, which is the property that actually matters here.

Verified on current `main`: the package builds, and all five suites these patches add or touch pass —
`ModelConstructionModalities` (10), `UniversalCapabilityConstruction` (10),
`Mistral3CapabilityConstruction` (7), `MuseGlimmerCapabilityLoad` (4) and `Gemma4VLM` (5).

### DIFF
--- a/Libraries/MLXLLM/Models/DiffusionGemma.swift
+++ b/Libraries/MLXLLM/Models/DiffusionGemma.swift
@@ -579,7 +579,15 @@ class DiffusionGemmaInnerModel: Module {
 
 // MARK: - Top-level model
 
-public class DiffusionGemmaModel: Module, LLMModel {
+public class DiffusionGemmaModel: Module, LLMModel, ModalityBearing {
+    /// What this instance actually carries.
+    ///
+    /// Unlike the other converted families this is a `var`, and honestly so: this type is built
+    /// as the text engine and GAINS its vision lane later, when `installVision` attaches a tower.
+    /// Its VLM creator returns this same type — there is no separate multimodal class — so the
+    /// modality set has to follow the installation rather than the initialiser.
+    public private(set) var modalities: Set<ModelRuntimeRequestModality> = [.text]
+
 
     @ModuleInfo var model: DiffusionGemmaInnerModel
 
@@ -746,6 +754,7 @@ public class DiffusionGemmaModel: Module, LLMModel {
             @escaping (LMInput, DiffusionGemmaModel) throws
                 -> (embeddings: MLXArray, visionBlockIds: MLXArray)?
     ) {
+        modalities.insert(.vision)
         model.encoder.visionTower = tower
         model.encoder.embedVision = embedder
         visionPromptEmbedder = promptEmbedder
--- a/Libraries/MLXLMCommon/ModelConstructionModalities.swift
+++ b/Libraries/MLXLMCommon/ModelConstructionModalities.swift
@@ -1,0 +1,124 @@
+// Copyright © 2026 Osaurus AI. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+import Foundation
+
+/// Construction-time counterpart to `ModelRuntimeCapabilitySnapshot.validate(request:)`.
+///
+/// Two different questions are being asked about the same bundle, and they are deliberately not
+/// merged:
+///
+/// - `ModelRuntimeCapabilitySnapshot` answers *"is this model trained for a modality"*. It merges
+///   an explicit stamp, a structural probe, and coarse `modality` tokens, and it is honest about
+///   the gap between them — hence `.unknown`. It gates a generation request.
+/// - This file answers *"which towers can this configuration instantiate, and which of those does
+///   the caller want built"*. It is structural, local to the config, and has no third state: a
+///   config either carries a vision section or it does not.
+///
+/// The two can legitimately disagree — a bundle may declare `supportsVision` while shipping a
+/// config with no vision section, which is a broken conversion rather than a policy question — so
+/// keeping them separate preserves that signal instead of averaging it away. They share the
+/// `ModelRuntimeRequestModality` vocabulary so no caller has to translate between them.
+///
+/// Nothing here implies `.text`. A model whose only lanes are audio-in and audio-out is unusual but
+/// not incoherent, and a default that quietly adds text would misdescribe it.
+/// A consequence worth stating, because it changes what bundles are worth SHIPPING.
+///
+/// Once a multimodal bundle can be constructed text-only, a text-only DERIVATIVE of that same
+/// checkpoint has almost no reason to exist. It saves disk only for someone who never once wants
+/// the vision lane; anyone who alternates now stores two copies of the same text tower to avoid
+/// allocating a tower they could simply not build. The saving is small either way — on the 30B
+/// Muse Glimmer the vision half is 1.92B of 29.8B parameters, about 6%.
+///
+/// The distinction that survives is between a derivative and a RELEASE. A natively text-only
+/// model in the same family is not the multimodal one with the tower removed — it is separately
+/// trained, often at a different parameter count, and its weights are its own. Those still need
+/// the text-only path, and always will.
+///
+/// Which of the two the dual registrations actually serve is an empirical question, and at the
+/// time of writing the local answer is stark: across 63 bundles in the nine families registered
+/// in BOTH factories, every single one carries a `vision_config` AND a processor config. Not one
+/// is a text-only release. So on this machine those LLM-side entries are never reached by the
+/// routed loader — `loadModelContainer` tries the VLM factory first and it succeeds every time —
+/// and they serve only callers who reach for `LLMModelFactory.shared.load` directly to express
+/// "text only". That intent is exactly what the `requesting:` parameter here makes explicit,
+/// which is a better place for it than a second registration.
+///
+public struct UnconstructibleModalities: Error, CustomStringConvertible {
+    public let requested: Set<ModelRuntimeRequestModality>
+    public let constructible: Set<ModelRuntimeRequestModality>
+
+    public init(
+        requested: Set<ModelRuntimeRequestModality>,
+        constructible: Set<ModelRuntimeRequestModality>
+    ) {
+        self.requested = requested
+        self.constructible = constructible
+    }
+
+    /// The part of the request this configuration cannot satisfy.
+    public var excess: Set<ModelRuntimeRequestModality> {
+        requested.subtracting(constructible)
+    }
+
+    public var description: String {
+        "this configuration cannot build \(excess.modalityDescription) "
+            + "(requested \(requested.modalityDescription); constructible \(constructible.modalityDescription))"
+    }
+
+}
+
+/// An empty selection would build a model with no lanes at all — loadable, and useless.
+public struct EmptyModalitySelection: Error, CustomStringConvertible {
+    public let source: String
+
+    public init(_ source: String) {
+        self.source = source
+    }
+
+    public var description: String {
+        "\(source) named no modalities; a model with no lanes cannot be constructed"
+    }
+}
+
+extension Set where Element == ModelRuntimeRequestModality {
+    /// Stable, readable rendering in the canonical modality order.
+    public var modalityDescription: String {
+        isEmpty
+            ? "nothing"
+            : ModelRuntimeCapabilityRequest(modalities: self)
+                .sortedModalities.map(\.rawValue).joined(separator: ", ")
+    }
+
+    /// Narrow what a configuration can build down to what the caller asked for.
+    ///
+    /// - Parameters:
+    ///   - requested: the caller's subset, or `nil` for "everything this configuration offers".
+    ///   - constructible: what the configuration structurally permits.
+    /// - Returns: exactly the lanes to instantiate.
+    /// - Throws: `UnconstructibleModalities` if the request exceeds what can be built — refusing
+    ///   here rather than yielding a model that fails later inside `prepare()`, where the cause is
+    ///   no longer visible; `EmptyModalitySelection` if either side names nothing.
+    public static func resolveForConstruction(
+        requested: Set<ModelRuntimeRequestModality>?,
+        constructible: Set<ModelRuntimeRequestModality>
+    ) throws -> Set<ModelRuntimeRequestModality> {
+        guard !constructible.isEmpty else {
+            throw EmptyModalitySelection("this configuration")
+        }
+        guard let requested else { return constructible }
+        guard !requested.isEmpty else {
+            throw EmptyModalitySelection("the request")
+        }
+        guard requested.isSubset(of: constructible) else {
+            throw UnconstructibleModalities(requested: requested, constructible: constructible)
+        }
+        return requested
+    }
+}
+
+/// A model that reports which lanes it actually carries, as opposed to which its bundle advertises.
+public protocol ModalityBearing {
+    /// What this instance was built with — never wider than its configuration allowed.
+    var modalities: Set<ModelRuntimeRequestModality> { get }
+}
--- a/Libraries/MLXVLM/Models/DiffusionGemmaVLM.swift
+++ b/Libraries/MLXVLM/Models/DiffusionGemmaVLM.swift
@@ -49,11 +49,45 @@ public struct DiffusionGemmaVLMConfiguration: Codable, Sendable {
 /// tower, install the vision modules + the prompt-embedding splice closure.
 /// Text-only bundles (no `vision_config` / `image_token_id`) come back as
 /// the plain text engine.
+/// Which lanes this configuration can actually build.
+///
+/// Note the second condition: a vision tower is only installable when the bundle ALSO declares an
+/// `image_token_id`, because without it there is nowhere to splice the features. A bundle with a
+/// `vision_config` and no image token is not a vision bundle, and saying so here keeps the
+/// declaration honest rather than optimistic.
+public func diffusionGemmaConstructibleModalities(
+    of config: DiffusionGemmaVLMConfiguration
+) -> Set<ModelRuntimeRequestModality> {
+    config.visionConfig != nil && config.core.imageTokenId != nil ? [.text, .vision] : [.text]
+}
+
+/// - Parameter requesting: the caller's subset, or nil for "everything this config offers".
+///   Asking for a lane the config cannot build throws rather than yielding a model that fails
+///   later at `prepare()`.
+public func makeDiffusionGemmaVLM(
+    _ config: DiffusionGemmaVLMConfiguration,
+    requesting: Set<ModelRuntimeRequestModality>?
+) throws -> DiffusionGemmaModel {
+    let resolved = try Set.resolveForConstruction(
+        requested: requesting,
+        constructible: diffusionGemmaConstructibleModalities(of: config))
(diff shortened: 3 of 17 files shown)


jjang-ai:
Deep architectural review: **please keep this draft; there are correctness and integration blockers before the capability-construction design can merge.** The goal—avoiding unused tower allocation—is valuable, but the current API conflates three different things:

1. what media a request contains;
2. what components that request requires;
3. what the configuration can construct.

That conflation produces invalid-but-accepted requests and prevents the feature from affecting the routed factory used by normal Osaurus loads.

# Merge blockers

## 1. The production factory never consumes `requesting:`

The PR adds `init(_:requesting:)`, but does not change `VLMTypeRegistry`, `ModelTypeRegistry`, the routed factory, or their callers. Registry creators remain `(Data) throws -> LanguageModel` and still call ordinary initializers such as:

```swift
"qwen3_5": create(Qwen35Configuration.self, Qwen35.init)
"muse_glimmer": create(MuseGlimmerConfiguration.self, MuseGlimmer.init)
"gemma4": create(Gemma4Configuration.self, Gemma4.init)
```

Those ordinary initializers intentionally build every modality offered by the config. Therefore the advertised RAM saving currently exists only for direct constructors in tests/custom callers—not normal routed Osaurus loading.

Please thread the selection through the actual path. One possible backward-compatible shape:

```swift
public typealias ModelCreator =
    (_ configuration: Data, _ request: ModelConstructionRequest?) throws -> any LanguageModel

public func createModel(
    configuration: Data,
    modelType: String,
    constructionRequest: ModelConstructionRequest? = nil
) throws -> sending LanguageModel
```

Existing callers passing `nil` retain “build everything.” A narrowed Osaurus/text-only caller must reach the family constructor through this route. The exact API can differ; the acceptance criterion is that a routed load demonstrably allocates the narrower component set.

## 2. Public designated initializers bypass every invariant

For example, Qwen35 exposes both the validated `requesting:` initializer and public:

```swift
public init(_ config: Qwen35Configuration,
            modalities: Set<ModelRuntimeRequestModality>)
```

The latter accepts empty and unsupported sets, bypasses `resolveForConstruction`, and records the unchecked value in `model.modalities`. The same pattern appears across families and in `makeDiffusionGemmaVLM(...modalities:)`.

Please expose only the validated public path. The unchecked constructor should be private/internal and receive a validated/resolved plan that cannot be forged accidentally:

```swift
struct ResolvedConstructionPlan {
    let requested: Set<ModelRuntimeRequestModality>
    let components: Set<ModelComponent>
}

public convenience init(
    _ config: Configuration,
    requesting: Set<ModelRuntimeRequestModality>? = nil
) throws {
    let plan = try Self.resolveConstruction(config, requesting: requesting)
    self.init(config, resolvedPlan: plan) // internal/private
}
```

Add tests proving no public path can construct an empty or unsupported instance.

## 3. Qwen35/Muse video-only requests pass validation but omit the required tower

Both families declare `[.text, .vision, .video]`, but construct their shared media tower only when `modalities.contains(.vision)`. Because the new contract explicitly says `.text` is not implied, these requests are currently accepted:

```swift
try Qwen35(config, requesting: [.video])
try MuseGlimmer(config, requesting: [.text, .video])
```

They report video capability but build no vision/video tower. This is precisely the late `prepare()` failure the PR says it prevents.

Resolve dependencies separately from requested lanes:

```swift
if requested.contains(.vision) || requested.contains(.video) {
    components.insert(.visionTower)
}
```

Do not silently add `.vision` to the *reported request modalities* unless that is the intended API. A video request requires the physical vision tower; it does not necessarily mean the caller requested an image lane.

Required tests for Qwen35, Muse and Qwen4Exp downstream:

| request | accepted | reported modalities | required component |
|---|---:|---|---|
| nil | yes | config default | full tower set |
| text | yes | text | language only |
| vision | yes | vision (plus text only if required by model contract) | vision tower |
| video | yes | video | shared vision/video tower |
| text+video | yes | text+video | language + shared tower |
| audio | no | — | explicit construction error |
| empty | no | — | explicit construction error |

## 4. Gemma4 unified audio is supported input but excluded from `constructibleModalities`

`gemma4_unified_audio` processes raw audio through the encoder-free chunking/projection path. The current resolver inserts `.audio` only for a conformer tower. Thus a valid unified-audio request can be rejected because there is “no tower,” even though the model has a proven audio processing path.

This demonstrates why request modalities cannot equal optional towers. Suggested component mapping:

```swift
case conformerAudio:
    request .audio -> [.audioTower, .audioProjection]
case unifiedRawAudio:
    request .audio -> [.rawAudioProjection] // no conformer tower
```

`constructibleModalities` should answer which request modalities can succeed; a separate construction plan should answer which physical modules are needed.

## 5. The language core is always built while `.text` is explicitly not implied

Requests like `[.vision]` and `[.audio]` still construct the language model, yet `model.modalities` can omit `.text`. Decide and encode one coherent contract:

- if these architectures cannot execute a useful request without the language core, require/imply text at request validation; or
- keep requested modalities literal, but report physical components separately and document that the language core is a mandatory internal dependency.

Currently “builds exactly the lanes named by `modalities`” is not true.

## 6. Optional `vision_config` changes routed model selection

Today the routed loader tries VLM first and falls through to LLM on an error. Making VLM configuration decoding succeed without `vision_config` means some text-only/third-party bundles that previously reached the LLM implementation can now remain on the VLM implementation. That can change sanitizer, template, processor and cache behavior independently of tower allocation.

Please add routed-loader tests with fixtures for each dual-path family:

- multimodal config + default request selects full VLM construction;
- multimodal config + text request selects the intended narrowed construction;
- vision-stripped config selects the explicitly intended implementation;
- malformed VLM config does not get silently reclassified because of an unrelated decode error;
- output/sanitized-key parity against the previous LLM route is demonstrated where routing changes.

Consider narrowing fallthrough to an explicit `unsupportedModelType`/unsupported-shape result instead of swallowing any VLM construction error, although that may be best as a separate prerequisite PR.

## 7. Unit tests are not hermetic

`UniversalCapabilityConstructionTests` searches `~/Library/MLModels` and asserts at least one local vision bundle exists. Standard CI machines cannot be assumed to contain those bundles. The current checks are all `SKIPPED`, so this portability problem has not been exercised.

Please split proof into:

### Hermetic unit suite

Use checked-in minimal JSON fixtures or programmatic configurations for every family. It must run with an empty home/model directory and cover every request/dependency/error row deterministically.

### Explicit real-bundle integration suite

Bundle-gated tests may scan a configured model root, but must emit machine-readable `PASS/SKIP/FAIL` per family. Absence must be `SKIP`, never make the unit suite look complete. Require an explicit environment variable for the model root rather than hard-coding `~/Library/MLModels`.

Also, the current `visionKeys` predicate does not appear to include Gemma4's `embed_vision` name, so “no vision parameters” may miss a remaining vision projection. Assert expected exact parameter-prefix sets per family, not a small substring allowlist.

# Required proof before ready-for-review

## Source/unit gates

- Final-head macOS build and focused tests actually execute; GitHub currently reports every check as `SKIPPED`.
- Empty/unsupported selections are impossible through every public API.
- Dependency closure is tested for video and both Gemma4 audio implementations.
- Routed factory receives and honors the construction request.
- Default/no-request behavior is byte-for-byte behavior-compatible with current main.
- Sanitizers retain every language key and discard only weights for components absent from the resolved plan.
- Clean CI with no local model directory.

## Real runtime matrix

For at least Gemma4, Qwen35, Muse, Mistral3 and Qwen4Exp after #321 rebases:

1. default/full load;
2. text-only narrowed load;
3. real image turn where supported;
4. real video turn for Qwen35/Qwen4Exp/Muse where supported;
5. conformer and unified-audio rows for Gemma4 where available;
6. coherent multi-turn follow-up;
7. cache telemetry and prefix reuse across the follow-up;
8. Activity Monitor `phys_footprint` for full vs narrowed load;
9. TTFT and token/s whenever generation occurs;
10. no protocol-marker leakage, silent media drop, loop, or length-cap fake pass.

For the RAM claim, record both parameter-count deltas and live `phys_footprint`; parameter enumeration alone is not proof that the user-facing load path saved memory.

After vMLX lands, the downstream proof must repin all Osaurus dependency surfaces to the final SHA, build the exact Release app, ensure one app instance, select/load through the visible UI, and demonstrate the same modality/narrowing behavior there. API-only evidence is diagnostic, not GUI proof.

# Suggested integration sequence

1. Land #311 after its invariant tests execute.
2. Land #312 independently after caller parity proof.
3. Choose one owner for #313's Gemma4 policy and remove the duplicate from this branch after rebase.
4. Refactor this PR around a validated construction plan and wire it through the routed factory.
5. Make unit tests hermetic and run the real-bundle matrix.
6. Rebase #321 to its single Qwen4Exp commit and apply the corrected dependency contract.
7. Perform the Osaurus repin + Release GUI proof in a separate integration PR.

The capability is worth pursuing; the hold is about preventing a public contract that can describe impossible instances and a RAM optimization that normal routed callers cannot reach.

rcfa:
Thanks — this was the right call, and blocker 1 was the one that mattered most. Pushed a redesign; still draft, because 6 and 7 are not done.

## Blocker 1 — the routed factory

You were right, and it was worse than a gap: the modality-aware initialisers existed and were correct, and **nothing called them**. `ModelTypeRegistry` stored `(Data) throws -> any LanguageModel`, so the request was accepted by the public API and dropped before reaching a model.

Collapsed to one creator shape — no second modality-aware dictionary, because a parallel path is how the request came to be dropped:

```swift
public typealias ModelCreator =
    (Data, Set<ModelRuntimeRequestModality>?) throws -> any LanguageModel
```

`requesting:` defaults to `nil`, so existing call sites are untouched. Selecting families say so in the table: `createSelecting(Qwen35Configuration.self, Qwen35.init(_:requesting:))` — a distinct name rather than an overload, so the table records which families consult the request.

`RoutedFactoryRequestDispatchTests` goes through the **registry**, not an initialiser: a test that constructs directly passes just as happily with the table unwired, which is how this shipped. And because that is the failure mode, `SelectingFamiliesAreWiredSourceCoverageTests` states it about the table itself — a family implementing `ModelComponentMapping` must not be registered through the plain creator. Unwiring one entry **compiles with zero errors**; the guard names it.

## Blockers 2–5

2. The real initialiser is `private init(_:plan:)`. A `ResolvedConstructionPlan` can only come from `resolveConstruction`, so no caller can hand it an empty or unsupported selection.
3. Fixed. `components(for:of:)` states the dependency once — `requested.contains(.vision) || requested.contains(.video)` requires the tower — so a video-only request allocates it.
4. Fixed. `constructibleModalities` inserts `.audio` whenever `audioConfig != nil` (both implementations); `.audioTower` is added only for the conformer path, gated on `hasConformerAudioTower`.
5. Took your second option. Lanes and modules are now separate types: `ResolvedConstructionPlan.components` always contains `.languageCore` and is documented as a mandatory internal component, while the requested set stays literal. `modalities` reports what the instance SERVES, from a family-declared `servedModalities(by:of:)` with **no default implementation** — a blanket "…and always text" describes a speech-to-speech model wrongly, and there is a test for that case. Components register what they provide as they are constructed, so a tower that was not built cannot contribute to the report.

## Not done

6 (routing selection) and 7 (hermetic tests) are not addressed. Keeping this draft until they are — I did not want to imply otherwise by promoting it.

On `SKIPPED` checks: that is #332. The workflow guards on the repository name and still names the upstream Apple repo, so every check short-circuits — it affects your own PRs too, not just these.

rcfa:
**Correction to my previous comment: the redesign is not on this branch yet.** I said "pushed a redesign" — that was wrong, and I would rather say so than let you review against something that is not there. This branch is still the earlier four commits; `ModelTypeRegistry` here still stores `(Data) throws -> any LanguageModel` and there is no `RoutedFactoryRequestDispatchTests`.

The work is real and verified locally — the description of blockers 1–5 in my previous comment is accurate as a description of the code — but it sits on a branch that will not build against current main. Three of our patches (the Qwen35 diagnostic split, the `VMLX_` migration, and the family-construction one under this PR) touch files that moved in the commits since our pin, and the conflicts are genuine rather than context drift.

Rebasing them is the actual blocker, and I am not going to push a half-rebased branch to make the diff look right — an earlier attempt at exactly that produced patches that applied cleanly while silently carrying each other's changes, which would have put unrelated work into these PRs.

So: please treat blockers 1–5 as unaddressed on this branch until I push. I will comment again when the branch actually carries it, and it stays draft until then. Sorry for the noise.

One thing that is independently on main and worth having regardless: #358 — `Qwen4ExpFusedAffineMoE.swift` does not type-check on Swift 6.4, which is what surfaced the rebase in the first place. It blocks the build before anything in this PR is reached.

rcfa:
Follow-up to my correction: the work is now on a branch, and it is **#359** rather than this one.

Rebasing onto current main was the actual blocker, and it is resolved — but by building the branches directly on main rather than by rebasing the patch files, which is what corrupted them on my first attempt. #359 (`feat/construction-plan-lanes-vs-components`) carries blockers 1–5; this PR stays as the earlier, narrower step.

Two build dependencies turned up while verifying it, and they affect **every** PR here, not just mine:

- **#358** — `Qwen4ExpFusedAffineMoE.swift` does not type-check on Swift 6.4. It became reachable when the hardcoded shape text was replaced by three interpolations.
- **#319** — the same defect in `Qwen35.swift`, which this PR's own branch hits.

With both applied, the rebased branch compiles on current main; without either, the build fails before reaching anything under review. That is consistent with your note about checks reporting `SKIPPED` — once #332 makes CI actually run, these would surface as compile failures rather than as skipped checks.

Blockers 6 (routing selection) and 7 (hermetic tests) remain open, so #359 is draft.

PR #321, opened 2026-08-28
Qwen4Exp: build the towers the caller asked for, like the other families

rcfa:
> **Depends on #315 — review only the LAST commit.**
>
> The first four commits here *are* #315. They are carried in this branch because a
> PR's base must be a branch in this repository, and ours live in a fork, so GitHub
> cannot stack this on #315 directly. Rather than show you a diff with its own
> foundation missing, the dependency rides along. Once #315 merges these four
> collapse to no-ops on rebase and this becomes a single-commit PR.
>
> #315 introduces the convention; this applies it to the first family that arrived
> after it existed.

`qwen4_exp` is the first multimodal family to arrive **after** the modality
convention existed, which makes it the real test of whether #315 produced a
convention or nine one-offs. It gets the same three pieces every converted
family has:

| piece | question it answers |
|---|---|
| `constructibleModalities(of:)` | what could this configuration build? |
| `init(_:requesting:)` | what did the caller ask for — is it available? |
| `ModalityBearing` | what does this instance actually carry? |

## The video lane is claimed, and that is a real claim

```swift
config.base.visionConfiguration != nil ? [.text, .vision, .video] : [.text]
```

`prepare` consumes `input.video` and threads video frames into the tower — the
same shape as `qwen3_5`. The Gemma and Pixtral families are image-only and
correctly do **not** claim it. Declaring the lane is only worth doing because
the declaration can be wrong, so it is asserted per family rather than assumed.

## Narrowing skips the tower, it does not just report a smaller set

`visionModel` becomes optional, and three sites follow:

1. **`prepare` refuses media** with a message naming the instance's modalities,
   rather than embedding the prompt without the media it was handed. A silently
   media-free answer looks like a bad model, not a bad call.
2. **`sanitize` drops `visual.` weights** when no tower exists to receive them —
   mirroring the `mtp.` guard directly above it. A multimodal bundle loaded
   text-only still *ships* its vision weights.
3. Two `visionConfiguration` reads become optional-safe.

## Tested in the shared suite, which is the point

The four cross-family tests from #315 now cover `qwen4_exp` too — it is added to
the family list rather than given a bespoke suite. One family-specific test
asserts the parts that are its own:

```swift
#expect(Qwen4Exp.constructibleModalities(of: cfg) == [.text, .vision, .video])
#expect(try Qwen4Exp(cfg, requesting: [.text]).modalities == [.text])
// narrowing must SKIP the tower, not merely declare a smaller set
#expect(visionKeys.isEmpty)
// and the weights it cannot receive are dropped rather than handed to nothing
#expect(out.keys.first { $0.contains("visual") } == nil)
```

## Coverage is reported, not assumed

```
exercised        : diffusion_gemma, gemma3, gemma4, gemma4_unified, mistral3,
                   muse_glimmer, qwen3_5, qwen3_5_moe, qwen4_exp
no bundle at all : ministral3
```

These assertions are bundle-gated, and a gated assertion that never runs looks
exactly like one that passed — so the suite states which families it actually
exercised on the machine it ran on.

### DIFF
--- a/Libraries/MLXLLM/Models/DiffusionGemma.swift
+++ b/Libraries/MLXLLM/Models/DiffusionGemma.swift
@@ -579,7 +579,15 @@ class DiffusionGemmaInnerModel: Module {
 
 // MARK: - Top-level model
 
-public class DiffusionGemmaModel: Module, LLMModel {
+public class DiffusionGemmaModel: Module, LLMModel, ModalityBearing {
+    /// What this instance actually carries.
+    ///
+    /// Unlike the other converted families this is a `var`, and honestly so: this type is built
+    /// as the text engine and GAINS its vision lane later, when `installVision` attaches a tower.
+    /// Its VLM creator returns this same type — there is no separate multimodal class — so the
+    /// modality set has to follow the installation rather than the initialiser.
+    public private(set) var modalities: Set<ModelRuntimeRequestModality> = [.text]
+
 
     @ModuleInfo var model: DiffusionGemmaInnerModel
 
@@ -746,6 +754,7 @@ public class DiffusionGemmaModel: Module, LLMModel {
             @escaping (LMInput, DiffusionGemmaModel) throws
                 -> (embeddings: MLXArray, visionBlockIds: MLXArray)?
     ) {
+        modalities.insert(.vision)
         model.encoder.visionTower = tower
         model.encoder.embedVision = embedder
         visionPromptEmbedder = promptEmbedder
--- a/Libraries/MLXLLM/Models/Gemma4Text.swift
+++ b/Libraries/MLXLLM/Models/Gemma4Text.swift
@@ -890,30 +890,49 @@ public class Gemma4TextModel: Module, LLMModel {
                 continue
             }
 
-            // Remap JANG expert naming to match module tree:
-            // JANG uses:          switch_mlp.{gate,up,down}_proj.*
-            // mlx-community uses: experts.switch_glu.{gate,up,down}_proj.*
-            // Module tree expects: experts.switch_glu.{gate,up,down}_proj.*
-            if newKey.contains(".switch_mlp.") {
-                newKey = newKey.replacingOccurrences(of: ".switch_mlp.", with: ".experts.switch_glu.")
-            }
+            newKey = Self.remappingSwitchMLP(newKey)
 
             processedWeights[newKey] = value
         }
 
-        // Trim vocab-dimension tensors to match config
-        let expectedVocab = config.vocabSize
-        for key in [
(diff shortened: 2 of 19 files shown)


jjang-ai:
Dependency review note: please keep this blocked on the #315 redesign rather than polishing the Qwen4Exp-specific commit against the current contract.

The same concrete video dependency bug applies here: `qwen4_exp` advertises `.video`, but a request containing video without `.vision` must still allocate the shared vision/video tower. Under the current #315 vocabulary, requested lanes and physical components are conflated.

After #315 resolves that, the Qwen4Exp-only commit should be rebased and prove at minimum:

- empty and unsupported requests fail through every public constructor;
- text-only construction has no Qwen4Exp visual/tower/projection parameters;
- image-only and video-only requests both allocate the shared media tower;
- sanitizer drops `visual.*` only for a plan without that tower;
- ordinary/default construction is unchanged;
- the routed factory—not only a direct initializer—honors the narrowed request;
- real Qwen3.8 Next/Flash text, image and video turns succeed in the exact Osaurus Release UI;
- text → tool → tool-result → follow-up remains coherent, including the known Qwen4Exp agent/tool-result path;
- MTP off/auto/manual-depth rows record effective strategy, accepted/drafted tokens, TTFT, token/s and cache counters;
- full versus text-only `phys_footprint` is measured live.

The PR description correctly says the first four commits are #315. Once #315 lands, please rebase this to one Qwen4Exp commit and rerun all proof on that final head. Current GitHub checks are `SKIPPED`, so none of the present verification text is current CI evidence.

rcfa:
Understood — keeping this blocked on #315 rather than polishing it against the old base.

The video dependency bug you name is fixed in the #315 redesign: `components(for:of:)` states the dependency once, as `requested.contains(.vision) || requested.contains(.video)` requiring the tower, so a video-only request allocates it instead of validating and then omitting it.

Your proof list is the right one, and the last item — *the routed factory, not only a direct initializer, honors the narrowed request* — turned out to be the core defect rather than a test gap: the registry stored a creator that took no request, so the modality-aware initialisers were unreachable. That is fixed in #315 and covered by a test that goes through the registry, plus a source guard that fails if a selecting family is ever registered through the plain creator.

I will rebase this onto the corrected #315 and prove the full list before marking it ready.

rcfa:
Folding this into #315 rather than landing it separately.

The split was for review — the capability API in one PR, each adopter in the next. It does not survive as a *merge* order. With the API in place and `Qwen4Exp` still on the old construction path, the tree does not compile:

```
Qwen4Exp.swift: error: value of optional type 'Qwen35Configuration.VisionConfiguration?'
```

So merging #315 first would leave `main` broken until this one followed. #315 now carries this change too, and I verified the combined set builds on current `main` with all five affected suites passing. Nothing here is dropped — closing so the same change isn't proposed twice.

PR #323, opened 2026-08-28
JangLoader: decode the sampling defaults bundles actually carry

rcfa:
> **Draft on purpose.** The decode fix here is not in question; whether the
> runtime should *apply* these defaults is, and that is your call. See the
> decision at the bottom.

`chat.sampling_defaults` is decoded into three fields and read by **nobody**.
The doc comment promises otherwise:

```swift
/// Consumers (BatchEngine / Evaluate) may apply these when the
/// caller doesn't pass explicit sampler params.
```

No consumer does. But the decode is wrong independently of that, and *that* part
isn't a policy question.

## What 14 local JANG bundles actually declare

| key | bundles | decoded today |
|---|---|---|
| `temperature` | 14 | ✅ |
| `top_p` | 14 | ✅ |
| **`top_k`** | **14** | ❌ |
| `repetition_penalty` | 9 | ❌ |
| `min_p` | 7 | ❌ |
| `presence_penalty` | 6 | ❌ |
| `max_tokens` | 2 | ❌ |
| **`max_new_tokens`** | **0** | ✅ |

The one field the struct models that no bundle uses is decoded. `top_k`, which
**every** bundle carrying the block sets, is dropped.

And every one of these already has a destination in `GenerateParameters` —
`temperature`, `topP`, `topK`, `minP`, `repetitionPenalty`, `presencePenalty`.
The loader was discarding values the runtime is fully able to act on.

## Why partial application is the crux

Every observed bundle that sets `temperature` also sets `top_k`. A runtime that
honoured the first and dropped the second would produce a sampler configuration
**the vendor never specified and nobody validated** — plausibly worse than
honouring none of it.

That is the shape of today's three-field decode. It is the argument for fixing
the decode *even if* the answer to "apply automatically?" turns out to be no.

## What this PR does, and does not

- completes the decode to the keys bundles carry;
- adds `applied(to:)`, overlaying only the declared knobs onto a caller's
  `GenerateParameters`;
- **calls it from nowhere.** Behaviour is unchanged by this PR.

`applied(to:)` is a caller's method rather than something the loader performs,
for a concrete reason: `GenerateParameters` stores its sampler fields
non-optionally, so "the caller left this unset" and "the caller set it to the
default value" are indistinguishable from inside the loader. Anything automatic
would have to overwrite deliberate choices or guess.

Provenance keys (`source`, `mode`, `temperature_note`, plus the generation_config
residue `DeepSeek-V4-Flash-0731-JANG` copies in wholesale — `_from_model_config`,
`bos_token_id`, `transformers_version`) are ignored by omission, as before.
`isEmpty` reports a block that carried nothing actionable.

## The decision this needs from you

1. **Apply automatically** when the caller passes no sampler params — matches the
   doc comment's promise, changes generation for every bundle with the block, and
   needs a way to tell "unset" from "set to the default".
2. **Expose only** — this PR as it stands. Osaurus (or any caller) opts in per
   request; vmlx's own defaults are untouched.
3. **Neither** — then the block is documentation, and `applied(to:)` should
   probably go, though the decode fix still stands on its own.

I have implemented (2) because it is the one that is correct under all three
answers. Say which you want and I will finish it in this PR.

### DIFF
--- a/Libraries/MLXLMCommon/JangLoader.swift
+++ b/Libraries/MLXLMCommon/JangLoader.swift
@@ -725,14 +725,55 @@ public struct JangChatToolCalling: Sendable, Equatable {
 public struct JangChatSamplingDefaults: Sendable, Equatable {
     public let temperature: Float?
     public let topP: Float?
+    public let topK: Int?
+    public let minP: Float?
+    public let repetitionPenalty: Float?
+    public let presencePenalty: Float?
+
+    /// The generation length cap. Bundles spell this `max_tokens`; `max_new_tokens` is accepted as
+    /// an alias because this type has always named it that, though no bundle observed uses it.
     public let maxNewTokens: Int?
 
     public init(
-        temperature: Float? = nil, topP: Float? = nil, maxNewTokens: Int? = nil
+        temperature: Float? = nil, topP: Float? = nil, topK: Int? = nil, minP: Float? = nil,
+        repetitionPenalty: Float? = nil, presencePenalty: Float? = nil, maxNewTokens: Int? = nil
     ) {
         self.temperature = temperature
         self.topP = topP
+        self.topK = topK
+        self.minP = minP
+        self.repetitionPenalty = repetitionPenalty
+        self.presencePenalty = presencePenalty
         self.maxNewTokens = maxNewTokens
+    }
+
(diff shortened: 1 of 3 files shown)


rcfa:
Tracking issue: #324 (the detail lives here, not there).

PR #332, opened 2026-08-29
CI: point the pull-request guard at this repository

rcfa:
`lint` and `mac_build_and_test` are gated on the guard inherited from Apple's
mlx-swift:

```yaml
if: github.repository == 'ml-explore/mlx-swift'
```

That is never true in `osaurus-ai/vmlx-swift`, so both jobs skip
unconditionally, and `linux_build_cmake_container` / `linux_build_cmake_cuda`
skip with them through `needs: lint`.

The last **20** workflow runs all concluded `skipped`. The workflow has never
executed in this repository — not on fork PRs, and not on maintainer PRs
(#329 shows the same four jobs `skipping`).

## Please read before merging — this is not purely cosmetic

Repointing the constant is the whole fix *textually*, but the repository
currently has **0 registered self-hosted runners**, and:

| job | `runs-on` | after this change |
|---|---|---|
| `lint` | `ubuntu-latest` | **runs** — GitHub-hosted |
| `mac_build_and_test` | `[self-hosted, macos]` | **queues indefinitely** unless a runner is registered |
| `linux_build_cmake_container` | `ubuntu-22.04` matrix | starts running (was gated behind `lint`) |
| `linux_build_cmake_cuda` | `gpu-t4-4-core` | **queues indefinitely** — self-hosted |

So merging this alone trades two silent skips for two jobs that never start,
which is arguably worse. That is your call and your infrastructure, so I have
left it draft rather than guessing.

## Options

1. **Merge as-is and register runners** — a self-hosted macOS runner and a
   `gpu-t4-4-core` host. Best outcome, most setup.
2. **Merge with `mac_build_and_test` on a GitHub-hosted runner** — change
   `runs-on: [self-hosted, macos]` to `runs-on: macos-15`. This is the one that
   actually gets the package built and tested on every PR without any
   infrastructure; the cost is GitHub-hosted macOS minutes. Say the word and I
   will add that commit here.
3. **Enable `lint` only** — keep explicit guards on the three jobs needing
   runners you do not have, so nothing queues. Smallest safe step; gives one
   executing check.
4. **Close this** — if CI is intentionally dormant, that is fine too, but then
   "please run real checks" cannot be a review requirement, and I will keep
   posting local runs labelled as local.

I have no preference beyond wanting review requests I can actually satisfy.
Happy to push whichever variant you pick.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -7,7 +7,7 @@ permissions:
 
 jobs:
   lint:
-    if: github.repository == 'ml-explore/mlx-swift'
+    if: github.repository == 'osaurus-ai/vmlx-swift'
     runs-on: ubuntu-latest
     container:
       image: swift:6.2-rhel-ubi9
@@ -71,7 +71,7 @@ jobs:
 
   mac_build_and_test:
     needs: lint
-    if: github.repository == 'ml-explore/mlx-swift'
+    if: github.repository == 'osaurus-ai/vmlx-swift'
     runs-on: [self-hosted, macos]
     steps:
       - uses: actions/checkout@v6


rcfa:
Update — this PR produced the first executed CI run this repository has ever
had, and the result changes the recommendation.

Because `pull_request` workflows run from the PR's own head, the guard fix took
effect on this branch. `lint` **ran for 9m16s and failed**; the other three jobs
then reported `skipping` because they all carry `needs: lint`.

## What lint found

`swift-format` wants to rewrite **595 files**. Only 31 of them are files any of
my branches touch, and the rest is the bulk of the tree — `Hub.swift`,
`HubApi.swift`, `Lexer.swift`, `Interpreter.swift`, `Billing.swift`,
`MLXDistributedCore/*`, the Jinja and vendored swift-transformers sources.

So the tree has never been swift-format clean, which is unsurprising given the
job has never run. This is not something my branches introduced.

## Why that matters for the choice

Every other job depends on `lint`. So enabling the guard as this PR does means
**nothing can pass until the tree is formatted** — including your own PRs. That
makes option 1 from the description worse than I described it, and I would not
merge this as it stands.

Revised options:

1. **Format the tree first**, then merge this. A mechanical `swift-format`
   pass over 595 files, best as its own PR with no logic changes. Clean end
   state; large diff; will conflict with anything in flight.
2. **Decouple build/test from style** — drop `needs: lint` from
   `mac_build_and_test` (and the Linux jobs), so a style failure does not gate
   correctness. Then this guard fix gives you a real build/test signal without
   the formatting debt blocking it. Still needs a macOS runner, or
   `runs-on: macos-15`.
3. **Enable `lint` only, non-blocking** — add `continue-on-error: true` so it
   reports without gating, and format opportunistically.
4. **Leave CI dormant.** Also fine — but then "please run real checks" cannot be
   a review requirement, and I will keep labelling my evidence as local.

My suggestion is 2, then 1 when convenient: correctness signal now, formatting
debt paid separately rather than as a precondition. Happy to prepare either.

I have left this draft and will not push further variants until you say which
shape you want.

rcfa:
rcfa asked whether the formatting could just be automated in CI rather than paid
off as a one-time debt. Three shapes, and I think the third is the one worth
having.

## 1. Auto-format and push back to the PR

An action that runs `swift-format format --in-place` and commits the result to
the PR branch. Works well for same-repo branches; for **fork** PRs the default
token is read-only, so it needs `pull_request_target` plus a write token — which
runs untrusted PR code with elevated permissions. I would not recommend it for a
repo that takes outside contributions.

## 2. Format the whole tree once, then keep lint blocking

Clean end state. One mechanical commit over 595 files, no logic change:

```bash
swift-format format --in-place --configuration .swift-format --recursive .
```

The cost is that it invalidates every in-flight patch and branch, mine included —
all 21 of my patches are line-context based and would need regenerating. Fine if
timed deliberately, unpleasant if it lands mid-review.

## 3. Lint only what the PR changed

The one I would actually suggest, because it makes the debt irrelevant instead of
requiring it be paid first:

```yaml
- name: Style check changed files
  run: |
    git fetch --no-tags --depth=1 origin "$GITHUB_BASE_REF"
    CHANGED=$(git diff --name-only "origin/$GITHUB_BASE_REF"...HEAD -- '*.swift')
    [ -z "$CHANGED" ] && exit 0
    swift-format lint --configuration .swift-format --strict $CHANGED
```

New and modified files must be clean; the other 595 stay as they are until
someone touches them, and get fixed naturally as files are edited. No flag day,
no mass conflict, and the rule is one a contributor can satisfy locally.

It also has a property the current setup lacks: it can actually pass today.

Combine with dropping `needs: lint` from `mac_build_and_test` and you get build
and test signal immediately, with style enforced going forward rather than
retroactively.

I am happy to prepare (3) — as a change to this PR or a separate one — but it is
your CI, so I will wait for a preference rather than pushing more variants.

rcfa:
Concrete evidence that this is costing real signal, since it is easy to read as cosmetic.

`.github/workflows/pull_request.yml` on main still says:

```yaml
  lint:
    if: github.repository == 'ml-explore/mlx-swift'
  …
  mac_build_and_test:
    if: github.repository == 'ml-explore/mlx-swift'
```

That is the Apple repository this was forked from, so the condition is false here and **every job skips**. Checked via the check-runs API on three PRs:

| PR | author | conclusions |
|---|---|---|
| #320 | rcfa | skipped, skipped, skipped, skipped |
| #356 | rcfa | skipped, skipped, skipped, skipped |
| #292 | jjang-ai | skipped, skipped, skipped, skipped |

So it is not just our PRs — yours get no build or test either.

Two consequences worth naming:

1. **It reads as failure.** A rollup of all-skipped is not success, so clients render it as a red ✗. It was reported to me as "PR #322 has a failing test"; it has no failing test, and no passing one either.
2. **It is hiding real breakage.** While rebasing our stack onto current main I found that `Qwen4ExpFusedAffineMoE.swift` does not type-check on Swift 6.4 (#358), and neither does `Qwen35.swift` (#319) — main does not build at all on that toolchain. CI would have caught both on the commit that introduced them. Related: #362, where a test only surfaces once the suite runs far enough to reach it.

This is a one-line change and it turns four permanently-grey checks into actual signal.

PR #337, opened 2026-08-29
Load per-expert Qwen3.5 MoE MTP weights

jjang-ai:
## Status

**PARTIAL / draft. Do not merge yet.** Source and synthetic forward contracts pass on `1725bf0f98b11f4c64640e4587aec72881892214`; the real Ornith-35B off/auto/depth matrix is still pending a clean model window.

## Reproduced boundary

`dealign.ai/Ornith-1.5-35B-A3B-UNCENSORED-JANG_4M` is a Qwen3.5-MoE VLM with one native-MTP layer and 256 experts. Its MTP shard contains exactly 2,304 routed projection tensors (`256 experts × gate/up/down × weight/scales/biases`). Current main stacks supported base expert layouts, but never traverses `mtp.layers.L.mlp.experts.E.*`; the runtime module expects `mtp.layers.L.mlp.switch_mlp.*`.

A focused pre-fix test on current main failed 10 key/shape assertions after the real MLX metallib was installed (`/private/tmp/ornith35-mtp-moe-before.log`). This is distinct from the already-fixed Qwen3.5 B=2 graph-offset crash.

## Change

- Stack complete per-expert affine MTP banks into the existing `SwitchGLU` layout.
- Materialize each stacked tensor at load time instead of retaining a lazy graph over every source mmap.
- Preserve already-stacked checkpoints unchanged.
- Leave incomplete source layouts intact so strict model update reports them; never silently drop expert tensors or silently disable MTP.
- Support both native checkpoint prefixes already recognized by the parent Qwen3.5 sanitizer.

## Current proof

- `Qwen35VLMMTPMoESanitizerTests`: **5/5 passed** (`/private/tmp/ornith35-mtp-moe-tests-forward.log`).
  - complete expert-bank stacking;
  - incomplete payload remains visible/fail-closed;
  - pre-stacked idempotence;
  - real Ornith index topology (`256 × 3 × 3` keys in the MTP shard);
  - synthetic strict update of the actual quantized module graph followed by a finite `nativeMTPForward`.
- Neighboring Qwen3.5 gated-delta/native-MTP suites: **16/16 passed** (`/private/tmp/ornith35-neighbor-tests.log`).

## Required before merge

- Matched current-main/current-head real bundle load evidence, including exact base rejected keys.
- Prove `--mtp off`; do not assume it loads.
- Live off, auto, and every supported explicit depth.
- Coherent multi-turn and tool-result continuation; MTP verify/accepted/rejected/AR telemetry.
- Stop/continue and restart/disk restore.
- TTFT, token/s, `phys_footprint`, prefix/paged/L2 and SSM/MTP companion cache counters.
- Exact-head Osaurus Release-app visual proof after downstream repin.

The heavy rows are deferred while an unrelated large-model Osaurus proof process is active; this lane will not kill or overlap it.

### DIFF
--- a/Libraries/MLXVLM/Models/Qwen35MoE.swift
+++ b/Libraries/MLXVLM/Models/Qwen35MoE.swift
@@ -58,7 +58,49 @@ public final class Qwen35MoE: Qwen35 {
             }
         }
 
+        for layer in 0 ..< config.textConfiguration.mtpNumHiddenLayers {
+            for prefix in [
+                "mtp.layers.\(layer).mlp",
+                "model.mtp_layers.\(layer).mlp",
+            ] {
+                stackAffineExpertsIfNeeded(prefix: prefix, weights: &remapped)
+            }
+        }
+
         return remapped
+    }
+
+    /// Some Qwen3.5-MoE checkpoints keep the base routed experts pre-stacked
+    /// but preserve the MTP head in Hugging Face's per-expert affine layout:
+    /// `mtp.layers.L.mlp.experts.E.{gate,up,down}_proj.{weight,scales,biases}`.
+    /// The runtime module is a `SwitchGLU`, so those tensors must become one
+    /// materialized expert bank per projection. Materializing at load avoids a
+    /// lazy stack retaining every source mmap through the first decode.
+    private func stackAffineExpertsIfNeeded(
+        prefix: String,
+        weights: inout [String: MLXArray]
+    ) {
+        for projection in ["gate_proj", "up_proj", "down_proj"] {
+            for suffix in ["weight", "scales", "biases"] {
+                let target = "\(prefix).switch_mlp.\(projection).\(suffix)"
+                if weights[target] != nil { continue }
+
+                let sourceKeys = (0 ..< config.textConfiguration.numExperts).map {
+                    "\(prefix).experts.\($0).\(projection).\(suffix)"
+                }
+                guard sourceKeys.first.flatMap({ weights[$0] }) != nil else { continue }
+                let tensors = sourceKeys.compactMap { weights[$0] }
+                guard tensors.count == config.textConfiguration.numExperts else {
+                    // Leave the incomplete source layout intact so strict model
+                    // update fails closed and reports the unsupported payload.
+                    continue
+                }
+                for key in sourceKeys {
+                    weights.removeValue(forKey: key)
+                }
+                weights[target] = loadTimeMaterializedStacked(tensors)
+            }
+        }
     }
 
     private func stackTurboQuantExpertsIfNeeded(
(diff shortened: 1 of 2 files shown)


jjang-ai:
Source-only RAM audit on current head `1725bf0f` (not live proof): the exact local Ornith-35B MTP shard contains 2,341 `mtp.*` tensors totaling 0.4425 GiB. The routed affine subset this PR stacks is 2,304 tensors: three 0.125 GiB packed-weight banks plus six 0.0078125 GiB scale/bias banks. `loadTimeMaterializedStacked` evaluates one target bank at a time, synchronizes, and clears the MLX cache; source keys for that bank are removed before assignment. Therefore this change can transiently hold the current source bank plus one target bank, but it does not imply a second 35B model-sized copy. The largest target bank is ~128 MiB and total resident MTP payload is ~453 MiB.

This lowers the source-level peak-risk concern, but the PR remains PARTIAL: only Activity Monitor `phys_footprint` during a real off/auto/explicit-depth load can prove the actual peak and post-load residency. Required live rows remain unchanged.

jjang-ai:
Exact-head live loader/forward proof for `1725bf0f98b11f4c64640e4587aec72881892214` (RunBench SHA-256 `7d78b5f3d87d3e18dc2039cedcc426aa8abeece8579ed0694429745a1d478221`):

- **MTP off:** loaded the real `dealign.ai/Ornith-1.5-35B-A3B-UNCENSORED-JANG_4M`, explicitly skipped 2,341 MTP tensors, generated 64 coherent tokens, 23.6 tok/s, 1,942 ms TTFT, 11.22 s load, 27,101 MB peak phys_footprint. Log: `/private/tmp/vmlx337-ornith35-off.log`.
- **MTP payload loaded, AR decode:** no unhandled expert keys, 64 coherent tokens, 23.7 tok/s, 1,016 ms TTFT, 2.79 s warm load, 27,532 MB peak. Log: `/private/tmp/vmlx337-ornith35-mtp-loaded-ar.log`.
- **Explicit d1 greedy:** real quantized MTP forward completed: 12 verify calls, accepted-by-depth d0=9/d1=3, 4 bonus, 8 rejected, 47 AR fallback, one cache refresh; 23.2 tok/s, 917 ms TTFT, 27,868 MB peak. Log: `/private/tmp/vmlx337-ornith35-mtp-d1.log`.

The loader defect is reversed without dropping tensors: the real 256-expert affine topology binds and executes. The MTP performance verdict is separately **FAIL for automatic use** on this workload: d1 is slower than AR, only 25% of verified drafts reach d1, adaptive fallback fires, and footprint rises. The bundle has a `prompt_class=code` tuning stamp; a separate generic policy lane is being opened so prompt-scoped measurements cannot authorize global Auto. This PR remains the loader correctness change only.

Tests: new MoE sanitizer/strict-bind/forward suite **5/5 pass**. The isolated neighboring Qwen35 run was **18/19 pass**; its sole failure is the pre-existing Hermes parser provenance assertion (`jangStamped` observed vs `modelTypeHeuristic` expected), outside this two-file diff. No claim is made for full multi-turn/app acceptance from these 64-token rows.

PR #338, opened 2026-08-29
Stream DSV4 affine routed experts from exact SSD regions

jjang-ai:
## Status

**PARTIAL / DRAFT. Do not merge yet.** This PR addresses the source mechanism behind the measured DSV4 footprint runaway, but the exact current head still requires the real-bundle performance, memory, cache, AgentLoop, and Osaurus UI gates below.

## Before reproduction and causal trace

The prior PR-head DSV4 eval completed both children via safe serialization (`[1,1]`, limited by `engineCapacity`) at 13.51 tok/s and 185 ms TTFT, but measured **104,993 MB peak physical footprint**. That is a RAM failure, not a blocked row.

Header-only inspection of `dealign.ai/DeepSeek-V4-Flash-0731-JANG-CRACK` found 43 routed layers, 256 experts per layer, 99,072 affine routed tensors, and **93,952,409,600 bytes (87.5 GiB)** of routed expert tensors across 102 shards. `DeepseekV4Model.sanitize` stacked every per-expert tensor into full layer/projection banks; first routed evaluation could fault/materialize the complete bank.

## Candidate correction

- Configure a model-owned safetensors catalog before generic tensor exclusion.
- Fail closed if any required expert weight/scales/biases descriptor is absent or malformed.
- Exclude only exact `layers.L.ffn.experts.E.{w1,w2,w3}.{weight,scales,biases}` keys.
- Map selected experts by exact safetensors file region, preserving mixed 1/2/3/4/5/6/8-bit affine geometry.
- Preserve DSV4 score-before-down and fp32 limited-SwiGLU ordering.
- Bound the per-layer expert cache; no permanent prestacked overlay and no JangPress.
- Disable compiled regions only for this dynamic expert path; ordinary DSV4 and Qwen paths remain unchanged.

## Current exact-head verification

Head: `c6bd478d`

- Xcode-toolchain `swift build --build-tests`: RC 0.
- Focused affine streaming suite: **9/9**.
- Real-bundle descriptor validation: all **43 x 256** expert layouts accepted; first/last regions mapped.
- DSV4 plus neighboring affine suites: **31/31**.
- Tests include scored-before-down numerical parity, exact exclusion, bounded cache, concurrent region opens, production loader ordering, mixed widths, and fail-closed missing-tensor rejection.

Artifacts:

- `/private/tmp/dsv4-streaming-build-final.log`
- `/private/tmp/dsv4-streaming-focused-final.log`
- `/private/tmp/dsv4-streaming-loader-ordering.log`
- `/private/tmp/dsv4-streaming-real-descriptor-tests3.log`
- `/private/tmp/dsv4-streaming-broad-tests.log`
- `docs/DSV4_AFFINE_EXPERT_FOOTPRINT_2026_08_29.md`

## Mandatory remaining gates

1. Matched base vs branch live bundle: load, TTFT, tok/s, coherent output, stop reason, process survival, current/peak `phys_footprint`, swap/compressor.
2. Reject if footprint remains near full model size or SSD routing makes decode unusably slow.
3. Ten-turn tool loop, Stop mid-stream then continue, restart and disk-backed restore.
4. Prefix/paged/L2 and DSV4 companion-cache counters.
5. Safe-serialized two-child eval `[1,1]` with `engineCapacity`, both children, parent final response.
6. Exact downstream Osaurus Release build driven through the real UI with expected permissions set to Always Allow.

No production/readiness claim is made until those live gates pass on this exact head.

### DIFF
--- a/Libraries/MLXLLM/Models/DeepseekV4.swift
+++ b/Libraries/MLXLLM/Models/DeepseekV4.swift
@@ -1404,9 +1404,11 @@ public class DeepseekV4Model: Module, LLMModel, KVCacheDimensionProvider, LoRAMo
     public func configureSafetensorsModelDirectory(_ modelDirectory: URL) throws {
         try affineExpertCatalog.configure(
             modelDirectory: modelDirectory, layerCount: config.numHiddenLayers)
-        for (layerIndex, layer) in model.layers.enumerated() {
+        var moduleUpdates: [(String, Module)] = []
+        moduleUpdates.reserveCapacity(model.layers.count)
+        for layerIndex in model.layers.indices {
             let limit = config.swigluLimit
-            layer.mlp.switchMLP = AffineStreamingSwitchGLU(
+            let streamingSwitch = AffineStreamingSwitchGLU(
                 inputDims: config.hiddenSize,
                 hiddenDims: config.moeIntermediateSize,
                 numExperts: config.nRoutedExperts,
@@ -1421,7 +1423,19 @@ public class DeepseekV4Model: Module, LLMModel, KVCacheDimensionProvider, LoRAMo
                     DeepseekV4Math.dsv4ScoredSwiGLU(
                         gate: gate, up: up, scores: scores, limit: limit)
                 })
+            moduleUpdates.append((
+                "model.layers.\(layerIndex).mlp.switch_mlp",
+                streamingSwitch
+            ))
         }
+        // `switchMLP` is registered through `@ModuleInfo`; direct assignment
+        // after construction violates MLXNN's module-tree invariants and
+        // process-fatals in `ModuleInfo.set`. Replace the registered children
+        // through the supported tree update API before generic weight loading.
+        try update(
+            modules: ModuleChildren.unflattened(moduleUpdates),
+            verify: .none
+        )
         FileHandle.standardError.write(
             Data(
                 ("[DSV4AffineExperts] production path=exact-mmap-region "
--- a/Libraries/MLXLLM/Models/DeepseekV4.swift
+++ b/Libraries/MLXLLM/Models/DeepseekV4.swift
@@ -130,7 +130,7 @@ class DeepseekV4RoPE: Module {
         if let e = tables[key], e.offset == offset, e.length == length {
             return (e.cos, e.sin)
         }
-        let positions = MLXArray(Int32(offset)..<Int32(offset + length)).asType(.float32)
+        let positions = MLXArray(Int32(offset) ..< Int32(offset + length)).asType(.float32)
         // positions: (L,), invFreq: (dim/2,) → angles: (L, dim/2)
         let angles = positions.expandedDimensions(axis: -1) * invFreq.expandedDimensions(axis: 0)
         let c = cos(angles)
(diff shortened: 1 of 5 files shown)


jjang-ai:
Audit correction pushed after the initial draft: the new validator accidentally included 1-bit affine tensors even though MLX explicitly supports only `{2,3,4,5,6,8}`. Head `34f191e6` removes 1-bit acceptance and updates the real-bundle descriptor assertions to the same exact set.

Current-head focused proof:

- `swift test --filter AffineStreamingSwitchGLUTests`: **9/9 pass**
- real DSV4 descriptor row executes and validates the installed per-expert metadata
- artifact: `/private/tmp/vmlx-pr338-affine-tests-after-bitfix.log`

Status remains **PARTIAL**. This source/test correction does not satisfy the required live DSV4 gates: actual footprint below the prior ~105 GB failure, usable tok/s, TTFT, coherent multi-turn output, tool-result continuation, Stop/continue, and disk/cache restoration are still mandatory before merge.

jjang-ai:
Documentation correction on current head: the proof contract now matches the runtime validator. MLX native affine supports 2/3/4/5/6/8-bit here; 1-bit is explicitly rejected fail-closed. No live footprint or speed claim is added. The 95 GB exact-head runtime matrix remains mandatory.

jjang-ai:
Additional historical baseline (comparison only, not current-head proof): Osaurus community report PR #2494 (`OsaurusAI/DeepSeek-V4-Flash-0731-JANG`, commit `bebc6f3`) recorded 273/316 overall, AgentLoop 115/139 with 0 errors, 17.02 mean tok/s, 422 ms mean TTFT, loop-free 0.9496, and **110,491.79 MB peak phys_footprint**. This independently confirms the full-residency RAM defect on the older path, but it is not a matched A/B for this branch. Current #338 must beat that footprint without making its 17 tok/s baseline unusable, and still satisfy the exact-head live matrix.

jjang-ai:
Current live status on head c09c7f28b9002032417310f80d5c22b469ee1bd0: PARTIAL / NOT MERGEABLE. The first real-bundle attempt exposed a process-fatal lifecycle bug after catalog creation (direct mutation of an @ModuleInfo child). This head fixes that exact crash by replacing all 43 registered switch modules through Module.update(modules:); focused suite 9/9 passes. The same installed DSV4 bundle now loads in 4.02 s, excludes 99,072 routed tensors from generic load, and completes a coherent 3-turn chat. However, performance is unacceptable: turn tok/s = 0.08, 0.75, 0.14 versus the current resident-path eval baseline 13.6766 tok/s. Therefore exact per-expert SSD mapping is rejected as the production strategy even though it avoids full-bank materialization. Artifact: /private/tmp/dsv4-338-live-chat2.log. Next fix must use a resident/reclaimable bounded-bank strategy that preserves usable decode speed; no RAM or production-ready claim is made.

jjang-ai:
Exact-head status update for `81a3dc46`: **FAILED PERFORMANCE GATE / PARTIAL**.

The real DSV4 bundle loads after the MLXNN `update(modules:)` lifecycle correction and completes three coherent turns without the prior process fatal. However, the exact-region strategy measured **0.08 / 0.75 / 0.14 tok/s**, versus **13.6766 tok/s** on the current resident path. It is rejected as a production strategy.

Source trace: `AffineStreamingSwitchGLU.routed` synchronizes route IDs to CPU, dynamically stacks selected projection banks, and forces `MLX.eval` in each of 43 routed layers. A larger mmap cache cannot remove those synchronization boundaries. Bundle inspection also shows each layer/projection spans 3–4 shards at irregular offsets, so it cannot be represented as a normal contiguous affine bank without a copy/overlay.

Next viable boundary: GPU-side shard-span + expert-offset affine dispatch, with a hard proof that MLX/Metal faults only selected pages. No JangPress and no permanent prestacked overlay. The current implementation remains draft and must not merge.

Updated evidence document: `docs/DSV4_AFFINE_EXPERT_FOOTPRINT_2026_08_29.md`.
Live artifact: `/private/tmp/dsv4-338-live-chat2.log`.

PR #378, opened 2026-08-31
Fix growing tool-loop cache boundaries

jjang-ai:
## Summary
- derive the newest reusable history boundary from an exact chat-template continuation proof instead of assuming the current no-generation render stays stable
- use canonical template boundaries for hybrid AR, native MTP, and batched cache stores, retaining the model-load suffix heuristic only for raw inputs without canonical chat metadata
- add trace-only first-divergent-token diagnostics for equal-length cache-key misses
- update the terminal-order source contract: completion metadata may be early, but the stream cannot finish until cache persistence and MLX/advisor drains complete

## Source evidence
The reproduced miss stored and fetched the same token count but diverged at the final boundary token: stored newline token 198 vs next-render `<role>` token 157151. The new continuation LCP excludes that template-rewritten separator without model-ID or token-ID special casing.

## Tests
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcrun swift test --filter CanonicalChatCacheBoundariesTests` — 12/12 passed
- `... swift test --skip-build --filter CacheStoreBudgetTests` — 17/17 passed
- `... swift test --filter BatchEngineGrowingChatCacheSourceTests` — all 27 selected source assertions passed; standalone runner printed a missing default metallib message after the source-only checks
- Osaurus Release app compiled successfully against this checkout

## Live proof status
PARTIAL: the rebuilt Osaurus app is open for the manual growing multi-file/tool-loop run. Do not merge solely from this description until its trace shows the newly stored canonical boundary is restored on successive tool steps and prefill no longer falls back to the static prefix.

### DIFF
--- a/Libraries/MLXLMCommon/Cache/DiskCache.swift
+++ b/Libraries/MLXLMCommon/Cache/DiskCache.swift
@@ -123,6 +123,15 @@ public final class DiskCache: @unchecked Sendable {
     /// an inherited file before it can take the fast path.
     private var validatedFiles: [String: ValidatedFileFingerprint] = [:]
 
+    /// Trace-only identity of the most recent boundary written by this cache
+    /// instance. Growing agent loops can store N tokens and immediately probe N
+    /// tokens under a different hash on the next turn; counts alone hide where
+    /// the prompt stopped being a prefix. Retain the IDs only while explicit
+    /// cache tracing is enabled so the miss log can report the first divergent
+    /// token without changing cache selection or normal-process memory use.
+    private var traceLastStoredTokens: [Int]?
+    private var traceLastStoredHash: String?
+
     /// Thread-safe copy of current disk-cache counters.
     public func snapshotStats() -> DiskCacheStats {
         lock.lock()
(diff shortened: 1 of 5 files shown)


jjang-ai:
Live growing-tool-loop proof on the exact pinned Release build (vMLX a0c5b801050863c79ec6189bd6bf77de4c194bb3, Osaurus dc0a942495b1e7402f0b48b9e8c51eff716dfeb3):

- Eric manually confirmed late tool steps now resume prefill from the expected recent boundary instead of dropping back to an old low prefix.
- Raptor 8B A1B JANG_6M, real multi-step XLSX file_read loop:
  - prompt 29,459 restored boundary 26,232 (3,227 new tokens)
  - prompt 30,254 restored boundary 29,450 (804 new tokens)
  - prompt 30,594 restored boundary 30,245 (349 new tokens)
- 7 tool executions and 9 terminal generation drains were observed.
- Zero cache key-divergence events in the new trace.
- Cache stores completed in about 0.31-0.51s before the associated tool execution became visible to the loop.
- Store-budget telemetry used phys_footprint; the tested rows were allowed under safe-auto. This is not a low-headroom denial proof.

Automated evidence:
- CanonicalChatCacheBoundariesTests: 12/12 passed under the Xcode Swift toolchain, including rewritten assistant separator regression.
- Cache/store budget focused suites: 17 passed.
- Growing-chat source checks: 27 assertions passed; the standalone runner subsequently emitted a missing default metallib diagnostic, so the packaged Release build is the production-path compile/launch evidence.
- Exact Osaurus Release app build completed and launched with OSU_MODELS_DIR=/Users/eric/models.

Scope caveat: vMLX fork workflow jobs are skipped by the upstream-repository guard; Osaurus PR #2568 is the active integration CI surface.

PR #390, opened 2026-09-01
Hybrid cache: stop persisting recurrent-state copies the restore path never applies

jjang-ai:
## Problem

Every stored prompt boundary for a MambaCache hybrid carries the recurrent state **three times**:

1. `mamba_{i}_state0/1` in the v2 payload — authoritative, applied by `deserializeV2`;
2. a folded `ssm_*` copy in the same file — only ever read to re-populate the companion store;
3. the `ssm_companion` sidecar file — fetched for validation, then **not applied**: the apply is gated on `fmtV < 2 || cacheHasArraysState`, and live tracing on Qwen3.8-27B (48× MambaCache + 16× KVCacheSimple) prints `willRestore=false`.

Measured on a fresh cache root (one short chat turn, preview build):
- 64-token boundary: **297MB** (149MB of it dead copies) + 154MB companion
- 3.1k boundary: **488MB** vs 341MB
- save time 0.11–0.12s per record, ~9–13 records per turn ⇒ ~3GB written for one Q&A turn
- the real box's `~/.osaurus/cache/kv_v2` sits at **87GB for 57 entries** in this shape

## Fix

Split the topology signal: the existing `requiresRecurrentSSMCompanionState` (mamba ‖ arrays) keeps its path-dependence semantics untouched (exact-boundary policy, stable-boundary re-derive, N-1 seeds); a new narrow `requiresSeparateRecurrentPayloadState` (**arrays only** — ArraysCache/GDN has no v2 LayerKind, so the companion is its only carrier) gates the storage question. Stores fold `ssm_*`/persist the companion only when the narrow flag is set **or the boundary just published a paged payload** (the tier-1 paged hit path requires `fetchCompleteSSMStates`, so paged/batch setups keep their companion). The fetch gate accepts v2 entries without a companion exactly for the topologies whose restore never applied it.

## Interaction audit

- **Paged tier**: hybrid paged hits reject without a companion → companion still persisted whenever a paged payload was published for the boundary.
- **BatchEngine exact-full-hit N-1 seed**: for mamba topologies the disk seed no longer exists; that path takes its existing correct full-prefill rollback (trade-off documented in the commit; the N-1 KV entry still carries the state in-file for a follow-up).
- **ArraysCache families** (FalconH1, BailingHybrid, …): narrow flag true, behavior byte-identical to before.
- **Old entries**: still readable — folded `ssm_*` reads and companion fetches are unchanged on the read side.
- Unchanged: `shouldPersistExactPromptBoundary`, `shouldForceStableBoundaryRederive`, ZAYA CCA classification.

## Proof

- **Unit**: 44/44 focused coordinator tests pass, incl. 2 new ones — mamba topology stores no `ssm_*`/companion and its v2 entry is accepted and restored without one; separate-payload topology still persists **and** requires the companion.
- **Live** (osaurus preview, fresh cache root, Qwen3.8-27B): companion files **0** (was 1/boundary); entry sizes/save times as above (−30%/−50% bytes, −70% save latency); app restart restores from the slim entry (`cacheRestore completed=2969/2970 detail=disk`), growing-turn boundary hit at `promptMs=10`, TTFT **0.59s** restored vs 4.33s cold; three consecutive restored-state generations coherent at 28–34 tok/s with MTP d1 active.
- **Structural equivalence**: for mamba topologies the restore applies the same `mamba_{i}_*` arrays before and after this change (`willRestore=false` both) — the restored cache state is bit-identical; only never-applied bytes stopped being written.

Also fixes the missing `MLXNN` import in `DeepseekV3MoEDtypeTests` that broke the whole SwiftPM test build.

### DIFF
--- a/Libraries/MLXLMCommon/BatchEngine/BatchEngine.swift
+++ b/Libraries/MLXLMCommon/BatchEngine/BatchEngine.swift
@@ -1810,7 +1810,9 @@ public actor BatchEngine {
                         coordinator.setHybrid(
                             true,
                             requiresRecurrentSSMCompanion:
-                                topology.requiresRecurrentSSMCompanionState)
+                                topology.requiresRecurrentSSMCompanionState,
+                            requiresSeparateRecurrentPayload:
+                                topology.requiresSeparateRecurrentPayloadState)
                         Self.logger.info(
                             "Coordinator flipped to isHybrid=true on first hybrid slot admission"
                         )
--- a/Libraries/MLXLMCommon/Cache/CacheCoordinator.swift
+++ b/Libraries/MLXLMCommon/Cache/CacheCoordinator.swift
@@ -103,6 +103,7 @@ public final class CacheCoordinator: @unchecked Sendable {
     /// hybrid topology that owns its companion tensors inside the v2 layer
     /// payload and therefore sets this false.
     private var _requiresRecurrentSSMCompanion: Bool = false
+    private var _requiresSeparateRecurrentPayload: Bool = false
 
     /// 2026-05-04 (DSV4 SWA/CSA/HSA correctness pass):
     /// Whether the model has hybrid pool caches (DeepseekV4 SWA+CSA+HSA)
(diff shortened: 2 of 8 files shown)


jjang-ai:
Merged-proof record — measurements from tonight's live A/B (osaurus preview @ this branch, fresh cache root, Qwen3.8-27B-JANG_4D, /tmp/osaurus-prefill-debug.log + stderr traces):

**Before** (main, same flow): 64-token boundary 297MB, 3.1k boundary 488MB, one 154MB `ssm_companion` file per boundary, `store-phase save=0.117s/0.121s`; one short turn wrote ~3GB.
**After**: same boundaries 149MB / 341MB, companion files **0**, `save=0.033–0.037s`.
**Restore** on slim entries: app restart → `cacheRestore completed=2969/2970 detail=disk` accepted with no companion on disk; growing-turn boundary hit `promptMs=10`; TTFT 0.59s restored vs 4.33s cold; three consecutive restored-state generations coherent at 28–34 tok/s, MTP d1 active.
**Never-applied proof**: `[vmlx][ssm-restore] types=["MambaCache": 48, "KVCacheSimple": 16] hasArrays=false fmtV=2 willRestore=false` — identical gate outcome pre/post change, so restored cache state is bit-identical; only dead bytes stopped being written.
**Unit**: 44/44 focused coordinator tests incl. 2 new structural tests.

PR #391, opened 2026-09-01
Add an auxiliary cache-prompt intent that never persists boundaries

jjang-ai:
## Problem

Internal utility generations (chat titles, follow-up suggestions, memory distillation, transcript cleanup) embed per-turn content in their prompts — their token streams are never an exact prefix of any future request. Persisting their prompt boundaries is pure write cost. Measured live on a MambaCache hybrid: one osaurus chat turn wrote **six ~150MB boundary records** for its title/suggestion prompts alone — about half the turn's remaining store volume even after #390.

## Change

- New `CachePromptIntent.auxiliary`: documented as restore-allowed, store-forbidden.
- Early-out of store-after-generation in all four engines (solo `TokenIterator`, `BatchEngine.finishSlot`, `NativeMTPTokenIterator`, `DFlash2TokenIterator`). The spec-dec iterators keep their verify rollback/release calls ahead of the guard — those are correctness cleanup, not persistence.
- `LMInput.withCachePromptIntent(_:)` helper mirroring `withCacheRestorePolicy` so hosts can mark a prepared input.

No behavior change for any existing caller — the case must be explicitly requested (osaurus side threads it from `CoreModelService`, whose callers are exactly the four utility services; PR to follow after this lands).

## Proof (live A/B, osaurus preview wired to this branch, fresh cache root, Qwen3.8-27B)

- store-phase records for one chat turn: **12 → 6** — every 144–239-token title/suggestion boundary record eliminated;
- the auto-title ('Why the Sky is Blue') and all four follow-up suggestions still generated normally (`STEP-STATS promptTokens=149 genTokens=5`, `promptTokens=188 genTokens=43`), main reply decoded at 42 tok/s;
- remaining stores are exactly the useful set: warm-up boundaries + main-turn exact/N-1/post-gen.

### DIFF
--- a/Libraries/MLXLMCommon/BatchEngine/BatchEngine.swift
+++ b/Libraries/MLXLMCommon/BatchEngine/BatchEngine.swift
@@ -3213,7 +3213,12 @@ public actor BatchEngine {
         // ring buffer + 5-tuple metaState via `.rotating` LayerKind. The
         // `mediaSalt` is passed through so the stored key matches the key
         // the next fetch will look for (VL multi-turn cache hits).
-        if reason != .cancelled, let coordinator = cacheCoordinator {
+        if reason != .cancelled,
+            // Auxiliary (title/suggestion/summary) prompts never store a
+            // boundary — see `CachePromptIntent.auxiliary`.
+            slot.originalInput.cachePromptIntent != .auxiliary,
+            let coordinator = cacheCoordinator
+        {
             let promptTokens = slot.cachePromptTokenIds
             let hasHybridPool = slot.cache.contains { $0 is HybridPoolCache }
             let promptCacheSnapshot = slot.promptCacheSnapshot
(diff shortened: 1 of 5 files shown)


jjang-ai:
Merged-proof record — live A/B 2026-09-01 03:02 (osaurus preview wired to this branch, fresh cache root, Qwen3.8-27B, /tmp/osaurus-prefill-debug.log + stderr store traces): one chat turn's `store-phase` records dropped **12 → 6** with every 144–239-token title/suggestion boundary record eliminated; auto-title 'Why the Sky is Blue' and all four follow-up suggestions still generated (`promptTokens=149 genTokens=5`, `promptTokens=188 genTokens=43`), main reply 132 tokens at 42 tok/s (screenshot verified in-app). Remaining stores are exactly the useful set: warm-up 41/48, main 3116/3117, post-gen 3007, warm 2969. MLXLMCommon compiles clean; no existing caller passes the new case, so behavior is opt-in only.

Issue #393, opened 2026-09-01
MTPRuntimeFocusedTests: 13 JANG quant-precedence expectations stale since #373, masked by #379's import break


jjang-ai:
`MTPRuntimeFocusedTests` ('MTP runtime metadata') fails 13 quantization-inference expectations on current main (86d23df8) — all of the form `(actual.bits → 8) == 4` / `(actual.groupSize → 64) == 128` in tests like 'JANG hidden anchor overrides stale quantization metadata', 'hidden-input projections override width-colliding metadata', 'declared skip wins across contradictory path aliases', 'lm_head uses uniquely realizable hidden width'.

Timeline: #373 ('JangLoader: a declared quantization that COULD have produced the tensors wins') deliberately inverted the declared-vs-semantic precedence these tests assert, without updating them. Immediately after, #379 shipped `DeepseekV3MoEDtypeTests` without `import MLXNN`, which made the entire SwiftPM test build fail ('cannot find ModuleParameters in scope') — so the stale expectations were unobservable until #390 fixed the import. Verified: at the shipped pin 1315fdcb (with only the import fix applied) the suite records 1 issue; on main it records 14.

Also red on main and predating tonight: 'Both shipped native-MTP models auto-launch at depth 3' (resolvedMTPDraftStrategy → nil for qwen3_5_moe synthetic status) and 'Native MTP manual depth contract' (5 issues). Note the depth-3 suite's premise is now questionable anyway: the 2026-09-01 settled ladder on Qwen3.8-27B measured d1 (31.6 tok/s) > d2 (25.4) > d3 (25.0) on prose — see #383.

Needs someone to reconcile the #373 semantics with the test contracts (update expectations or narrow the declared-wins rule), and to re-point or retire the depth-3 auto-launch expectations.

PR #394, opened 2026-09-01
Native MTP: gate the exact-prompt store behind the canonical hybrid boundary

jjang-ai:
## Problem

Solo TokenIterator, BatchEngine.finishSlot, and DFlash2 all skip the full-prompt boundary record when a path-dependent hybrid has a proven cross-turn boundary (`usesCanonicalHybridBoundary`) — the next turn replaces the generation suffix so the full-prompt key never matches again, and hybrid fetch skips exact boundaries by construction. **The native-MTP iterator was the fourth store path and never got the gate** (same drift its own `CacheStoreBudget` note records for the budget guard).

Measured live on Qwen3.8-27B tool chains: **two ~100ms, ~350–500MB synchronous records per tool call** (exact 3446 beside canonical 3445), where the exact record is never restored — the new `VMLX_STRIP_BOUNDARY_TRACE` diagnostic shows the canonical publish working on every step (`prefixCounts=[1791, …, N-1]`, the assistant-continuation LCP proof layer) and every continuation LCPs at N-1.

## Fix

Mirror the other three engines: compute `usesCanonicalHybridBoundary` and gate the exact-prompt `store` on its absence. Hybrids without a canonical/strip boundary (raw/benchmark prompts) keep the exact record; dense models unaffected; identical reruns resume via the N-1 seed + one-token eval (the same fullhit-trim-eval1 contract fetch already enforces).

Also keeps the env-gated `VMLX_STRIP_BOUNDARY_TRACE=1` diagnostic in `hybridStripBoundaryIndex` that isolated this.

## Proof (live A/B, osaurus preview wired to this branch, 27B, real 7-step file-tool chain with tools auto-approved)

- Before: every tool cycle wrote the pair (e.g. `store-phase count=3446` + `count=3445`, ~0.10s each).
- After: **one record per cycle** — canonical boundaries 2649/2731/2813/2945/3230/3612/4042 across the chain, no exact-N records — while every tool continuation still restored at `promptMs=10` and decode held 20–25 tok/s.
- `CacheCoordinatorTopologyFocusedTests`: 44/44 pass. (`MTPRuntimeFocusedTests` / depth-3 auto-launch suites are red on main independently of this change — stale #373 expectations masked by #379's import break; filed as #393 with pin-vs-main verification.)

### DIFF
--- a/Libraries/MLXLMCommon/Evaluate.swift
+++ b/Libraries/MLXLMCommon/Evaluate.swift
@@ -2262,6 +2262,15 @@ public struct TokenIterator: TokenIteratorProtocol {
         let canonicalBoundary = input.cachePrefixTokenCounts
             .filter { $0 > 0 && $0 < promptTokenIds.count }
             .max()
+        if ProcessInfo.processInfo.environment["VMLX_STRIP_BOUNDARY_TRACE"] == "1" {
+            FileHandle.standardError.write(Data(
+                ("[vmlx][strip-boundary] prompt=\(promptTokenIds.count) "
+                    + "canonical=\(canonicalBoundary.map(String.init) ?? "nil") "
+                    + "prefixCounts=\(input.cachePrefixTokenCounts) "
+                    + "stableCounts=\(input.cacheStablePrefixTokenCounts) "
+                    + "genSuffixTokens=\(coordinator?.genPromptSuffixTokens ?? []) "
+                    + "heuristic=\(heuristicBoundary.map(String.init) ?? "nil")\n").utf8))
+        }
         guard ProcessInfo.processInfo.environment["VMLX_HYBRID_STRIPPED_STORE"] != "0",
             let coordinator,
             coordinator.isHybrid,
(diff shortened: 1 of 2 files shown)


jjang-ai:
Merged-proof record — live A/B 2026-09-01 05:27 (osaurus preview wired to this branch, Qwen3.8-27B, 7-step file-tool chain with `-chatAutoAllowAllTools YES`, /tmp/osaurus-prefill-debug.log + stderr traces): before the gate every tool cycle wrote the exact+canonical pair (`count=3446 save=0.097s` + `count=3445 save=0.105s`); after, exactly one canonical record per cycle (2649/2731/2813/2945/3230/3612/4042) with zero exact-N records, every tool continuation restoring at `promptMs=10` (lcpVsPrev = N-1 on each step, canonical `prefixCounts=[1791,…,N-1]` published throughout), decode 20–25 tok/s across the chain. 44/44 focused coordinator tests pass; pre-existing main test reds documented in #393 with pin-vs-main verification.

PR #396, opened 2026-09-01
Tool dispatch ends the turn: early stop instead of drain-to-EOS

jjang-ai:
## Problem

When a host dispatches a parsed tool call, everything decoded afterward is discarded — the public stream already terminated at the call and post-tool prose is suppressed by design. But consumer termination mapped to `.cancelled` (which skips boundary stores on the batched path), so hosts protected the store by **draining to EOS**: up to `maxTokens` of dead decode at full GPU cost while the executed tool's continuation waited on the generation lease.

## Change

- **`generateLoopTask`** (solo, native-MTP, and DFlash2 engines all run through it): consumer termination / task cancellation after the handler emitted a tool call resolves to **`.stop`** — the dispatched tool defines the turn's end. Non-tool Stop-button semantics unchanged (`.cancelled`, per the #2388 contract).
- **`BatchEngine.finishEarly(_:)`** — the natural-stop sibling of `cancel(_:)` for active slots: `finishSlot(reason: .stop)`, boundary stores run.
- **Batched bridge**: records surfacing a `.toolCall` (`markToolCallEmitted`) and routes consumer termination to `finishEarly` instead of `cancel` from that point on.

Opt-in by construction: a host that keeps draining behaves exactly as before; a host that stops consuming at dispatch gets the early stop with stores intact.

## Proof (live, Qwen3.8-27B, force_on d3, identical multi-call prompt, quiet machine)

| leg | dispatch-step genTokens | decode time |
|-----|------------------------|-------------|
| before (drain-to-EOS) | 5,383 | 263.7s |
| after (early stop) | **3,492** | **153.9s** |

~110 seconds of post-dispatch decode eliminated on one step. Two earlier contended-machine samples show the same dominance (3,764 vs 2,674). Run-to-run token variance is thinking-length (adaptive AR-fallback samples at temperature); every after-sample is below every before-sample. After the early stop: `==GEN STREAM-DRAINED` → `LEASE-RELEASED` 32ms apart, full `STEP-MTP` summary finalized, boundary stored, and the tool continuation restored at `promptMs=10`. The fixed build's multi-step agent loop ran normally on non-dispatch turns (natural stops, 10ms continuations).

### DIFF
--- a/Libraries/MLXLMCommon/BatchEngine/BatchEngine.swift
+++ b/Libraries/MLXLMCommon/BatchEngine/BatchEngine.swift
@@ -154,10 +154,25 @@ private func debugDumpReasoningPrompt(
 private final class BatchStreamTerminationState: @unchecked Sendable {
     private let lock = NSLock()
     private var completed = false
+    private var toolCallEmitted = false
 
     func markCompleted() {
         lock.lock()
         completed = true
+        lock.unlock()
+    }
+
+    /// Record that the bridge surfaced a parsed `.toolCall` to the consumer.
+    /// A consumer that terminates AFTER this point has everything it needs —
+    /// the dispatched tool defines the turn's end — so termination routes to
+    /// ``BatchEngine/finishEarly(_:)`` (natural `.stop`, boundary stores run)
+    /// instead of ``BatchEngine/cancel(_:)`` (stores skipped). Without this,
+    /// a host that stops consuming at tool dispatch either loses the
+    /// boundary store or must drain the model's post-tool-call prose to EOS
+    /// at full decode cost (measured up to maxTokens=16384 of zombie decode).
+    func markToolCallEmitted() {
+        lock.lock()
+        toolCallEmitted = true
         lock.unlock()
     }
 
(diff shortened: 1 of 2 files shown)


jjang-ai:
Merged-proof record — quiet-machine A/B 2026-09-01 11:41–11:48 (Qwen3.8-27B force_on d3, identical multi-call prompt, osaurus preview wired to this branch vs merged main): dispatch step BEFORE 5,383 tokens / 263.7s decode → AFTER **3,492 tokens / 153.9s** (~110s of post-dispatch decode eliminated). Two earlier contended samples dominate identically (3,764 → 2,674). After the early stop: `==GEN STREAM-DRAINED` → `LEASE-RELEASED` 32ms apart, full STEP-MTP summary finalized (verifyCalls=1072, downshifts=35), boundary stored, tool continuation restored at `promptMs=10`. Multi-step agent-loop turns without a pending dispatch completed with natural stops on the fixed build. MLXLMCommon compiles clean; behavior is opt-in (a host that keeps draining is unchanged).

PR #401, opened 2026-09-01
Decode perf: GDN grouped input-projection fusion + MLA bf16 decode SDPA

jjang-ai:
## What

The four GDN input projections (qkv/z/b/a) read the same decode hidden state but ran as four separate quantized matmuls. The existing all-or-nothing fusion only engaged when every projection shared one quantization scheme, so mixed stamps — like the 27B's qkv/z at 5-bit gs128 and b/a at 4-bit gs128 — never fused.

This groups the projections by (groupSize, bits, mode, scales/biases dtype) and fuses each compatible group into one `quantizedMM` per group (the 27B fuses 4→2). The uniform whole-4 path in the VLM class keeps its existing bf16-affine kernel and fused record, so compiled decode and Flash Next behavior are byte-for-byte untouched; grouped fusion is the fallback for stamps the uniform path refuses. Decode-only (S==1), disabled inside compiled-decode tracing, opt-out via `VMLX_GDN_FUSE_DECODE_INPUTS=0`.

## Proof

**Bit identity (unit, gates the kernel):** `Qwen35FusedInputProjectionTests` 2/2 pass — uniform stamp (4/4/4/4, fuses 4→1) and the 27B's real mixed shape (8/8/4/4, fuses 4→2). The suite caught a real divergence during development: fusing through the bf16-affine dense kernel was NOT bit-identical to the separate `QuantizedLinear` calls; the shipped fused op is the exact `quantizedMM` those calls dispatch, so outputs match bit-for-bit.

**Live A/B (visual harness, quiet machine):** Qwen3.8-27B-JANG_4D, agent chat (4,175-token prompt), MTP off (AR decode path), 3 valid reps per arm, fresh app instance per rep:

| arm | decode tok/s (reps) | median |
|---|---|---|
| fused (default) | 23.2 / 23.3 / 23.5 | **23.3** |
| `VMLX_GDN_FUSE_DECODE_INPUTS=0` | 22.5 / 22.8 / 23.0 | **22.8** |

+2.2% sustained decode with clean separation (every fused rep beats every unfused rep). Fusion activation confirmed per-rep via the `fused_gdn_decode_input_projections=active groups=[2, 2]` stderr diagnostic in the fused arm only.

### DIFF
--- a/Libraries/MLXLLM/Models/Qwen35.swift
+++ b/Libraries/MLXLLM/Models/Qwen35.swift
@@ -215,6 +215,158 @@ final class Qwen35GatedDeltaNet: Module {
     private var qScaleWeight: MLXArray?
     private var kScaleWeight: MLXArray?
 
+    // MARK: fused decode input projections
+    //
+    // All four input projections read the SAME hidden state, so decode pays
+    // four quantized-matmul kernel launches per GDN layer per token where
+    // scheme-compatible projections could share one. Port of the proven
+    // Qwen4Exp/VLM `fusedDecodeInputs` (MLXVLM/Qwen35.swift), generalized to
+    // GROUPED fusion: the VLM guard is all-or-nothing on quantization scheme,
+    // but real JANG bundles mix schemes — the 27B stamps qkv+z at 5-bit and
+    // b+a at 4-bit (group size 128 throughout), which the all-or-nothing
+    // check refuses entirely. Partitioning the ordered list [qkv, z, b, a]
+    // into runs of identical (groupSize, bits, mode, dtypes) fuses what CAN
+    // fuse (4→2 on the 27B, 4→1 on uniform stamps) and leaves the rest on
+    // their original modules. Row-wise concatenation of affine-quantized
+    // weights/scales/biases preserves each output row's dequantization
+    // exactly, so the fused matmul is bit-identical to the separate calls.
+    //
+    // Decode-only (S == 1): prefill is compute-bound where launch overhead is
+    // noise, and the compiled decode trace must not capture the lazily built
+    // fused arrays. `VMLX_GDN_FUSE_DECODE_INPUTS=0` disables for diagnosis.
+    private enum FusedInputSegment {
+        case fused(
+            members: [Int],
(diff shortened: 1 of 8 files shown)


jjang-ai:
CI triage: the `lint` and four `Linux Container CMake Swift Build` failures are pre-existing and repo-wide — the same five lanes fail identically on every currently-open PR regardless of content (checked `fix/drain-test-current-contract` run 33555467659 and `feat/vmlx-env-prefix` run 33550320878). The lint lane builds the latest swift-format tag and reformats the whole tree; against the current tag the existing tree produces 600+ file diffs, so the failure is formatting drift on main, not this branch's changes. The meaningful gate for this PR is `mac_build_and_test` (queued behind the Metal runner).

jjang-ai:
Second commit added: **MLA decode bf16 SDPA** (3093a1fa) — removes the S==1 fp32 cast in `mlaScaledDotProductAttention` that copied the full KV cache to fp32 every decode token for every MLA family (DeepseekV3/V4, BailingHybrid, BailingMoeV3).

**Root cause of the field report** "95 tok/s short output → 20 tok/s @ 2k in the news agent" on Raptor 8B-A1B (32-core M5 Max): measured here (40-core, RunBench, MTP off, 3 runs/cell, median) the old path loses ~0.95µs per context-token per decode step; controls: qwen3_5 27B (GDN) and Nanbeige (dense) are context-flat in the same rig, Ling-3.0-tiny (same MLA family) collapses identically.

| ctx | fp32 cast (old) | bf16 (this PR) |
|---|---|---|
| 370 | 109 | 98 |
| 3k | 94 | 122 |
| 12.8k | 52 | **101 (1.9x)** |
| 25.9k | 30 | **73 (2.4x)** |

Residual slope after the fix is 0.24µs/ctx-token — actual KV read physics. Precision gate `MLADecodeSDPADtypeTests` 2/2: bf16-vs-fp32 closeness at real MLA geometry (192/128, 4096 ctx) and a saturated-softmax bound (~10 bf16 ulps; prefill has always run these families bf16, decode joins existing behavior). `VMLX_MLA_SDPA_FP32=1` restores the old cast.

PR #404, opened 2026-09-02
Qwen3VL: reorder vision rows to placeholder order before the pooled scatter

jjang-ai:
## What

Qwen3VL (qwen3_vl / Qwen3-VL family) carries the same media-order defect that was fixed for the qwen3_5 family in #298: feature rows arrive image-batch-first (`prepare` concatenates image pixels before video pixels) while placeholders sit in conversation order, and the pooled scatter over `imageMask .|| videoMask` assumes those orders agree. A video turn followed by an image turn silently swaps the two blocks (total element count still matches). Qwen3VL additionally re-applies the SAME rows once per deepstack layer (`applyDeepstack` adds them at `visualMask` positions), so a per-kind scatter split would not be enough.

Fix: compute one permutation mapping batch order → placeholder order and apply it to the merged features AND every deepstack layer's rows. Single-kind inputs return nil and keep the existing fast path untouched; per-kind count mismatches still throw.

## Proof

`Qwen3VLFeatureOrderTests` 3/3 on this branch (post-rebase onto main): video-before-image pulls video rows first, image-first is the identity, interleaved kinds consume each queue in placeholder order. MLXVLM builds clean.

## Note

Originally opened carrying the full #298 chain by mistake (#298 already merged Aug 22 and shipped in 0.24.3); rebased to exactly this one commit.

### DIFF
--- a/Libraries/MLXVLM/Models/Qwen3VL.swift
+++ b/Libraries/MLXVLM/Models/Qwen3VL.swift
@@ -1754,6 +1754,82 @@ public final class Qwen3VL: Module, VLMModel, KVCacheDimensionProvider {
         return indices
     }
 
+    /// Permutation that reorders vision feature rows from batch order
+    /// (image rows first, then video rows — the order `prepare`
+    /// concatenates pixels in) to PLACEHOLDER order (conversation order).
+    ///
+    /// The pooled scatter over `imageMask .|| videoMask` assumes those two
+    /// orders agree, and they do not: a chat whose earlier turn carried a
+    /// video and whose current turn carries an image has video placeholders
+    /// FIRST, so the pooled scatter lays the image's rows onto the video's
+    /// pads and vice versa — the blocks swap, silently, because the total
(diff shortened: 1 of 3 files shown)


jjang-ai:
Extended to Qwen3VL (`458cb074`): same combined-mask scatter defect, plus Qwen3VL re-applies the mis-ordered rows once per deepstack layer. Fixed with a single batch-order→placeholder-order permutation applied to the merged features and every deepstack layer's rows; single-kind inputs keep the existing fast path untouched. `Qwen3VLFeatureOrderTests` 3/3 (video-first pulls video rows, image-first is identity, interleaved kinds consume queues in placeholder order).

Modality audit alongside: Gemma4's audio and per-image scatters are per-kind masks (correct by construction); Gemma4 has no combined-mask feature scatter. Provenance: the defect predates the May-14 repo consolidation in both qwen files — no recent PR introduced it; the new agent orchestration made multi-turn mixed-media reachable.

PR #437, opened 2026-09-06
Load: Ling 2.6 JANGTQ materialises its affine weights to bf16 under mmap — the osaurus#2652 '!!!!' root cause

jjang-ai:
Root cause of osaurus-ai/osaurus#2652 ("!!!!" answers from Ling 2.6 flash JANGTQ since 0.24.5), reproduced in the app and isolated at the engine level with a single-variable control.

## What

`requiresJANGTQMmapBFloat16` gains `bailing_hybrid`: under the mmap load path (the one the app uses for large bundles) the JANGTQ-native Ling 2.6 flash bundle's non-TurboQuant affine weights (8-bit attention/embedding with JANG's f16 scales) are now materialised to bf16, exactly as nemotron_h and qwen3_5_moe linear-attention bundles already are. TurboQuant tensors (`tq_packed`/`tq_norms`) stay raw.

## Why

Under mmap the loader keeps every file-backed tensor as stored, so the 8-bit embedding with f16 scales seeds an **f16** residual stream. Since #407 (in 0.24.4+) the GLA layer casts its fp32 recurrence output back to the activation dtype; in f16 that output overflows (the exact overflow the 2026-05 fp32 promotion in `recurrentGLAReference` documents) → NaN → every logit NaN → token 0 `!`. Before #407 the fp32 GLA output promoted the whole stream and hid the f16 seed — which is why the reporter's bundle answered on 0.24.3 and streamed `!` from the first #407 pin on. The non-mmap engine path materialises to bf16 and never showed it, which is why the earlier probes were clean.

## Tests

`LoadWeightShardSelectionTests.testJANGTQMmapBFloat16FamiliesIncludeBailingHybrid` (bailing_hybrid true incl. case; nemotron_h; qwen3_5_moe with/without linear_attention; llama/empty false). `Ling26GLAProbe` (env-gated) is the measurement below.

# PROOF:
Same bundle (`/Volumes/EricMLWork/models/JANGQ-AI/Ling-2.6-flash-JANGTQ`, Hub rev 68e3bda2), same 542-token prompt, Xcode toolchain, `Ling26GLAProbe`:

| arm | prefill non-finite logits | token-0 fraction (greedy / sampled) | text |
|---|---|---|---|
| `MLX_SAFETENSORS_MMAP=1` (app path), before | 157184/157184 | 1.0 / 1.0 | `!!!!…` |
| `MLX_SAFETENSORS_MMAP=1 VMLX_JANGTQ_BF16_MMAP=1` (control: only the materialisation differs) | 0/157184 | 0.0 / 0.0 | coherent, top-5 identical to the non-mmap run |
| `MLX_SAFETENSORS_MMAP=1`, this branch (no override) | 0/157184 | 0.0 / 0.0 | coherent; disk-restore + iterator-restore coherent, 0 token-0 |

In-app reproduction (before): combined osaurus dev build (#2647 + repin eeccf887, linker-signed), harness OBSIDIAN leg, Ling 2.6 flash JANGTQ: every assistant row 1073 tokens of `!`, run log `[vmlx][nan-logits] site=solo step=0 model=BailingHybridModel nan=157184` on every generation including 143-token fresh-prefill title generations (not the cache). The in-app after-run is taken with the repin to this fix (next step in the same lane; recorded on the repin PR). CI baseline: lint/Linux identical to merged #424; no mac runner registered.

### DIFF
--- a/Libraries/MLXLMCommon/Load.swift
+++ b/Libraries/MLXLMCommon/Load.swift
@@ -1447,7 +1447,16 @@ private func isJANGTQParameterKey(_ key: String) -> Bool {
     key.hasSuffix(".tq_packed") || key.hasSuffix(".tq_norms")
 }
 
-private func requiresJANGTQMmapBFloat16(_ modelDirectory: URL) -> Bool {
+/// JANGTQ-native bundles whose NON-TurboQuant affine weights must be
+/// materialised to bf16 even under mmap. Under mmap the loader otherwise keeps
+/// every file-backed tensor as stored, and JANG stamps f16 affine scales: the
+/// 8-bit embedding then seeds an f16 residual stream and every f16-range
+/// overflow in the runtime becomes NaN. Ling 2.6 flash JANGTQ (`bailing_hybrid`,
+/// GLA linear attention, osaurus#2652) is the measured case: with the mmap
+/// path every logit was NaN (157184/157184, token 0 "!" streams) and with the
+/// bf16 materialisation alone the same load answered coherently. The
+/// TurboQuant tensors themselves (`tq_packed`/`tq_norms`) stay raw either way.
+func requiresJANGTQMmapBFloat16(_ modelDirectory: URL) -> Bool {
     let configURL = modelDirectory.appendingPathComponent("config.json")
     guard
         let data = try? Data(contentsOf: configURL),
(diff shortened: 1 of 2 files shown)


jjang-ai:
Correction to the PROOF table: the row labelled `MLX_SAFETENSORS_MMAP=1 VMLX_JANGTQ_BF16_MMAP=1` was run through `env $arm` in zsh, which does not word-split — it set a single variable `MLX_SAFETENSORS_MMAP="1 VMLX_JANGTQ_BF16_MMAP=1"`, so that arm was in fact a NON-mmap load, not the override control. The valid before/after is the pair that share one environment: `MLX_SAFETENSORS_MMAP=1` on main before this PR → 157184/157184 NaN, token-0 fraction 1.0; `MLX_SAFETENSORS_MMAP=1` on this branch → 0 NaN, coherent (single variable = the code change). The always-on loader line from #438 now prints the decision (`bf16=true mmap=true … params[bfloat16=651 float16=93 uint32=295]`). The in-app after-run is still open: the first integrated build (engine 16379811) still produced NaN logits in the app, which is being traced with that line.

PR #450, opened 2026-09-07
SSM replay: own each boundary snapshot; throwing replay contract so a failed forward publishes nothing (fixes the GLM-5.3 post-generation crash)

jjang-ai:
## Scope (narrow)
Two defects in the SSD-cache replay path, plus the crash that exposed them. **Not** a GLM-5.3 qualification and **not** a change to generation behaviour.

1. **Crash.** `Glm5Next` never implemented the raw token forward; the cache store's boundary replay called it through `LanguageModel`'s `fatalError` default, so every GLM-5.3 generation persisted its turn and then killed the process (`LanguageModel.swift:511`, osaurus 2026-09-07).
2. **Boundary aliasing (pre-existing, every KDA/Mamba hybrid).** `extractSSMStates` returns the live cache arrays and the KDA layer writes its next state through the `ArraysCache` subscript, which updates them in place. `reDeriveSSMStatesAtBoundaries` kept the list for an earlier boundary while replaying on, so the store published boundary-8 state under the boundary-3 key, in memory and on disk. Each boundary's snapshot is now materialised before the replay continues.
3. **Failed replays published state.** A thrown forward became zero logits inside the replay and whatever the cache held was stored as legitimate. `LanguageModel.replayForward(_:cache:) throws` (default = the generation contract, exact for models whose forward cannot fail; Glm5Next rethrows) is used at all three replay sites; the store's existing catch then publishes nothing.

## Evidence (tiny 64-hidden fixture, geometry refused before construction, 2M-parameter ceiling, bounded supervisor with descendant accounting)
- **Failing-before / passing-after**, isolated: with only the ownership-copy hunk reverted, `earlierBoundaryIsNotAdvancedByLaterReplay` fails (`identical → 6` of 6 slots: boundary 3 == boundary 8) and boundary-3 parity is off in every element; restored (diff hash verified), both pass. Runs 170351 (before) / 165800, 172353 (after).
- Parity: re-derived states equal fresh-cache passes with the replay's own segmentation; deterministic; disk tier serves what a successful replay stored, equal to the published lists; boundary 3 from disk is the boundary-3 state.
- Failed replay (induced: `index_kpool_compress:false`) and late failure (after a successful boundary): nothing published, memory or disk; `reDerives == 0`.
- 16/16 with `MLX_ENABLE_TF32=0`; 16/16 with the MLX default, where 11 chunk-segmentation comparisons run as attributed known issues (see below). Receipt test records the effective setting each run.

## Separately tracked, NOT fixed here
- **TF32.** This MLX build defaults `MLX_ENABLE_TF32=1`; on NAX hardware float32 GEMM (M ≥ 2) runs at TF32 precision while GEMV (M = 1) and CPU are fp32 (matmul.cpp:367, utils.h:169). A lone-token chunk and a multi-token chunk of the same tokens therefore diverge (~1e-3); measured on plain `matmul`, confirmed gone with TF32 off. The diagnostics stay in the suite as known issues; no tolerance was widened. Whether vmlx should disable TF32 for fp32 paths is a policy decision.
- **Generation-error substitution — UNRESOLVED.** Glm5Next's generation overloads still return zero logits (with a stderr line) when the forward throws. Exactly: those substituted logits **can be sampled** (a uniform distribution → a random token is streamed and decoding continues) and the live cache after such a step **can be captured** by `captureCleanSSMStateInline` and persisted, because nothing on the generation path signals the failure. The KV disk tier refuses only NaN/Inf. This PR makes the replay path safe; it does not make the generation path safe, and the GLM generation/cache path must not be described as corrected.
- Cross-instance disk restore of replay-only companion rows: a second coordinator on the same directory found no rows and the files were gone afterward (companion rows are linked to KV rows the engine writes after generation) — observed, not asserted, not investigated.

## Merge gate
The `Build and Test` workflow runs on this PR, but `mac_build_and_test` has been queued since 2026-09-08T00:24Z with 0 self-hosted runners registered (`GET /actions/runners` → 0; the 2026-09-07 11:55 runs are still queued), and the lint/Linux jobs fail on every PR of this repository. An exact-head green cannot be produced here; the gate is the supervised local evidence above. Exact tested commit: 25d1f343.

### DIFF
--- a/Libraries/MLXLMCommon/Cache/SSMReDerive.swift
+++ b/Libraries/MLXLMCommon/Cache/SSMReDerive.swift
@@ -307,7 +307,8 @@ public func reDeriveSSMStates(
             } else {
                 tailInput = tail.tokens.reshaped([1, tail.tokens.size])
             }
-            _ = model.callAsFunction(tailInput, cache: freshCache)
+            _ = model.callAsFunction(
+                LMInput.Text(tokens: tailInput), cache: freshCache, state: nil)
         }
     case .logits:
         break
@@ -376,11 +377,19 @@ public func reDeriveSSMStatesAtBoundaries(
                     let tailInput = tail.tokens.ndim >= 2
                         ? tail.tokens
                         : tail.tokens.reshaped([1, tail.tokens.size])
-                    _ = model.callAsFunction(tailInput, cache: freshCache)
+                    _ = model.callAsFunction(
+                        LMInput.Text(tokens: tailInput), cache: freshCache, state: nil)
                 }
                 needsFreshReplayPreparation = false
             } else {
-                _ = model.callAsFunction(tokenArray, cache: freshCache)
+                // Generation's own contract (`LMInput.Text`), not the raw token
+                // overload: vision-language models implement the former and
+                // inherit a trapping default for the latter, so replaying a
+                // hybrid VLM's prompt here used to crash the process at the
+                // end of every generation (GLM-5.3, 2026-09-07). Plain LLMs
+                // reach their token overload through the protocol default.
+                _ = model.callAsFunction(
+                    LMInput.Text(tokens: tokenArray), cache: freshCache, state: nil)
             }
             MLX.eval(freshCache)
             cursor = end
--- a/Libraries/MLXLMCommon/Cache/SSMReDerive.swift
+++ b/Libraries/MLXLMCommon/Cache/SSMReDerive.swift
@@ -168,9 +168,9 @@ public func maybeReDeriveSSMState(
 ///     contain the SSM state at the boundary because Mamba caches are
 ///     state-replacing (O(1) per token), not appending — the live
 ///     cache IS the snapshot.
-///   • We deep-copy via `extractSSMStates(from:)` (which calls
-///     `state.value.copy()` per CacheHelpers.swift) so subsequent
-///     mutation by generation-prompt tokens doesn't alias.
+///   • `extractSSMStates(from:)` returns the LIVE cache arrays (it does not
+///     copy); the copy that protects this snapshot from later in-place
+///     writes is the `* 1` materialisation in `SSMStateCache.store`.
 ///   • Storage matches `maybeReDeriveSSMState`: keyed on the stripped
 ///     prefix hash + modelKey so multi-turn fetch hits exactly.
 ///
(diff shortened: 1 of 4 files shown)


jjang-ai:
## PROOF (exact commit 25d1f343; supervised runs, tiny fixture only, no model)
| run | setting | result |
|---|---|---|
| 170351 | ownership copy reverted (revert/restore patches sha bf545524/192af801, diff hash verified after) | `earlierBoundaryIsNotAdvancedByLaterReplay` FAILS: `identical → 6` of 6; boundary-3 parity off 576/576 conv, ~1020/1024 recurrent |
| 172353 | MLX_ENABLE_TF32=0 | 16/16 pass, strict |
| 172407 | MLX default (TF32) | 16/16 pass, 11 attributed known issues (segmentation/GEMM), receipt maxAbs 1.98e-3 |
| 171552 / 171649 | TF32=0 receipts | GEMM-vs-CPU maxAbs 7.2e-7 |

Supervisor: peak tracked footprint ≤ 7.12 GB (compiler jobs, seen via descendant accounting: 29 pids, 31 outside the launch group), swap flat 1.21 GB, pressure normal, cleanup verified 0/0/0 every run. Logs: `~/vmlx-private-evidence/mtp-swift-2026-09-04/logs/SWIFTTEST_Glm5NextTokenForwardTests_*` and ISSUES.md 05:30 → 12:10.

Not covered: live GLM run (frozen), generation-error substitution (unresolved, stated in the body), cross-instance companion restore (observed, untested).

jjang-ai:
Reviewed exact head 22b807cc70c7579102370806e77ad13cccefa7ab. Runtime diff remains the scoped replay/snapshot change; review added a test-only portability correction: permitting TF32 does not require every Metal device to exhibit NAX rounding. No runtime precision policy changed. Clean pinned submodules 73312d3e / 7bede8f1, freshly compiled Metal library. Supervised local run SWIFTTEST_PR450PortableReview3__232525: default policy 16/16 with 11 attributed known issues; MLX_ENABLE_TF32=0 16/16 strict, GEMM-vs-CPU maxAbs 7.15e-7. Swap flat 1.64GB; cleanup group/tracked/watchdog 0/0/0. Earlier two isolated test-package attempts failed on a missing helper and metallib placement, not a runtime regression; retained in local logs. Evidence: /Users/eric/vmlx-private-evidence/mtp-swift-2026-09-04/logs/SWIFTTEST_PR450PortableReview3__232525.log and matching telemetry. Maintainer explicitly waived engine CI. No app repin or live GLM run; generation zero-logit substitution and cross-instance companion restore remain unqualified. Osaurus consumption and visual qualification are separate gates.

PR #455, opened 2026-09-07
fix(mlxllm): enable Qwen3.5 text routed-MoE decode

### LINKED ISSUE #453: Qwen3.5 LLM path leaves the trusted routed-MoE decode region disabled
## Problem

The Qwen3.5 text-only model path does not enable the trusted routed-MoE decode region that already exists in `SwitchGLU`.

`Libraries/MLXLMCommon/SwitchLayers.swift` contains `Qwen4ExpCompiledRoutedSwitchGLU`. Its call site is guarded by `SwitchGLU.compileSeparatedDecode` and otherwise falls back to the existing eager routed path.

The VLM Qwen3.5 implementation passes its architecture-scoped `compileDecodeRegions` policy when constructing `SwitchGLU`, but `Libraries/MLXLLM/Models/Qwen35.swift` constructs the same module without enabling that policy. As a result, eligible text-only Qwen3.5 MoE bundles cannot reach the existing compiled routed-MoE region.

## Expected behavior

The text-only Qwen3.5 path should use the same architecture- and environment-scoped policy as the VLM path, so eligible routed-MoE decode calls can use the existing trusted region while unsupported shapes and disabled configurations continue to use the current eager implementation.

## Proposed scope

This should be a small wiring fix, not a new kernel or a change to the generic `SwitchGLU` fallback. Reuse or factor the existing Qwen3.5 compile-region predicate rather than enabling compilation unconditionally for unrelated model shapes.

## Acceptance criteria

- Eligible Qwen3.5 text-only MoE bundles reach `Qwen4ExpCompiledRoutedSwitchGLU` during single-token decode.
- The explicit compile-region opt-out remains honored.
- Unsupported shapes, dtypes, quantization layouts, and prefill calls retain the existing eager path.
- Focused eager-versus-compiled output/parity coverage is added.
- A small decode measurement demonstrates the path is engaged without claiming an end-to-end speedup that was not measured.

tijs:
## Proposed changes

Fixes #453

- Factor the Qwen3.5 compiled routed-MoE decode predicate into `MLXLMCommon`.
- Wire the text-only Qwen3.5 decoder to the same architecture- and environment-scoped policy already used by the VLM path.
- Preserve `SwitchGLU`'s existing eager fallback and exact-shape, dtype, quantization, and single-token guards.
- Keep MTP layers on their existing eager default.

## Verification

- `swift test --filter Qwen35CompiledDecodePolicyTests --jobs 4` — 4/4 passed.
- `swift test --filter Qwen35VLMGatedDeltaTests/compiledDecodePolicyIsArchitectureScoped --jobs 4` — 1/1 passed.
- `swift test --filter Qwen4ExpConfigurationTests --jobs 4` — 4/4 passed.
- `swift-format lint` passes for the new policy and focused test files.
- The CMake pre-commit hook passes; the repository-wide in-place Swift formatter was not run because it rewrites unrelated existing files.

No checkpoint fixture was available for a runtime eager-versus-compiled parity run or decode measurement, so this PR adds focused policy and construction coverage rather than claiming runtime parity or a speedup.

## Checklist

### DIFF
--- a/Libraries/MLXLLM/Models/Qwen35.swift
+++ b/Libraries/MLXLLM/Models/Qwen35.swift
@@ -853,6 +853,12 @@ final class Qwen35SparseMoeBlock: Module, UnaryLayer {
     let normTopkProb: Bool
     let numExperts: Int
     let topK: Int
+    /// Whether exact-shape single-token decode may reach the trusted compiled
+    /// routed-MoE region in `SwitchGLU` (see `Qwen35CompiledDecodePolicy`).
+    /// Mirrors the VLM construction: the main decoder stack consults the
+    /// shared architecture/environment policy; MTP layers keep the eager
+    /// default, matching the VLM's MTP path.
+    let compileDecodeRegions: Bool
 
     @ModuleInfo(key: "gate") var gate: Linear
     @ModuleInfo(key: "switch_mlp") var switchMLP: SwitchGLU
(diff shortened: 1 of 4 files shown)


jjang-ai:
## Exact-head review and bounded dispatch evidence

Reviewed a16ecc85b8cdd0e2d32e28e4a2094eaa119e3020. The shared default predicate preserves all previous VLM conditions; text construction now forwards it, while MTP keeps the eager default. No overlap with intervening main changes. Engine CI is waived for this merge.

Supervised local run PR455QwenRouteReview__234442:5/5 selected checks, exit0. The three policy tests cover eligibility, neighboring shapes and current/legacy explicit overrides. A separate61760-parameter text-construction fixture covers ON/OFF wiring and the unchanged MTP default; the production-size constructor test was not executed.

A tiny64-wide, four-expert affine4-bit BF16 fixture actually entered the compiled routed region (runtime active trace). Compiled/eager outputs were identical; both relativeL2errors versus a float32-input reference were0.011081707, within the existing0.02 numerical-test bound. Prefill and float32 fallback comparisons were exact. The compiled region has its own dtype/quantization/row guards; architecture eligibility comes from the shared policy.

Peak tracked footprint0.73GB, swap unchanged1.60GB, cleanup0/0/0. No checkpoint load, speed/coherence claim, native-MTP qualification or app repin is implied. No release included.

PR #456, opened 2026-09-07
fix(cache): reuse stripped boundary for standalone rotating caches

### LINKED ISSUE #454: Standalone rotating caches miss the growing-chat stripped boundary
## Problem

Standalone rotating/sliding-window cache topologies do not persist the generation-suffix-stripped boundary needed for growing multi-turn chat reuse.

The cache coordinator already has a `cacheHasStandaloneRotatingWindowState` predicate for identifying standalone rotating caches. However, `TokenIterator.hybridStripBoundaryIndex` currently only considers `coordinator.isHybrid`, and the corresponding stripped-boundary store in `BatchEngine.finishSlot` has the same hybrid-only gate.

A growing chat prompt replaces the template's final generation suffix with the previous assistant response and the next user turn. Therefore the full prompt boundary and post-answer keys do not match the next prompt as a prefix. For a standalone rotating cache, the generation-suffix-stripped boundary is the reusable boundary; without it, each growing turn can fall back to a full prefill even though the cache topology is safe to persist at that boundary.

## Expected behavior

For a cache proven by `cacheHasStandaloneRotatingWindowState`, the generation-suffix-stripped boundary should be selected and stored in the same way as the existing hybrid boundary. Dense paged caches and path-dependent hybrid or unsupported cache types should retain their current safety gates.

## Proposed scope

Keep this limited to the existing boundary plumbing:

- pass the cache topology to the shared boundary-selection helper;
- allow the standalone rotating-cache predicate there;
- apply the same condition to the `BatchEngine` stripped-boundary store;
- update the alternate iterator call sites so all generation paths use the same decision.

No new cache format or model-specific behavior is required.

## Acceptance criteria

- A growing-chat run using a standalone rotating/sliding-window cache stores a generation-suffix-stripped boundary.
- The following turn restores that boundary and prefills only the unmatched suffix.
- Deterministic output matches a cold prefill.
- Dense paged models, media prompts, path-dependent hybrid caches, and unsupported cache types remain on their existing paths.
- A focused regression test covers both the positive rotating-cache case and the existing negative cases.

tijs:
## Proposed changes

Fixes #454

- Pass the cache topology into the shared generation-suffix-stripped boundary helper.
- Admit standalone rotating/sliding-window caches through the existing proven predicate.
- Apply the same stripped-boundary store condition in `BatchEngine` and the native MTP path; DFlash2 uses the shared helper as well.
- Keep the existing hybrid-only `usesCanonicalHybridBoundary` and exact/N-1 disk-seed policy unchanged.
- Preserve dense, paged, media, path-dependent hybrid, and unsupported-cache safety gates.

## Verification

- `swift test --filter standaloneRotatingSharesStrippedBoundaryPolicy --jobs 4` — 1/1 passed.
- `swift test --filter BatchEngineGrowingChatCacheSourceTests --jobs 4` — the filtered source assertions completed, but the process exited during teardown because this machine cannot load the default Metal library (`Failed to load the default metallib`).
- `swift-format lint` exits 0 on the changed files; the repository also reports existing baseline formatting diagnostics.
- The CMake pre-commit hook passes; the repository-wide in-place Swift formatter was not run because it rewrites unrelated existing files.

No rotating/sliding-window checkpoint fixture was available for a runtime multi-turn restore or cold-prefill parity run, so the regression coverage is source-level and the runtime limitation is left explicit.

## Checklist

### DIFF
--- a/Libraries/MLXLMCommon/BatchEngine/BatchEngine.swift
+++ b/Libraries/MLXLMCommon/BatchEngine/BatchEngine.swift
@@ -3328,7 +3328,8 @@ public actor BatchEngine {
             let sharedPromptStripBoundary = TokenIterator.hybridStripBoundaryIndex(
                 coordinator: coordinator,
                 promptTokenIds: promptTokens,
-                input: slot.originalInput)
+                input: slot.originalInput,
+                cache: slot.cache)
             var sharedPromptRederivedStates: [Int: [MLXArray]]?
             let sharedPromptAdditionalBoundaries = Array(Set(
                 slot.originalInput.cachePrefixTokenCounts
@@ -3347,6 +3348,8 @@ public actor BatchEngine {
             // existing prompt/post-answer policy.
             // Still `isHybrid` on purpose — see the note in Evaluate.swift. This
             // suppresses every other boundary; only an SSM hybrid can afford that.
+            // Standalone rotating/SWA caches keep this policy untouched and only
+            // gain the stripped-boundary store itself (below).
             let usesCanonicalHybridBoundary =
                 coordinator.isHybrid && sharedPromptStripBoundary != nil
             let isReusablePrefixWarmup =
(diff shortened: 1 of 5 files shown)


jjang-ai:
## Exact-head review and bounded runtime control

Reviewed all runtime call sites at321cb40f72a6b471b464776510db0a29f297de74. No overlap in these files with the intervening main changes. Engine CI is waived for this merge; local evidence is the gate.

An additional tiny random-weight attention fixture exercised the production TokenIterator and BatchEngine.submit paths with two rotating rings of capacity8. A23-token first prompt stored boundary20; a new coordinator restored from disk for a27-token next prompt with a changed suffix.

- Before-control: removing only the shared helper's standalone-rotating admission lost the boundary and disk record in both paths (run PR456BeforeGateControl__234003, exit1).
- Exact-head after:3/3 checks with TF32 off and3/3 with default precision (PR456FinalExactHead__234048, exit0). Both paths restored20 tokens and forwarded exactly the seven changed suffix tokens. Solo warm/cold logits max absolute difference0; both paths' warm/cold greedy token94/94.
- Negative gates remain closed for missing coordinator, persistence disabled, dense cache, reusable warmup and rotating-plus-Mamba topology.
- Final bounded run peaked0.20GB tracked footprint, swap unchanged1.61GB, cleanup0/0/0. Earlier fixture compile errors are retained, not counted as runtime results.

This is cache-path evidence, not real-model coherence/speed, native-MTP/DFlash execution, media qualification or proof of an Osaurus repin. No release is included.

PR #468, opened 2026-09-12
Native MTP: measured admission, direct AR parking, and Flash cache checkpoints

jjang-ai:
## Scope

Consolidated native-MTP checkpoint, including the calibration correction proposed in #467 and the bounded Flash text prefill previously merged into the feature branch in #466. This branch now includes main; it does not remove the recurrent-cache work already in #465.

- Measure the initial native-MTP trial against true AR steps, including draft preparation and probe cycles.
- Park directly at AR after a measured AR-cost loss. Bounded re-entry starts at D1 and can recover within the selected depth ceiling. Calibration retains a productive depth; failed re-entry backs off.
- Bound Qwen Flash text prefill and preserve an owned prefill cache snapshot, avoiding unnecessary full-prompt replay after generation.
- Recognize durable self-contained Mamba disk snapshots without demanding an unrelated separate SSM payload. Separate-payload topologies retain their companion checks.
- Correct sampled staged verification's emitted correction/bonus token IDs. Flash sampled staged verification remains **explicitly opt-in**, with a default-false protocol capability; Qwen 27B is not opted in.
- Add opt-in monotonic iterator-delivery receipts, including the stream termination cause. These measure iterator delivery, not GPU or UI paint time.

## Evidence

Source tested: `3141a5c6305637a4cbaebc52f83e8a6610526f96`. The merge commit has the same tree as `81cf5586aa19a41103bade60b36c6c58026e36d3`.

Supervised local focused gate: 25 XCTest cases, 2 explicitly skipped, no failures; 18 Swift Testing cases, no failures. Exit 0. Peak tracked footprint 0.77 GiB; swap unchanged at 0.51 GiB; process/watchdog cleanup 0/0/0. Strict fixture parity used TF32 disabled and sampled-staged opt-in. The tests cover admission, immediate parking, bounded recovery, emitted IDs, Flash PLE/GDN/QSA continuation and disk restore, and timing receipts.

Earlier isolated before/after controls reproduce the initial-admission bypass and the D2/D3 extra-speculation-after-loss defect. Tests also exercise a changed-cost regime that recovers to D2/D3 within the ceiling.

[...]

PR #470, opened 2026-09-12
MTP: measure initial admission and park on measured AR loss

jjang-ai:
## Scope

Separate the governor correction from draft #468's unqualified sampled-verifier
and cache work. This supersedes #467's calibration-only proposal.

- Measure the initial selected depth through the existing timed admission probe.
- Park at measured AR after a speculative loss; re-entry probes start at D1 and
  may recover within the selected ceiling. Relative-depth/acceptance demotions
  remain separate from an absolute AR-loss verdict.
- Preserve a previously productive depth during periodic AR calibration, while
  still requiring the resumed trial to demonstrate benefit.
- Do not count initial admission as a resume or apply failed-reentry backoff to it.

No sampler, precision, model routing, projection, verifier or cache-format change.
This does not claim instantaneous or universal performance parity with Off.

## Evidence / remaining gate

Source head a5cf8a9ecff3ab5978285ee30799c881bd8a4c28. With only the runtime file
restored to main and identical candidate tests, all three new regression tests
fail (12 assertions): initial loss unjudged, no immediate parking, and D2/D3
calibration restarting at D1. Restoring the candidate reproduces its exact hash.

Bounded local run MTPAdmissionAfter0912__112011: 23 XCTest cases (2 explicit
dispatch-only skips), plus 3 Swift Testing cases; no failures. These are generated
fixtures, not model throughput. Peak tracked footprint1.26GB, flat swap.49GB,
normal pressure, cleanup0/0/0. Disabled-governor dispatch control recorded separately.

PARTIAL: exact-pinned Release app measurements, UI selected-depth execution,
short/long context and continuation proof are pending. Default sampled Flash
still uses sequential verification; experimental staged verification stays off.

### DIFF
--- a/Libraries/MLXLMCommon/SpecDec/NativeMTPTokenIterator.swift
+++ b/Libraries/MLXLMCommon/SpecDec/NativeMTPTokenIterator.swift
@@ -468,6 +468,9 @@ struct NativeMTPTokenIterator: TokenIteratorProtocol {
     private var arSafetyProbeStartedEmitted = 0
     private var arSafetyLastMeasuredToken = 0
     private var arSafetyCalibrationRemaining = 0
+    // Calibration refreshes the AR baseline; it is not evidence that the
+    // productive depth lost. Real loss recovery still begins at D1.
+    private var arSafetyCalibrationResumeDepth: Int?
     /// True while a kept re-entry is speculating: a trip in that state backs
     /// the resume interval off further instead of resetting it.
     private var arSafetyReentered = false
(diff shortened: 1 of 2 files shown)


jjang-ai:
Live gate remains PARTIAL at a5cf8a9e, consumed by app46e7274e (Release binary cdde1951). Flash: seven exact-output short/8339-token depth/Off rows and two visible tool turns completed; D3 warm35.50 versus Off35.61tok/s, long32.33 versus32.24. This demonstrates less speculative work, not a universal speed floor. The separate Qwen27B compatibility run failed its clock followup (fresh tool result present in prompt but answer reused an older timestamp). Its raw-completion control then trapped in Qwen35.prepare -> GatedDeltaNet1891 -> MLXArray indexing696, before any governor depth cycle. Investigating flat-token input handling with a tiny fixture. No merge while these live failures are unresolved.

